PR #78, opened 2025-09-25
Add getStages for case instances [PLT-90154]

swati354:
SDK method Response :
```
{
      {
        id: '',
        name : '',
        sla : {
            length: '',
            duration: '',
            escalationRule: []
            },
        status : ''
        tasks: [
              [
                      {
                        id : '',
                        name: '',
                        completedTime: '',
                        startedTime: '',
                        status: '',
                        type: ''',        
                      },
              ],
              [], 
            ]
      },
      {
      
      }
}
```

### DIFF
--- a/src/models/maestro/case-instances.types.ts
+++ b/src/models/maestro/case-instances.types.ts
@@ -86,3 +86,133 @@ export interface CaseAppConfig {
   caseSummary?: string;
   overview?: CaseAppOverview[];
 }
+
+/**
+ * Case stage task type 
+ */
+export enum StageTaskType {
+  EXTERNAL_AGENT = 'external-agent',
+  RPA = 'rpa',
+  AGENTIC_PROCESS = 'process',
+  AGENT = 'agent',
+  ACTION = 'action',
+  API_WORKFLOW = 'api-workflow'
+}
+
+/**
+ * Stage task information
+ */
+export interface StageTask {
+  id: string;
+  name: string;
+  completedTime: string;
+  startedTime: string;
+  status: string;
+  type: StageTaskType;
+}
+
+/**
+ * Escalation recipient scope
+ */
+export enum EscalationRecipientScope {
+  USER = 'user',
+  USER_GROUP = 'usergroup'
+}
+
+/**
+ * Escalation rule recipient information
+ */
+export interface EscalationRecipient {
+  /** Type of recipient (user or usergroup) */
+  scope: EscalationRecipientScope;
+  /** Identifier for a user/usergroup */
+  target: string;
+  /** The email id of the user/usergroup */
+  value: string;
+}
+
+/**
+ * Escalation action type
+ */
+export enum EscalationActionType {
+  NOTIFICATION = 'notification'
+}
+
+/**
+ * Escalation rule action configuration
+ */
+export interface EscalationAction {
+  type: EscalationActionType;
+  recipients: EscalationRecipient[];
+}
+
+/**
+ * Escalation rule trigger type
+ */
+export enum EscalationTriggerType {
+  SLA_BREACHED = 'sla-breached',
+  AT_RISK = 'at-risk'
+}
+
+/**
+ * Escalation rule trigger metadata
+ */
+export interface EscalationTriggerMetadata {
+  type?: EscalationTriggerType;
+  atRiskPercentage?: number;
+}
+
+/**
+ * Escalation rule configuration
+ */
+export interface EscalationRule {
+  triggerInfo: EscalationTriggerMetadata;
+  action?: EscalationAction;
+}
+
+/**
+ * SLA duration unit
+ */
+export enum SLADurationUnit {
+  HOURS = 'h',
+  DAYS = 'd',
+  WEEKS = 'w',
+  MONTHS = 'm'
+}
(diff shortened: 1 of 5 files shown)


src/services/maestro/case-instances.ts
```diff
@@ -269,4 +275,164 @@ export class CaseInstancesService extends BaseService implements CaseInstancesSe
...
+   * @param bindingsMap - Map of binding IDs to binding objects
+   * @returns CaseStage object
+   * @private
+   */
+  private createStageFromNode(
+    node: any,
+    executionMap: Map<string, CaseExecutionMetadata>,
+    bindingsMap: Map<string, any>
+  ): CaseStage {
+    const execution = executionMap.get(node.id);
```
Raina451, line 389: node?.id

src/services/maestro/case-instances.ts
```diff
@@ -269,4 +275,164 @@ export class CaseInstancesService extends BaseService implements CaseInstancesSe
...
+            taskName = this.resolveBinding(task.data.name, bindingsMap);
+          }
+          
+          const stageTask: StageTask = {
+            id: taskId || task.elementId || 'Undefined',
+            name: taskName || 'Undefined',
+            completedTime: taskExecution?.completedTimeUtc || 'Undefined',
+            startedTime: taskExecution?.startedTimeUtc || 'Undefined',
+            status: taskExecution?.status || 'NotStarted',
+            type: task.type || 'Undefined'
```
Raina451, line 366: lets have const for Undefined and NotStarted

Raina451: also since this is a string only we should have `Not Started ` instead of `NotStarted`

src/models/maestro/case-instances.types.ts
```diff
@@ -85,4 +85,86 @@ export interface CaseAppOverview {
...
+}
+
+/**
+ * Case element execution metadata
+ */
+export interface CaseExecutionMetadata {
+  elementId: string;
+  status: string;
+  startedTimeUtc: string;
+  completedTimeUtc: string;
```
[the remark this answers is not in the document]
swati354: Right now it is. We are not showing this anywhere. I was thinking of mapping it to startedTime and completedTime when we add the element-execution api.

src/models/maestro/case-instances.types.ts
```diff
@@ -85,4 +85,86 @@ export interface CaseAppOverview {
...
+}
+
+/**
+ * Case instance element execution response
+ */
+export interface CaseInstanceElementExecutionsResponse {
+  elementExecutions: CaseExecutionMetadata[];
+}
```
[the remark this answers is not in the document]
Raina451: Okay we may need to rename `ElementExecutions` also, to ElementHistory maybe . Let's do this also in next PR only

src/models/maestro/case-instances.types.ts
```diff
@@ -85,4 +85,127 @@ export interface CaseAppOverview {
...
   caseSummary?: string;
   overview?: CaseAppOverview[];
-}
\ No newline at end of file
+}
+
+/**
+ * Case stage task type 
+ */
+export enum CaseTaskType {
```
Raina451, line 93: *StageTaskType

src/models/maestro/case-instances.types.ts
```diff
@@ -85,4 +85,127 @@ export interface CaseAppOverview {
...
+  target: string;
+  /** The email id of the user/usergroup */
+  value: string;
+}
+
+/**
+ * Escalation rule action configuration
+ */
+export interface EscalationAction {
+  type: string;
```
Raina451, line 138: what is this type?

src/models/maestro/case-instances.types.ts
```diff
@@ -85,4 +85,127 @@ export interface CaseAppOverview {
...
+  DAYS = 'd',
+  WEEKS = 'w',
+  MONTHS = 'm'
+}
+
+/**
+ * SLA configuration for stages
+ */
+export interface StageSLA {
+  length: number;
```
Raina451, line 180: if there is no sla, does this comes as 0? and what will be duration then? null or some default value (m/d.. etc) ?

src/models/maestro/case-instances.types.ts
swati354: I see it always being 'notification'
Screenshot 2025-09-26 at 12 13 56 AM

src/services/maestro/case-instances.ts
```diff
@@ -269,4 +275,176 @@ export class CaseInstancesService extends BaseService implements CaseInstancesSe
...
+    ]);
+
+    // Extract execution history if successful, otherwise use null
+    const executionHistory = executionHistoryResponse.status === 'fulfilled' 
+      ? executionHistoryResponse.value.data 
+      : null;
+
+    if (executionHistoryResponse.status === 'rejected') {
+      console.debug(`Failed to fetch execution history for case instance ${caseInstanceId}:`, executionHistoryResponse.reason);
+    }
```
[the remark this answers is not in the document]
swati354: hmm right no need to tell about this failed fetch.
Will remove.

PR #79, opened 2025-09-26
Fix: initialize for oauth

Raina451:


src/core/auth/service.ts
```diff
@@ -257,6 +327,18 @@ export class AuthService extends BaseService {
...
+      clientId,
+      redirectUri,
+      baseUrl: this.config.baseUrl,
+      orgName: this.config.orgName,
+      tenantName: this.config.tenantName,
+      scope
+    };
+    
+    sessionStorage.setItem('uipath_sdk_oauth_context', JSON.stringify(oauthContext));
     sessionStorage.setItem('uipath_sdk_code_verifier', codeVerifier);
```
swati354, line 342: Same here. should we add browser check?

src/uipath.ts
```diff
@@ -174,6 +247,21 @@ export class UiPath {
   get assets(): AssetService {
     return this.getService(AssetService);
   }
+
+  /**
+   * Get stored OAuth context from session storage
+   */
+  private getStoredOAuthContext(): OAuthContext | null {
+    if (typeof window === 'undefined') return null;
```
swati354, line 255: if (!isBrowser)

swati354:
Could you add in description the problem and how session storage now fixes this?

src/core/auth/service.ts
```diff
@@ -33,15 +83,40 @@ export class AuthService extends BaseService {
...
   public async authenticate(config: Config): Promise<boolean> {
-    // Try to load token from storage first (only works for OAuth tokens)
+    // Check if we have URL parameters indicating OAuth callback
+    if (!isBrowser) {
+      return false;
+    }
+    
+    const urlParams = new URLSearchParams(window.location.search);
+    const code = urlParams.get('code');
+    const hasCodeVerifier = sessionStorage.getItem('uipath_sdk_code_verifier');
```
swati354, line 93: Could we use the isInOAuthCallback() method here
 https://github.com/UiPath/uipath-typescript/pull/79/files#diff-85eab706fc1dc043df813c330629512e1fa81dcfb55014a61d956ba9a9f74b96R137

src/core/auth/types.ts
```diff
@@ -18,3 +18,16 @@ export interface AuthToken {
...
+ */
+export interface OAuthContext {
+  codeVerifier: string;
+  clientId: string;
+  redirectUri: string;
+  baseUrl: string;
+  orgName: string;
+  tenantName: string;
+  scope: string;
+}
```
swati354, line 33: We are storing code verifier separately.
Do we need it here?

src/uipath.ts
```diff
@@ -26,66 +27,97 @@ export class UiPath {
...
+        baseUrl: storedContext.baseUrl,
+        orgName: storedContext.orgName,
+        tenantName: storedContext.tenantName,
+        clientId: storedContext.clientId,
+        redirectUri: storedContext.redirectUri,
+        scope: storedContext.scope
+      } as UiPathSDKConfig;
+    } else {
+      effectiveConfig = config;
+    }
```
swati354, line 47: isn't this the same thing we are doing in `_mergeConfigWithContext`

src/core/auth/types.ts
Raina451: its not necessary to be here, but its better to have clubbed in one place.

swati354 (approved):

PR #81, opened 2025-09-26
Add getTasks for Case Instances [PLT-90155]

swati354:

PR #74, opened 2025-09-24
Add case instances [PLT-90130]

swati354:

PR #85, opened 2025-09-29
Add Case Service 

swati354:
#73 #74 #78 #80 #81 #83


Raina451 (changes requested):
in mkdocs.yaml, w need to add cases in nav under maestro

PR #90, opened 2025-10-03
Fix: package compatibility issues across platform and modules - [PLT-87136] [PLT-90390] [PLT-90489]

Raina451:
SDK should be fully compatible across:

- Node.js
- Web browsers
- All bundlers (Webpack, Vite, Rollup, etc.)
- All frameworks (React, Vue, Angular, Next.js, etc.)
- Both ESM and CJS module systems

Issues:
- Single `index.d.ts` file was being used for both ESM and CommonJS, causing TypeScript to incorrectly identify the module format.
- Node.js-specific dependencies were breaking browser bundles
- Module Resolution Conflicts

Fix:
- Updated entry points: main → CJS, module → ESM with proper extensions
- added conditional exports with format-specific type mappings
- added optionaldependencies in package.json (npm installs them automatically but doesn't fail if installation fails (browser environments))
-  Gets externalized by Rollup automatically (in our allDependencies array)

  rollup.config.js

  - Added UMD browser bundle with Node.js dependency exclusions
  - Created separate type declaration builds (.mts, .cts, .d.ts)
  - Added environment-specific bundling logic

 - index.d.mts → ESM type declarations
  - index.d.cts → CommonJS type declarations
  - index.d.ts → Fallback for legacy tools
  
Removing unused pkg dependnecies from package.json
removing content-type from upload file method of buckets, as content type is not needed

Tested across:
- [x] react (create react app)
- [x] react + ts + vite
- [x] vue
- [x] webpack react
- [x] nodejs
- [x] nextjs
  - [x]  server side
  - [x]  client side
- [x] cjs

Screenshot 2025-10-03 at 4 19 44 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,14 +12,19 @@
         "agentic-ai"
     ],
     "type": "module",
-    "main": "./dist/index.js",
-    "module": "./dist/index.esm.js",
+    "main": "./dist/index.cjs",
+    "module": "./dist/index.mjs",
     "types": "./dist/index.d.ts",
     "exports": {
         ".": {
-            "import": "./dist/index.esm.js",
-            "require": "./dist/index.js",
-            "types": "./dist/index.d.ts"
+            "import": {
+                "types": "./dist/index.d.mts",
+                "default": "./dist/index.mjs"
+            },
+            "require": {
+                "types": "./dist/index.d.cts",
+                "default": "./dist/index.cjs"
+            }
         }
     },
     "files": [
@@ -28,38 +33,29 @@
     ],
     "scripts": {
         "build": "rollup -c",
+        "build:watch": "rollup -c -w",
         "clean": "rimraf dist && rimraf node_modules && rimraf package-lock.json",
         "docs:api": "typedoc"
     },
     "dependencies": {
         "@opentelemetry/sdk-logs": "^0.204.0",
         "axios": "^1.12.0",
-        "eventemitter3": "^5.0.1",
-        "file-type": "^21.0.0",
-        "mime-types": "^3.0.1",
-        "uuid": "^9.0.1",
-        "ws": "^8.16.0",
-        "zod": "^3.22.4",
-        "typedoc": "^0.28.13",
-        "typedoc-plugin-markdown": "^4.8.1",
-        "rimraf": "^6.0.1"
+        "zod": "^3.22.4"
     },
     "devDependencies": {
         "@rollup/plugin-commonjs": "^25.0.0",
         "@rollup/plugin-json": "^6.0.0",
         "@rollup/plugin-node-resolve": "^15.2.0",
         "@rollup/plugin-typescript": "^11.1.0",
-        "@types/jest": "^29.5.14",
-        "@types/mime-types": "^3.0.1",
         "@types/node": "^20.11.19",
         "@types/uuid": "^9.0.8",
-        "@types/ws": "^8.5.10",
         "builtin-modules": "^3.3.0",
         "dotenv": "^17.2.0",
-        "jest": "^29.7.0",
+        "rimraf": "^6.0.1",
         "rollup": "^4.0.0",
         "rollup-plugin-dts": "^6.1.0",
-        "ts-jest": "^29.4.0",
+        "typedoc": "^0.28.13",
+        "typedoc-plugin-markdown": "^4.8.1",
         "typescript": "^5.3.3"
     },
     "repository": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,65 +1,126 @@
-import resolve from '@rollup/plugin-node-resolve';
-import commonjs from '@rollup/plugin-commonjs';
-import typescript from '@rollup/plugin-typescript';
-import dts from 'rollup-plugin-dts';
-import json from '@rollup/plugin-json';
-import builtins from 'builtin-modules';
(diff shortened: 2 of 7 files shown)


rollup.config.js
```diff
@@ -15,43 +16,115 @@ const allDependencies = [
...
+const createBrowserPlugins = () => [
+  replace({
+    preventAssignment: true,
+    values: {
+      // Replace Node.js dynamic imports with empty promises for browser bundle
+      "import('file-type')": "Promise.resolve({})",
+      "import('ws')": "Promise.resolve({})",
+      // Replace Node.js crypto require with Web Crypto API
+      "const crypto = require('crypto')": "const crypto = globalThis.crypto || window.crypto",
+      "require('crypto')": "globalThis.crypto || window.crypto"
```
Raina451, line 53: need better approach here

swati354: Wouldn't the nodejs crypto be different from browser crypto? Is it okay to just replace it?

rollup.config.js
```diff
@@ -15,43 +16,115 @@ const allDependencies = [
   ...builtins // Node.js built-in modules
 ];
 
+// Node.js-specific dependencies to exclude from browser bundle
+const nodeDependencies = [
+  'file-type',
+  'ws', 
+  '@opentelemetry/sdk-logs'
+];
+
```
swati354, line 25: Should this be handled here or in source code?

rollup.config.js
```diff
@@ -15,43 +16,115 @@ const allDependencies = [
...
+    input: 'src/index.ts',
+    output: {
+      file: 'dist/index.umd.js',
+      format: 'umd',
+      name: 'UiPath',
+      globals: {
+        'file-type': 'FileType',
+        'ws': 'WebSocket',
+        '@opentelemetry/sdk-logs': 'OpenTelemetry'
+      }
```
swati354, line 104: what is globals for?

package.json
```diff
@@ -37,17 +43,18 @@
         "eventemitter3": "^5.0.1",
         "file-type": "^21.0.0",
         "mime-types": "^3.0.1",
-        "uuid": "^9.0.1",
-        "ws": "^8.16.0",
-        "zod": "^3.22.4",
+        "rimraf": "^6.0.1",
```
swati354, line 46: shouldn't `rimraf` be a dev dependency?

rollup.config.js
Raina451: Made the changes to better approach

rollup.config.js
Raina451: [yes](https://github.com/UiPath/uipath-typescript/pull/90#discussion_r2409270796)

src/utils/platform-imports.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Handles conditional imports for Node.js-specific modules with browser fallbacks
+ */
+import { isBrowser } from './platform';
+
+/**
+ * Crypto implementation utility
+ * Returns Web Crypto API in browser or Node.js crypto module
+ */
+export function getCrypto(): any {
+  if (isBrowser && typeof window !== 'undefined') {
```
swati354, line 12: the isBrowser check alredy checks for `typeof window !== 'undefined'`

src/utils/platform-imports.ts
swati354, line 1: Do we need a separate file for this or can we merge this with the browser check file (`platform.ts`)

package.json
```diff
@@ -28,38 +33,30 @@
...
         "clean": "rimraf dist && rimraf node_modules && rimraf package-lock.json",
         "docs:api": "typedoc"
     },
     "dependencies": {
         "@opentelemetry/sdk-logs": "^0.204.0",
         "axios": "^1.12.0",
-        "eventemitter3": "^5.0.1",
-        "file-type": "^21.0.0",
-        "mime-types": "^3.0.1",
         "uuid": "^9.0.1",
```
swati354, line 43: remove

rollup.config.js
swati354: nice. Could you update the description also.

src/utils/platform-imports.ts
Raina451: lets keep same, will make the change

src/models/orchestrator/buckets.types.ts
```diff
@@ -93,16 +93,6 @@ export interface BucketGetUriOptions extends BaseOptions {
...
- * Request options for getting a write URI for a file in a bucket
- */
-export interface BucketGetWriteUriOptions extends BucketGetUriOptions {
-  /**
-   * ContentType for S3 access policy
-   */
-  contentType?: string;
-}
-
 /**
```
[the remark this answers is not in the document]
Raina451: BucketGetUriOptions does not have `content` and also it has `expiryInMinutes` which is not needed for writeoptions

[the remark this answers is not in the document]
Raina451: we did remove get write uri method back then when we onboard uploadfile method, hence BucketGetWriteUriOptions is internal and not needed, so we are using BucketGetUriOptions directly in [_getWriteUri](https://github.com/UiPath/uipath-typescript/blob/62b692451a2605b0ab44c6a71d2a3b6c7bc08de9/src/services/orchestrator/buckets.ts#L447)

swati354 (approved):

PR #111, opened 2025-10-28
Add orchestrator buckets unit tests [PLT-91793]

Raina451:


tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+// Import mock objects using vi.hoisted() - this ensures they're available before vi.mock() calls
+const mocks = vi.hoisted(() => {
+  // Import/re-export the mock utilities from core
+  return import('../../../utils/mocks/core');
+});
+
+// Setup mocks at module level
+// NOTE: We do NOT mock transformData - we want to test the actual transformation logic!
+vi.mock('../../../../src/utils/pagination/helpers', async () => (await mocks).mockPaginationHelpers);
+vi.mock('../../../../src/core/context/execution', async () => (await mocks).mockExecutionContext);
```
swati354: Do we need this?

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+      expect(PaginationHelpers.getAll).toHaveBeenCalledWith(
+        expect.objectContaining({
+          serviceAccess: expect.any(Object),
+          getEndpoint: expect.any(Function),
+          getByFolderEndpoint: BUCKET_ENDPOINTS.GET_BY_FOLDER,
+          transformFn: expect.any(Function),
+          pagination: expect.objectContaining({
+            paginationType: 'offset',
+            itemsField: 'value',
+            totalCountField: '@odata.count'
```
swati354: We have constants for these values as part of pagination constants.

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+
+    it('should handle API errors', async () => {
+      const error = createMockError(TEST_CONSTANTS.ERROR_MESSAGE);
+      vi.mocked(PaginationHelpers.getAll).mockRejectedValue(error);
+
+      await expect(bucketService.getAll()).rejects.toThrow(TEST_CONSTANTS.ERROR_MESSAGE);
+    });
+  });
+
+  describe('getFileMetaData', () => {
```
swati354: in one of the tests for `getFileMetaData`, can we also add the check if the correct endpoint was called?

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+            [BUCKET_TEST_CONSTANTS.CONTENT_TYPE_HEADER]: BUCKET_TEST_CONSTANTS.CONTENT_TYPE, // Verify header transformation from array to record
+            [TEST_CONSTANTS.AUTHORIZATION_HEADER]: TEST_CONSTANTS.BEARER_PREFIX + ' ' + TEST_CONSTANTS.DEFAULT_ACCESS_TOKEN // Verify secret token authentication
+          })
+        })
+      );
+
+      expect(result.success).toBe(true);
+    });
+
+    it('should upload file whith oauth token and refresh when expired for requiresAuth true', async () => {
```
swati354: 1.  typo for `with` 
2. Could we slightly change the wordings for this?
`should automatically refresh expired OAuth token during file upload`

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+      expect(result.statusCode).toBe(BUCKET_TEST_CONSTANTS.UPLOAD_SUCCESS_STATUS_CODE);
+
+      // Verify axios.put was called with refreshed token
+      expect(axios.put).toHaveBeenCalledWith(
+        BUCKET_TEST_CONSTANTS.UPLOAD_URI,
+        fileContent,
+        expect.objectContaining({
+          headers: expect.objectContaining({
+            [BUCKET_TEST_CONSTANTS.CONTENT_TYPE_HEADER]: BUCKET_TEST_CONSTANTS.CONTENT_TYPE,
+            [TEST_CONSTANTS.AUTHORIZATION_HEADER]: TEST_CONSTANTS.BEARER_PREFIX + ' ' + TEST_CONSTANTS.DEFAULT_ACCESS_TOKEN
```
swati354: If both the expired and the refresh token use the same constant - `TEST_CONSTANTS.DEFAULT_ACCESS_TOKEN`, is this check valid?
Should we use different constant values for expired and refresh?

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+
+  describe('uploadFile', () => {
+
+    it('should upload file with Buffer content', async () => {
+
+      // Mock the internal _getWriteUri call
+      const mockUploadUriResponse = createMockWriteUriApiResponse({ 
+        RequiresAuth: false,
+        Headers: { Keys: [], Values: [] }
+      });
```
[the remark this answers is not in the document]
swati354: Oh okay, this is the mock response for the actual get call that happens?

tests/utils/constants/buckets.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Test constants for Bucket service tests
+ * Provides consistent test data for bucket-related operations
+ */
+
+export const BUCKET_TEST_CONSTANTS = {
+  // Bucket identifiers
+  BUCKET_ID: 123,
+  BUCKET_ID_2: 456,
+  BUCKET_NAME: 'Test Bucket',
+  BUCKET_NAME_2: 'Test Bucket 2',
```
swati354: remove `BUCKET_ID_2` and `BUCKET_NAME_2`

tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -0,0 +1,689 @@
...
+        expect.objectContaining({
+          headers: expect.objectContaining({
+            [BUCKET_TEST_CONSTANTS.CONTENT_TYPE_HEADER]: BUCKET_TEST_CONSTANTS.CONTENT_TYPE,
+            [TEST_CONSTANTS.AUTHORIZATION_HEADER]: TEST_CONSTANTS.BEARER_PREFIX + ' ' + TEST_CONSTANTS.DEFAULT_ACCESS_TOKEN
+          })
+        })
+      );
+    });
+
+    it('should upload file with oauth token when not expired', async () => {
```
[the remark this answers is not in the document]
Raina451: Test 1: "should refresh expired OAuth token and use the new one"
Uses expired token, verifies refresh is called and new token is used.

this test "should use valid OAuth token without refreshing"
Uses valid token, verifies refresh is not called and original token is used.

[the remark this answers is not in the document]
swati354: no the second one -`should upload Blob content file with authentication when requiresAuth is true`
and this `should upload file with oauth token when not expired` test the same thing?

PR #112, opened 2025-10-29
Add unit tests for pagination [PLT-91763]

swati354:


tests/unit/utils/pagination/helpers.test.ts
```diff
@@ -0,0 +1,871 @@
...
+      expect(result).toEqual({});
+    });
+  });
+
+  describe('getAllPaginated', () => {
+    let mockServiceAccess: PaginationServiceAccess;
+
+    beforeEach(() => {
+      mockServiceAccess = createMockPaginationServiceAccess();
+    });
```
Raina451: in all tests we are setting up same createMockPaginationServiceAccess in before each, can't we have this before each for whole test suite?

Raina451: we should also rest mocks after each test

tests/unit/utils/pagination/helpers.test.ts
```diff
@@ -0,0 +1,871 @@
...
+        PaginationType.OFFSET,
+      )).not.toThrow();
+    });
+
+    it('should throw error when cursor value is undefined', () => {
+      const options = { cursor: { value: undefined as any } };
+
+      expect(() => PaginationHelpers.validateCursor(options))
+        .toThrow(PAGINATION_TEST_CONSTANTS.ERROR_CURSOR_MUST_BE_VALID);
+    });
```
Raina451: isnt first test covers this?

tests/unit/utils/pagination/helpers.test.ts
```diff
@@ -0,0 +1,871 @@
...
+        PAGINATION_TEST_CONSTANTS.ENDPOINT_API_FOLDERS_ITEMS,
+        expect.objectContaining({
+          headers: { [FOLDER_ID]: TEST_CONSTANTS.FOLDER_ID.toString() }
+        })
+      );
+    });
+
+    it('should handle folderId properly in paginated flow', async () => {
+      const mockResponse = createMockPaginatedResponse([]);
+      vi.mocked(mockServiceAccess.requestWithPagination).mockResolvedValue(mockResponse);
```
Raina451: shouldnt the logic be same when folderid is provided irrespective of paginated or non paginated flow? isnt only one test enough?

swati354: no they are actually testing two different code flows. 
We need to verify that it works in both code paths.
```
// Paginated flow
    if (isPaginationRequested) {
      return PaginationHelpers.getAllPaginated<TRaw, TTransformed>({
        serviceAccess: config.serviceAccess,
        getEndpoint: config.getEndpoint,
        folderId,
        paginationParams: cursor ? { cursor, pageSize } : jumpToPage ? { jumpToPage, pageSize } : { pageSize },
        additionalParams: prefixedOptions,
        transformFn: config.transformFn,
        options: {
          ...paginationOptions,
          paginationParams: config.pagination?.paginationParams
        }
      }) as any; // Type assertion needed due to conditional return
    }
    
    // Non-paginated flow
    const byFolderEndpoint = config.getByFolderEndpoint || config.getEndpoint(folderId);
    return PaginationHelpers.getAllNonPaginated<TRaw, TTransformed>({
      serviceAccess: config.serviceAccess,
      getAllEndpoint: config.getEndpoint(),
      getByFolderEndpoint: byFolderEndpoint,
      folderId,
      additionalParams: prefixedOptions,
      transformFn: config.transformFn,
      options: {
        itemsField: paginationOptions.itemsField,
        totalCountField: paginationOptions.totalCountField
      }
    }) as any;
```

tests/unit/utils/pagination/helpers.test.ts
Raina451: first test: https://github.com/UiPath/uipath-typescript/blob/b20b96ee1b8aa06d7ad5b999f3af019e250c8f22/tests/unit/utils/pagination/helpers.test.ts#L111
has no cursor, which means cursor value is also undefined?

PR #115, opened 2025-10-31
Add configurable token storage and logout method

swati354:
1. Adds a tokenStorage configuration option to allow users to choose between `localStorage` and `sessionStorage` for OAuth token storage, with `sessionStorage` as the secure default. 
2. Also implements a proper `logout()` method to clear all authentication state.

Motivation
**Problem 1: Cross-Tab Token Persistence**
With sessionStorage, OAuth tokens don't persist across browser tabs opened via `target="_blank"` in modern browsers.
Impact: Since sessionStorage isn’t shared across tabs, authentication tokens aren't available when a new page opens in a new tab, causing reauthentication or redirect issues.

**Problem 2: SDK lacked a proper logout mechanism.**

### DIFF
--- a/src/core/auth/service.ts
+++ b/src/core/auth/service.ts
@@ -213,6 +213,29 @@ export class AuthService extends BaseService {
   }
 
   /**
+   * Clears authentication state (logout)
+   * Removes tokens from storage and clears OAuth context
+   */
+  public logout(): void {
+    // Clear the token from storage and memory
+    this.tokenManager.clearToken();
+
+    // Clear OAuth-specific session storage items
+    // Note: These items are normally cleared after successful authentication in _handleOAuthCallback(),
+    // but we clear them here as well to handle edge cases where OAuth flow was interrupted:
+    // - User navigated away during OAuth flow (before callback completed)
+    // - OAuth error occurred that didn't clear these items
+    if (isBrowser) {
+      try {
+        sessionStorage.removeItem('uipath_sdk_oauth_context');
+        sessionStorage.removeItem('uipath_sdk_code_verifier');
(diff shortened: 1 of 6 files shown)


src/core/auth/service.ts
```diff
@@ -212,6 +212,29 @@ export class AuthService extends BaseService {
...
+    // Note: These items are normally cleared after successful authentication in _handleOAuthCallback(),
+    // but we clear them here as well to handle edge cases where OAuth flow was interrupted:
+    // - User navigated away during OAuth flow (before callback completed)
+    // - OAuth error occurred that didn't clear these items
+    if (isBrowser) {
+      try {
+        sessionStorage.removeItem('uipath_sdk_oauth_context');
+        sessionStorage.removeItem('uipath_sdk_code_verifier');
+      } catch (error) {
+        console.warn('Failed to clear OAuth context from session storage', error);
```
vnaren23: @Raina451 @swati354 Something to ponder later. We need to maintain a style for any/all info/warn/error statements throughout the SDK

src/core/auth/token-manager.ts
```diff
@@ -53,28 +59,28 @@ export class TokenManager {
...
       const tokenInfo = this._parseTokenInfo(storedToken);
       if (!tokenInfo) {
         // Invalid token format, clear it
-        sessionStorage.removeItem(this._getStorageKey());
+        this.storage!.removeItem(this._getStorageKey());
         return false;
       }
       
       // Check if token is expired
       if (this.isTokenExpired(tokenInfo)) {
```
vnaren23: Why not combine both conditions?

PR #124, opened 2025-11-18
Add 'asTaskAdmin' flag in options for fetching tasks [PLT-90774]

swati354:
Add `asAdmin` parameter to `tasks.getAll()`

Adds optional asAdmin parameter to `tasks.getAll()` to support fetching tasks with different permission levels.

**Changes**
New parameter: `asAdmin?: boolean` in TaskGetAllOptions
Endpoint selection:
```
asAdmin: true → Uses GetTasksAcrossFoldersForAdmin (fetches tasks from folders where the user has Task.View, Task.Edit and TaskAssignment.Create permissions)
asAdmin: false or omitted → Uses GetTasksAcrossFolders (fetches tasks from folders where the user has Task.View and Task.Edit permissions)
```

Tests: Added 2 new unit test cases covering these scenarios
**Usage**
```
// Admin permissions
const tasks = await sdk.tasks.getAll({ asAdmin: true });

// Normal permissions (default)
const tasks = await sdk.tasks.getAll();
```

Is this a breaking change? 
No. Existing code works without changes. We have just added an optional parameter.

### DIFF
--- a/src/models/action-center/tasks.types.ts
+++ b/src/models/action-center/tasks.types.ts
@@ -225,6 +225,14 @@ export type TaskGetAllOptions = RequestOptions & PaginationOptions & {
    * Optional folder ID to filter tasks by folder
    */
   folderId?: number;
+  /**
+   * Optional flag to fetch tasks using admin permissions
+   * When true, fetches tasks across folders
+   * where the user has at least Task.View, Task.Edit and TaskAssignment.Create permissions
+   * When false or omitted, fetches tasks across folders
+   * where the user has at least Task.View and Task.Edit permissions
+   */
+  asTaskAdmin?: boolean;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/models/action-center/tasks.types.ts
```diff
@@ -225,6 +225,14 @@ export type TaskGetAllOptions = RequestOptions & PaginationOptions & {
...
    */
   folderId?: number;
+  /**
+   * Optional flag to fetch tasks using admin permissions
+   * When true, fetches tasks from folders
+   * where the user has at least Task.View, Task.Edit and TaskAssignment.Create permissions
+   * When false or omitted, fetches tasks from folders
+   * where the user has at least Task.View and Task.Edit permissions
+   */
+  asAdmin?: boolean;
```
Raina451: `asActionAdmin`? or something else? because orgAdmin doesn't necessarily have these permissions

swati354: What about `asTaskAdmin`?

Raina451 (approved):

PR #126, opened 2025-11-20
feat: add auth arguments to cli cmds - [PLT-92507]

Raina451:
- Adding auth envs are cli arguments 

#### `uipath register app`
```bash
--baseUrl         # UiPath base URL (e.g., https://cloud.uipath.com)
--orgId           # Organization ID
--tenantId        # Tenant ID
--tenantName      # Tenant name
--folderKey       # Folder key
--accessToken     # token for authentication
```

#### `uipath publish`
```bash
--baseUrl         # UiPath base URL
--orgId           # Organization ID
--tenantId        # Tenant ID
--tenantName      # Tenant name
--accessToken     # token for authentication
```

- introduce client credential flow for uipath auth

if client id client secret provided then use client credential flow instead of pkce and just provide authtoken

Screenshot 2025-12-04 at 4 29 24 PM

### DIFF
--- a/packages/cli/src/auth/core/token-manager.ts
+++ b/packages/cli/src/auth/core/token-manager.ts
@@ -5,10 +5,11 @@ import { SelectedTenant } from '../services/portal.js';
 import { AUTH_CONSTANTS } from '../../constants/auth.js';
 import { getBaseUrl } from '../utils/url.js';
 import { calculateExpirationTime } from '../utils/date.js';
+import { atomicWriteJson } from '../../utils/env-config.js';
 
-const UIPATH_DIR = path.join(process.cwd(), '.uipath');
-const AUTH_FILE = path.join(UIPATH_DIR, '.auth.json');
-const ENV_FILE = path.join(process.cwd(), '.env');
+const UIPATH_DIR = path.join(process.cwd(), AUTH_CONSTANTS.FILES.UIPATH_DIR);
+const AUTH_FILE = path.join(UIPATH_DIR, AUTH_CONSTANTS.FILES.AUTH_FILE);
+const ENV_FILE = path.join(process.cwd(), AUTH_CONSTANTS.FILES.ENV_FILE);
 
 export interface StoredAuth {
   accessToken: string;
@@ -155,8 +156,3 @@ const updateEnvFile = async (vars: Record<string, string>): Promise<void> => {
   await fs.writeFile(ENV_FILE, newLines.join('\n') + '\n', 'utf-8');
 };
 
-const atomicWriteJson = async (filePath: string, data: any): Promise<void> => {
-  const tmpPath = `${filePath}.tmp`;
-  await fs.writeJson(tmpPath, data, { spaces: 2 });
-  await fs.rename(tmpPath, filePath);
-};
--- a/packages/cli/src/commands/register/app.ts
+++ b/packages/cli/src/commands/register/app.ts
@@ -9,7 +9,7 @@ import { EnvironmentConfig, AppConfig, AppType } from '../../types/index.js';
 import { ACTION_SCHEMA_CONSTANTS, API_ENDPOINTS, AUTH_CONSTANTS } from '../../constants/index.js';
 import { MESSAGES } from '../../constants/messages.js';
 import { createHeaders, buildAppUrl } from '../../utils/api.js';
-import { getEnvironmentConfig, isValidAppName } from '../../utils/env-config.js';
+import { getEnvironmentConfig, isValidAppName, atomicWriteFileSync } from '../../utils/env-config.js';
 import { handleHttpError } from '../../utils/error-handler.js';
 import { track } from '../../telemetry/index.js';
 import { readAndParseActionSchema } from '../../utils/action-schema.js';
@@ -78,7 +78,7 @@ export default class RegisterApp extends Command {
     // Validate environment variables or flags
     const envConfig = getEnvironmentConfig(AUTH_CONSTANTS.REQUIRED_ENV_VARS.REGISTER_APP, this, flags);
     if (!envConfig) {
-      return;
+      process.exit(1);
     }
 
     // Validate name flag if provided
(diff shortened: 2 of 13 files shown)


packages/cli/src/auth/core/oidc.ts
```diff
@@ -105,4 +108,53 @@ export const parseJWT = (token: string): AccessTokenData => {
...
     organizationId: claims.organization_id || claims.prt_id,
   };
+};
+
+export interface ConfidentialClientConfig {
+  clientId: string;
+  clientSecret: string;
+  domain: string;
+  scope?: string;
+}
```
swati354: move above the methods

packages/cli/src/auth/core/oidc.ts
```diff
@@ -105,4 +108,53 @@ export const parseJWT = (token: string): AccessTokenData => {
...
+    headers: {
+      'Content-Type': AUTH_CONSTANTS.CONTENT_TYPES.FORM_URLENCODED,
+    },
+    body: params.toString(),
+  });
+
+  if (!response.ok) {
+    const errorText = await response.text();
+    throw new Error(`Authentication failed: ${errorText || response.status}`);
+  }
```
swati354: can we use the `handleHttpError` method here?

packages/cli/src/auth/server/auth-server.ts
```diff
@@ -146,15 +145,19 @@ export class AuthServer {
     });
   }
 
+  private buildRedirectUri(): string {
+    return AUTH_CONSTANTS.OAUTH.REDIRECT_URI_TEMPLATE.replace('{PORT}', this.port.toString());
+  }
```
swati354: A similar method is added in the `oidc.ts` file. Can we extract this method?

packages/cli/src/commands/register/app.ts
```diff
@@ -77,18 +96,27 @@ export default class RegisterApp extends Command {
     await this.registerApp(appName, appVersion, isActionApp, envConfig);
   }
 
-  private async validateEnvironment(): Promise<EnvironmentConfig | null> {
+  private async validateEnvironment(flags: any): Promise<EnvironmentConfig | null> {
```
swati354: `flags: any`
Here we do know the flags right?

packages/cli/src/commands/register/app.ts
```diff
@@ -77,18 +96,27 @@ export default class RegisterApp extends Command {
...
     const requiredEnvVars = [
       'UIPATH_BASE_URL',
-      'UIPATH_ORG_ID', 
+      'UIPATH_ORG_ID',
       'UIPATH_TENANT_ID',
       'UIPATH_TENANT_NAME',
       'UIPATH_FOLDER_KEY',
       'UIPATH_BEARER_TOKEN'
     ];
```
swati354: should we move this to the auth constant file?

packages/cli/src/commands/auth.ts
```diff
@@ -265,9 +288,39 @@ export default class Auth extends Command {
...
+        clientSecret,
+        domain,
+        scope,
+      });
+
+      spinner.succeed(MESSAGES.SUCCESS.AUTHENTICATION_SUCCESS.replace('✓ ', ''));
+
+      // For client credentials flow, just output the token
+      this.log(chalk.green('\nAccess Token:'));
+      this.log(tokens.accessToken);
```
swati354: We just log the token? The users will copy paste this?

packages/cli/src/constants/messages.ts
```diff
@@ -9,7 +9,7 @@ export const MESSAGES = {
     ACTION_SCHEMA_REQUIRED: '❌ action-schema.json file not found in current working directory',
     
     // Authentication & Authorization  
-    AUTHENTICATION_FAILED: 'Authentication failed. Please check your UIPATH_BEARER_TOKEN or run "uipath auth" to authenticate',
+    AUTHENTICATION_FAILED: 'Authentication failed. Please check your UIPATH_BEARER_TOKEN or run "uipath auth" to authenticate or auth pass arguments',
```
swati354: `or auth pass arguments`
typo?

packages/cli/src/commands/register/app.ts
```diff
@@ -42,6 +43,24 @@ export default class RegisterApp extends Command {
...
+    }),
+    tenantName: Flags.string({
+      description: 'UiPath tenant name',
+    }),
+    folderKey: Flags.string({
+      description: 'UiPath folder key',
+    }),
+    authToken: Flags.string({
+      description: 'UiPath authentication token',
+    }),
```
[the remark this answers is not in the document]
Raina451: I actually now changed the naming to accessToken and also the env variable name from `UIPATH_BEARER_TOKEN` to `UIPATH_ACCESS_TOKEN` as python cli uses `UIPATH_ACCESS_TOKEN`

packages/cli/src/commands/auth.ts
Raina451: Yes, this is intentional. The client credentials flow and this whole arguments thing is meant for non-interactive flow, so it should not depend on .env files or any local file storage. Users can either copy the token manually or, in automated deployment, capture it directly from the command output for use in their CI/CD pipelines or scripts.

packages/cli/src/utils/env-config.ts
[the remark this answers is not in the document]
Raina451: done also changed the file name to `env-config` , validator.ts was not making much sense

packages/cli/src/auth/core/token-manager.ts
swati354: unrelated to this PR, but there is an unused `os` import in this file, could you remove that too.

swati354:
Can you update the cli readme as well. The token flag has been changed from `UIPATH_BEARER_TOKEN` to `UIPATH_ACCESS_TOKEN`.
Change the base url as well. Right now it is
`UIPATH_BASE_URL=https://your-orchestrator.com`

packages/cli/src/utils/env-config.ts
```diff
@@ -0,0 +1,163 @@
...
+  Object.values(ENV_CONFIG).map(cfg => [cfg.envVar, `${cfg.flag} ${cfg.example}`])
+) as Record<string, string>;
+
+/**
+ * Validates app/package name to ensure it only contains allowed characters.
+ * Allowed: letters (a-z, A-Z), numbers (0-9), underscores (_), and hyphens (-)
+ */
+export function isValidAppName(name: string): boolean {
+  const validNameRegex = VALID_NAME_REGEX;
+  return validNameRegex.test(name);
```
swati354: `return VALID_NAME_REGEX.test(name);` ?

Sarath1018 (approved):

PR #130, opened 2025-11-26
fix: prevent special char in app name for register and pack cmd - [PLT-93593]

Raina451:
Screenshot 2025-11-27 at 8 22 22 AM

### DIFF
--- a/packages/cli/src/commands/pack.ts
+++ b/packages/cli/src/commands/pack.ts
@@ -8,6 +8,7 @@ import { v4 as uuidv4 } from 'uuid';
 import JSZip from 'jszip';
 import { AppConfig } from '../types/index.js';
 import { MESSAGES } from '../constants/messages.js';
+import { isValidAppName } from '../utils/validator.js';
 import { track } from '../telemetry/index.js';
 
 export default class Pack extends Command {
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/pack.ts
```diff
@@ -124,7 +125,13 @@ export default class Pack extends Command {
       this.log(chalk.red(MESSAGES.ERRORS.PACKAGE_NAME_REQUIRED));
       process.exit(1);
     }
-    
+
+    // Validate package name characters
+    if (!isValidAppName(packageName)) {
+      this.log(chalk.red(MESSAGES.VALIDATIONS.APP_NAME_INVALID_CHARS));
```
swati354: The message says `App name` though? Would this be confusing?

swati354:
> > Is there any other validation in app name as well? Other than this, what is allowed/ not allowed?
> 
> `_` , `-` can exists but other special chars like `&`, `%`, `^`, `*`, `@`, `!`, `#`, `^`, `$`, `(` ,`)`, `{` , `}` etc cannot exist. I'll add validation in same pr

Can `.` exist?

packages/cli/src/commands/pack.ts
Raina451: this was intentional, because we want to surface app name is same as package name

Raina451:
> Can . exist?

nope, no special char can exist except `-` and `_` . That's why regex is: `/^[a-zA-Z0-9_-]+$/`

PR #131, opened 2025-11-26
Plt 92708 autharization modular

SunilYadav-uipath:
## Description
Pr contains the authentication simplification with modularization. Usage is give below

## Changes
1. New Authentication class UipathCore which contains logic of authentication only
2. Authentication Logic of Uipath class will reuse the same authentication logic of UipathCore
3. Entities Service is currently migrated to modular imports, contains enteries of this file path in package.json and rollup bundler
4. Fixed all the test cases due to modularization changes

## Screenshots (to show code Changes if any)

Pr contains the authentication simplification with modularization approach of client injection. Usage is give below
```
 
  const uiPath = new UiPath({
    baseUrl: 'https://cloud.uipath.com',
    orgName: 'myorg',
    tenantName: 'mytenant',
    clientId: 'xxx',
    redirectUri: 'http://localhost:3000/callback'
  });
 
  await uiPath.initialize();
 
  const entitiesService = new Entities(uiPath);
  const allEntities = await entitiesService.getAll();
```

### DIFF
--- a/src/core/internals/registry.ts
+++ b/src/core/internals/registry.ts
@@ -60,18 +60,4 @@ export class SDKInternalsRegistry {
     }
     return internals;
   }
-
-  /**
-   * Check if an instance is registered
-   */
-  static has(instance: object): boolean {
-    return this.store.has(instance);
-  }
-
-  /**
-   * Remove an instance from registry (optional cleanup)
-   */
-  static delete(instance: object): boolean {
-    return this.store.delete(instance);
-  }
 }
--- a/tests/utils/setup.ts
+++ b/tests/utils/setup.ts
@@ -3,48 +3,26 @@ import { UiPathConfig } from '../../src/core/config/config';
 import { ExecutionContext } from '../../src/core/context/execution';
 import { TokenManager } from '../../src/core/auth/token-manager';
 import type { UiPath } from '../../src/core/uipath';
-import { SDKInternalsRegistry, type PrivateSDK } from '../../src/core/internals';
+import type { BaseConfig } from '../../src/core/config/sdk-config';
+import { SDKInternalsRegistry } from '../../src/core/internals';
 
 /**
- * Helper functions for accessing private SDK components in tests.
- * Now uses SDKInternalsRegistry instead of Symbol pattern.
+ * Interface for mockable UiPath properties used in tests.
+ * Only includes the public surface needed for testing services.
  */
+interface MockableUiPath {
+  config: Readonly<BaseConfig>;
+  isAuthenticated: () => boolean;
+  isInitialized: () => boolean;
+}
 
 /**
- * Gets the private SDK components from a UiPath instance
- * @param instance - UiPath instance
- * @returns PrivateSDK containing config, context, and tokenManager
+ * Interface for mockable TokenManager properties used in tests.
  */
-export const getPrivateSDK = (instance: UiPath): PrivateSDK => {
-  return SDKInternalsRegistry.get(instance);
-};
-
-/**
- * Gets the config from a UiPath instance
- * @param instance - UiPath instance
- * @returns UiPathConfig object
- */
-export const getConfig = (instance: UiPath): UiPathConfig => {
-  return SDKInternalsRegistry.get(instance).config;
-};
-
-/**
- * Gets the execution context from a UiPath instance
- * @param instance - UiPath instance
- * @returns ExecutionContext object
- */
-export const getContext = (instance: UiPath): ExecutionContext => {
-  return SDKInternalsRegistry.get(instance).context;
-};
-
-/**
- * Gets the token manager from a UiPath instance
- * @param instance - UiPath instance
- * @returns TokenManager object
- */
-export const getTokenManager = (instance: UiPath): TokenManager => {
-  return SDKInternalsRegistry.get(instance).tokenManager;
-};
+interface MockableTokenManager {
+  getToken: () => string | undefined;
+  hasValidToken: () => boolean;
+}
 
 // Mock console methods to avoid test output noise
 global.console = {
@@ -63,10 +41,8 @@ import { TEST_CONSTANTS } from './constants/common';
 
 /**
  * Creates a mock UiPathConfig object for testing
- * @param overrides - Optional overrides for specific config values
- * @returns Mock UiPathConfig object
  */
-export const createMockConfig = (overrides?: Partial<UiPathConfig>): UiPathConfig => {
+const createMockConfig = (overrides?: Partial<UiPathConfig>): UiPathConfig => {
   return new UiPathConfig({
     baseUrl: TEST_CONSTANTS.BASE_URL,
     orgName: TEST_CONSTANTS.ORGANIZATION_ID,
@@ -79,23 +55,21 @@ export const createMockConfig = (overrides?: Partial<UiPathConfig>): UiPathConfi
 
 /**
  * Creates a mock ExecutionContext for testing
- * @returns Mock ExecutionContext instance
  */
-export const createMockExecutionContext = (): ExecutionContext => {
+const createMockExecutionContext = (): ExecutionContext => {
   return new ExecutionContext();
 };
 
 /**
  * Creates a mock TokenManager for testing
- * @param overrides - Optional overrides for specific methods
- * @returns Mock TokenManager object
  */
-export const createMockTokenManager = (overrides?: Partial<TokenManager>): TokenManager => {
-  return {
+const createMockTokenManager = (overrides?: Partial<MockableTokenManager>): TokenManager => {
+  const mock: MockableTokenManager = {
     getToken: vi.fn().mockReturnValue('mock-access-token'),
     hasValidToken: vi.fn().mockReturnValue(true),
     ...overrides,
-  } as unknown as TokenManager;
+  };
+  return mock as TokenManager;
 };
 
 /**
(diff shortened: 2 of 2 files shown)


src/core/registry/service-registry.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @example
+ * ```typescript
+ * const registry = new ServiceRegistry(config, context, tokenManager);
+ *
+ * // Auto-registers and creates on first access
+ * const entities = registry.get(EntityService);
+ *
+ * // Returns cached instance on subsequent calls
+ * const entities2 = registry.get(EntityService); // Same instance
```
ayush-uipath: If this is going to go as an example in documentation and consumed by LLMs - I would use a better name for the const, ex.

```suggestion
 * const anotherEntities = registry.get(EntityService); // Same instance
```

src/services/data-fabric/index.ts
```diff
@@ -1 +1,21 @@
...
+ *
+ * Provides access to UiPath Data Fabric services for entity management.
+ * @module
+ */
+
+// Export service with cleaner name and keep EntityService for legacy UiPath class
+export { EntityService as Entities, EntityService } from './entities';
+
+// Re-export types for convenience
+export type * from '../../models/data-fabric/entities.types';
```
ayush-uipath: Same here - we should use decorators instead of relative paths
```suggestion
export type * from '@models/data-fabric/entities.types';
```

src/core/registry/service-registry.ts
```diff
@@ -0,0 +1,138 @@
...
+ * ```
+ */
+
+import type { UiPathConfig } from '../config/config';
```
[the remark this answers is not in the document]
SunilYadav-uipath: @ayush-uipath Currently, the SDK does not have shortcut decorators set up, and we are using relative imports throughout the code. I will create a task to introduce decorators into our SDK.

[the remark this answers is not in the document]
ayush-uipath: Sure. That task should introduce decorators, and also fix all relative imports

Raina451:
> import { UiPathClient, Entities } from '@uipath/uipath-typescript/entities';

Why the client is bundled  in services module? auth/client concerns should have their own entry point for eg: `@uipath/uipath-typescript/auth` ?

```
@uipath/uipath-typescript/auth   -> Client, UiPathSDKConfig etc.
@uipath/uipath-typescript/entities -> Entities, entity types only
```

SunilYadav-uipath:
> Current approach bundles the entire core (UiPathClient, Authservice, tokenmanager, telemetry, config, errors) into each service module (/entities, /tasks, etc.), which is code duplication.
> 
> we should consider creating a separate `/auth` or `/core` entry point for shared code, and have service modules import from the same?

The reason I used:

import { UiPathClient, Entities } from '@uipath/uipath-typescript/entities';


was mainly to improve developer experience. This avoids making consumers write two separate imports (one for auth/client and one for the service) and keeps the onboarding friction low.

From a bundle perspective, the client/auth code is relatively small, and after gzip the additional size becomes negligible.

That said, I agree with the broader architecture concern — ideally, the client/auth layer should live in its own dedicated entry point, for example:

@uipath/uipath-typescript/auth     → UiPathClient, config, token mgmt, etc.
@uipath/uipath-typescript/entities → Entities + entity-related types only


This separation keeps concerns cleaner and avoids any risk of duplication if more services are added later.

I’m open to adjusting the import pattern if we want strict separation, but the choice in this PR was intentionally made for better DX.

src/services/data-fabric/index.ts
```diff
@@ -1 +1,32 @@
...
+export type * from '../../models/data-fabric/entities.models';
+
+// Re-export common utilities users might need
+export { UiPathError } from '../../core/errors';
+export type { PaginatedResponse, NonPaginatedResponse } from '../../utils/pagination';
+export type { UiPathSDKConfig } from '../../core/config/sdk-config';
+export type { UiPath } from '../../core/uipath';
```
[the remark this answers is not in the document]
swati354: Yes please, lets keep these separate. Same for UiPathError? It is also getting exported from core.

swati354:
Also lets not merge this to main directly. We could create another branch, add all modularization related PRs to that and then merge that to main.

swati354:
Instead of bundling utilities like `error` and `pagination` with each service, shouldn't we keep them as part of the core module?
We should then externalize the core module in the rollup config so that it is not bundled with each service.

swati354:
@SunilYadav-uipath  I see you have asked for a review again but a lot of previous comments are still unaddressed. Some form of reaction or a brief reply to those comments would help understand the constraints and reasoning before reviewing further.

src/uipath.ts
```diff
@@ -14,147 +14,31 @@ import {
...
-  private executionContext: ExecutionContext;
-  private authService: AuthService;
-  private initialized: boolean = false;
+/**
+ * UiPath SDK - Legacy class providing all services through property getters.
+ *
+ * Extends UiPathClient. For modular usage, use UiPathClient.connect() with .get() instead.
+ */
+export class UiPath extends UiPathClient {
   private readonly _services: Map<string, any> = new Map();
```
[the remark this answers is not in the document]
SunilYadav-uipath: there will be one registry which is for backward compatible structure, for new modular implementation we don't need the registry

rollup.config.js
```diff
@@ -122,5 +122,77 @@ const configs = [
...
+  },
+  {
+    name: 'entities',
+    input: 'src/services/data-fabric/index.ts',
+    output: 'entities/index'
+  }
+];
+
+// Generate build configs for each service entry
+serviceEntries.forEach(({ input, output }) => {
```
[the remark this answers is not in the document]
Raina451: type import is fine, but
/entities is importing base.ts which imports from core/* (ApiClient, Config, ExecutionContext, etc)

The issue is: When Rollup builds /entities and other services, it follows all imports transitively. Since base.ts imports from core/*, and those core modules import more core modules, eventually the entire core gets bundled.

Please check the entities bundle after built you will see core code gets bundeld

src/services/orchestrator/buckets.ts
```diff
@@ -32,13 +32,15 @@ import { track } from '../../core/telemetry';
...
+   * Creates an instance of the Buckets service.
+   *
+   * @param uipathInstance - UiPath SDK instance providing authentication and configuration
    */
-  constructor(config: Config, executionContext: ExecutionContext, tokenManager: TokenManager) {
-    super(config, executionContext, tokenManager);
-    this.tokenManager = tokenManager;
+  constructor(uipathInstance: UiPath) {
+    super(uipathInstance);
+    this.tokenManager = uipathInstance[__PRIVATE__].tokenManager;
```
Raina451: earlier tokenManager was passed as a constructor parameter and BucketService didn't know or care where it came from. Now it directly accesses __PRIVATE__, coupling it to the internal Symbol pattern.
so If the internal pattern changes, both BaseService and BucketService need updates. The __PRIVATE__ access should be isolated to one place. 
Currently BaseService already accesses __PRIVATE__ to extract config, context, tokenManager - but doesn't expose tokenManager. check if we should add it

rollup.config.js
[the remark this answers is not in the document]
Raina451: type import is fine, but `/entities` is importing base.ts which imports from `core/*` (ApiClient, Config, ExecutionContext, etc)

The issue is: When Rollup builds /entities and other services, it follows all imports transitively. Since base.ts imports from core/*, and those core modules import more core modules, eventually the entire core gets bundled.

Please check the entities bundle after built you will see core code gets bundeld

tests/unit/services/action-center/tasks.test.ts
```diff
@@ -47,7 +47,7 @@ describe('TaskService Unit Tests', () => {
 
   beforeEach(() => {
     // Create mock instances using centralized setup
-    const { config, executionContext, tokenManager } = createServiceTestDependencies();
+    const { uiPath } = createServiceTestDependencies();
```
Raina451: * `uipathInstance`

src/core/internals/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  /** Configuration including base URL, organization name, and tenant name */
+  config: UiPathConfig;
+  /** Execution context for request tracking and metadata */
+  context: ExecutionContext;
+  /** Token manager for authentication */
+  tokenManager: TokenManager;
+}
+
+// Re-export SDKClientConfig as PrivateSDK alias for compatibility
+export type { PrivateSDK as SDKClientConfig };
```
swati354: Why do we need to re-export it here? What is the compatibility for?

src/core/uipath.ts
```diff
@@ -0,0 +1,186 @@
...
+ * ```
+ */
+export class UiPath {
+  // Private fields - true runtime privacy, not visible via Object.keys()
+  #config: UiPathConfig;
+  #authService: AuthService;
+  #initialized: boolean = false;
+
+  /** Read-only config for user convenience */
+  public readonly config: Readonly<{ baseUrl: string; orgName: string; tenantName: string }>;
```
swati354: Can we type this to [BaseConfig](https://github.com/UiPath/uipath-typescript/blob/147805bef56cff4ce44b0e8418932a0410963365/src/core/config/sdk-config.ts#L2) ?

src/core/internals/registry.ts
```diff
@@ -0,0 +1,77 @@
...
+  static has(instance: object): boolean {
+    return this.store.has(instance);
+  }
+
+  /**
+   * Remove an instance from registry (optional cleanup)
+   */
+  static delete(instance: object): boolean {
+    return this.store.delete(instance);
+  }
```
Raina451: are we using this anywhere? do we need these?

PR #150, opened 2026-01-09
Rename INSERT_BY_ID to BATCH_INSERT_BY_ID [PLT-94710]

Sarath1018:
Renaming INSERT_BY_ID to BATCH_INSERT_BY_ID as this method is using insert-batch API 
Will implement a new method INSERT_BY_ID which will use insert API

### DIFF
--- a/src/models/data-fabric/entities.models.ts
+++ b/src/models/data-fabric/entities.models.ts
@@ -108,7 +108,7 @@ export interface EntityServiceModel {
   >;
 
   /**
-   * Inserts data into an entity by entity ID using batch insert
+   * Inserts one or more records into an entity by entity ID using batch insert
    *
    * @param id - UUID of the entity
    * @param data - Array of records to insert


swati354:
> Renaming INSERT_BY_ID to BATCH_INSERT_BY_ID as this method is using insert-batch API Will implement a new method INSERT_BY_ID which will use insert API

ultimately we'll only expose `insert_by_id` which will internally call this? or are we going to expose different methods

swati354 (approved):

src/models/data-fabric/entities.models.ts
```diff
@@ -108,8 +108,8 @@ export interface EntityServiceModel {
   >;
 
   /**
-   * Inserts data into an entity by entity ID
-   * 
+   * Inserts data into an entity by entity ID using batch insert
```
Raina451: have a better info here.

src/models/data-fabric/entities.models.ts
```diff
@@ -133,7 +133,7 @@ export interface EntityServiceModel {
    * });
    * ```
    */
-  insertById(id: string, data: Record<string, any>[], options?: EntityInsertOptions): Promise<EntityInsertResponse>;
+  batchInsertById(id: string, data: Record<string, any>[], options?: EntityInsertOptions): Promise<EntityInsertResponse>;
```
Raina451: we should also inform user by docs, what batch size is default, and if they can customise the batch size. check if there is any param in api which provides batch size option

Sarath1018: Checked the API swagger, there's no batch size parameter exposed by the API.

Raina451 (commented):
please ask data fabric team about default batchsize, and document the same in next pr.

PR #157, opened 2026-01-16
feat: pack cmd changes to support externalize config - [PLT-95750]

Raina451:
Updates the `uipath pack` cmd to handle sdk configuration (`uipath.json`)  for the externalized config.                                            
                                                                                                     
                                                                                 
  ## Changes                                                                     
                                                                                 
  ### During `uipath pack` command:                                              
                                                                                 
  1. **If `uipath.json` exists:**                                                
     - Validate that `scope` field is present (required)                         
     - If `clientId` exists in the config, prompt the user:                      
       > "Do you want to reuse this clientId or let UiPath create one?"          
       - If user wants to **reuse**: Copy `uipath.json` to dist folder as-is     
       - If user wants **UiPath to create one**: Copy `uipath.json` to dist      
  folder with `clientId` cleared (empty string)                                  
     - If `--reuse-client` flag is provided, skip the prompt and always reuse    (non interactive)
                                                                                 
  2. **If `uipath.json` does not exist:**                                        
     - Prompt user to enter required scopes:                                     
       > "Enter the required scopes for your app (e.g., OR.Execution OR.Folders)"
     - Create `uipath.json` in the project root with:                            
       - `scope`: user-provided value                                            
       - `clientId`, `orgName`, `tenantName`, `baseUrl`, `redirectUri`: empty    
  strings (placeholders)                                                         
     - Copy the created config to dist folder  to get packaged   
     
                                  
  Scenario: Version upgrade (v2.0.0+) + clientId exists                       
  Behavior: No prompt - automatically reuse (publish API handles it)          
  ────────────────────────────────────────                                                             
  Scenario: No clientId                                                       
  Behavior: No prompt - UiPath creates new client                             
  The info message about scope also only shows for new apps, not version      
  upgrades.


packages/cli/src/commands/pack.ts
```diff
@@ -196,16 +221,71 @@ export default class Pack extends Command {
...
+        this.log(chalk.red(MESSAGES.ERRORS.CONFIG_FILE_MISSING_SCOPE));
+        this.log(chalk.dim('  The scope field is required for OAuth client creation during deployment.'));
+        return null;
+      }
+
+      return config;
+    }
+
+    // uipath.json doesn't exist - prompt for scopes and create it
+    this.log(chalk.yellow(`⚠️  ${CONFIG_FILE_NAME} not found in project root.`));
```
swati354: message constant.

packages/cli/src/commands/pack.ts
```diff
@@ -242,14 +322,29 @@ export default class Pack extends Command {
...
+        // Clear clientId so UiPath creates a new OAuth client during deployment
+        configToWrite.clientId = '';
+        this.log(chalk.green(MESSAGES.SUCCESS.CLIENT_ID_CLEARED));
+      } else if (config.sdkConfig.clientId) {
+        this.log(chalk.green(MESSAGES.SUCCESS.CLIENT_ID_REUSED));
+      }
+
+      fs.writeFileSync(configDest, JSON.stringify(configToWrite, null, 2));
+      this.log(chalk.green(MESSAGES.SUCCESS.CONFIG_FILE_INCLUDED));
+
```
swati354: Should we extract this into a method ?

packages/cli/src/constants/messages.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * Config file name constant
+ */
+export const CONFIG_FILE_NAME = 'uipath.json';
```
swati354: perhaps put this in `constants/auth.ts` under `Files`?

packages/cli/src/constants/messages.ts
```diff
@@ -167,6 +179,8 @@ export const MESSAGES = {
     CONTINUE_WITH_DIFFERENT_VALUES: 'Do you want to continue with these different values?',
     COMPLETE_AUTH_IN_BROWSER: 'Please complete the authentication in your browser',
     BROWSER_FALLBACK_INSTRUCTION: 'If the browser didn\'t open automatically, visit:',
+    REUSE_CLIENT_ID: 'Do you want to reuse the existing clientId from uipath.json in production or let UiPath create one?',
```
swati354: Isn't the prompt slightly confusing? 
If I say yes to this, I am not completely sure what is going to happen i.e., are we reusing or creating a new one.
It should be clear what yes would do and what no would do
```
"Yes" = reuse existing clientId
"No" = let UiPath create a new one
```

Raina451:
> Why are scopes required even when user provides a client id?

because scope is required by sdk config, although if clientid is provided technically we can get the scopes and set that by default.
Let me update the logic.

PR #166, opened 2026-01-28
Merge feature/insert-by-id into main

Sarath1018:


src/models/data-fabric/entities.models.ts
```diff
@@ -115,23 +120,50 @@ export interface EntityServiceModel {
...
+   * @returns Promise resolving to the inserted record with generated record ID
+   * {@link EntityInsertResponse}
+   * @example
+   * ```typescript
+   * // Basic usage
+   * const result = await sdk.entities.insertById(<entityId>, { name: "John", age: 30 });
+   *
+   * // With options
+   * const result = await sdk.entities.insertById(<entityId>, { name: "John", age: 30 }, {
+   *   expansionLevel: 1
```
Raina451: can you also add what happens with expansion level 1?

Raina451: also can we make it clear that we can only insert one record from this method?

PR #170, opened 2026-01-29
feat: added pr check for develop and main

viswa-uipath:
Added pr check file to run test and build when a PR is raised

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,0 +1,59 @@
+name: PR Checks
+
+on:
+  pull_request:
+    branches:
+      - main
+      - develop
+    types:
+      - opened           # When PR is first created
+      - synchronize      # When new commits are pushed to PR branch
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+          cache: 'npm'
+
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Run Tests
+        run: npm test
```
Raina451: this by defaults runs in watch mode, should we run in non watch mode? 
`npm test -- --run`

Raina451: also should we add coverage as well? 
`npm run test:coverage -- --run`

Raina451 (commented):
can we also catch in pr check that whenever package.json is updated, the corresponding package-lock.json changes are included in the same PR?

PR #171, opened 2026-01-29
Sort methods of each service in alphabatical order [PLT-94050]

Sarath1018:
Sorted the methods of all services in the documentation in alphabatical order.

For example methods of entity:
image


Raina451 (approved):
run tests once.

swati354 (approved):

PR #181, opened 2026-02-05
chore:update directory

deepeshrai-tech:
Updated remote folder directory from source to source/buildDir


packages/cli/src/core/webapp-file-handler/folder-ops.ts
```diff
@@ -21,13 +22,15 @@ export async function ensureContentRootExists(
   config: WebAppPushConfig,
   lockKey: string | null,
   getStructure: () => ProjectStructure | null,
-  setStructure: (s: ProjectStructure) => void
+  setStructure: (s: ProjectStructure) => void,
+  bundlePath: string
```
swati354: can `bundlePath` be nested ? i.e., something like `dist/abc` ?

Raina451 (approved):

PR #193, opened 2026-02-09
feat: push metadata

deepeshrai-tech:
Associated Jira Ticket- PLT-96330

## Summary

Introduces `push_metadata.json` as the push manifest , stores it at `.uipath/push_metadata.json` locally and `source/push_metadata.json` on the remote, and adds **schemaVersion** semantics so the version is bumped consistently with the push plan (major on removals, minor on add/update).

### Push metadata file and flow

- **Local path**: `.uipath/push_metadata.json` (project root)
- **Remote path**: `source/push_metadata.json`
- `prepareMetadataFileForPlan` creates/updates the local file; `uploadPushMetadataToRemote` uploads it after the push plan is prepared

### Author email

- **`getPushAuthorEmail(accessToken?, logger?)`** reads the push author from the access token JWT `email` claim only 

### schemaVersion behavior

- **First push**: `schemaVersion` is always **1.0.0** when the metadata file is created (no existing local or remote).
- **Remote as source of truth**: If the local `.uipath/push_metadata.json` is missing but the file exists on the remote, we download it and use it as the base, then apply `lastPushDate`, `email`, and version bump from the plan. Only when both are missing do we create new metadata with 1.0.0.
- **Version bump rules** (aligned with the plan log: add / update / delete / folders to delete):
  - **Major** (e.g. 1.0.0 → 2.0.0): when something is **removed** — `deleteFiles.length > 0` or `deleteFolders.length > 0`
  - **Minor** (e.g. 1.0.0 → 1.1.0): when there are **adds or updates** and no removals — `uploadFiles.length > 0` or `updateFiles.length > 0`
  - **Unchanged**: no add/update and no delete
- `createFolders` is **not** used for schemaVersion so content-only updates (e.g. “0 add, 1 update, 0 delete, 0 folders to delete”) correctly produce a **minor** bump.


packages/cli/src/core/webapp-file-handler/handler.ts
```diff
@@ -118,10 +119,19 @@ export class WebAppFileHandler {
...
       this.config,
       fullRemoteFiles,
-      (config, fileId) => api.downloadRemoteFile(config, fileId)
+      (config, fileId) => api.downloadRemoteFile(config, fileId),
+      plan
     );
 
     const folderIdMap = buildFolderIdMap(this.projectStructure!);
+    const metadataPath = path.join(this.config.rootDir, this.config.manifestFile);
+    await uploadPushMetadataToRemote(
```
swati354: 1. shouldn't we upload the metadata file after all the file operations? What if some operations fail, we'll have inconsistent `schemaVersion`.
2. a log before the upload?
`this.config.logger.log(chalk.gray('[push] Uploading metadata to remote...'));`

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -1,60 +1,192 @@
...
+
+/** Fallback codeVersion when remote metadata cannot be read or does not contain codeVersion. */
+const DEFAULT_CODE_VERSION = '0.1.1';
+
+/** Max length of error message sent to telemetry to avoid oversized payloads. */
+const MAX_TELEMETRY_ERROR_LENGTH = 500;
 
 export function getCurrentUser(): string {
   return process.env.USER || process.env.USERNAME || process.env.UIPATH_USER || 'unknown';
 }
```
swati354: not being used anymore, remove.

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -1,60 +1,192 @@
...
-        lastPushAuthor: getCurrentUser(),
-      };
+    const err = error as NodeJS.ErrnoException & { name?: string };
+    if (err.code === 'ENOENT') {
+      // Local file missing: use remote as source of truth when present.
+      const remoteEntry =
+        remoteFiles.get(PUSH_METADATA_REMOTE_PATH) ||
+        remoteFiles.get(PUSH_METADATA_RELATIVE_PATH) ||
+        remoteFiles.get(PUSH_METADATA_FILENAME);
+      if (remoteEntry) {
```
swati354: why are we checking all these paths? Is this necessary?

packages/cli/src/core/webapp-file-handler/push-plan.ts
```diff
@@ -98,8 +98,8 @@ export async function computeExecutionPlan(
 
   for (const [filePath, remoteFile] of remoteFiles.entries()) {
     if (
-      filePath === STUDIO_METADATA_RELATIVE_PATH ||
-      filePath === STUDIO_METADATA_FILENAME
+      filePath === PUSH_METADATA_RELATIVE_PATH ||
+      filePath === PUSH_METADATA_FILENAME
```
swati354: shouldn't we check for `PUSH_METADATA_REMOTE_PATH` here?

packages/cli/src/core/webapp-file-handler/push-plan.ts
```diff
@@ -110,15 +110,26 @@ export async function computeExecutionPlan(
...
+    const normalized = folderPath.replace(/\\/g, '/').toLowerCase();
+    if (!remoteFoldersByNormalized.has(normalized)) {
       plan.createFolders.push({ path: folderPath });
     }
   }
 
   const contentRootPrefix = remoteContentRoot + '/';
   for (const [folderPath, folder] of remoteFolders.entries()) {
-    if (requiredFolders.has(folderPath)) continue;
+    const normalizedFolder = folderPath.replace(/\\/g, '/').toLowerCase();
```
swati354: Path normalization logic is repeated multiple times in the file. Consider extracting to a helper method.

packages/cli/src/core/webapp-file-handler/handler.ts
```diff
@@ -179,6 +181,16 @@ export class WebAppFileHandler {
...
     );
+
+    this.config.logger.log(chalk.gray('[push] Uploading metadata to remote...'));
+    const metadataPath = path.join(this.config.rootDir, this.config.manifestFile);
+    await uploadPushMetadataToRemote(
+      this.config,
+      metadataPath,
+      fullRemoteFiles,
+      folderIdMap,
+      this.lockKey
```
Raina451: uploadPushMetadataToRemote has no error handling. should a metadata upload failure be fatal? If yes, consider moving it after file ops so failure doesn't block the push. If no, wrap in try/catch.

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -1,87 +1,254 @@
...
+ * Source of truth: if local file is missing, we use remote push_metadata.json when present;
+ * only when both are missing do we create new metadata with schemaVersion 1.0.0.
+ * Updates lastPushDate, lastPushAuthor, and schemaVersion (from plan) and optionally codeVersion from remote.
+ */
 export async function prepareMetadataFileForPlan(
   config: WebAppPushConfig,
   remoteFiles: Map<string, ProjectFile>,
-  downloadRemoteFile: (config: WebAppPushConfig, fileId: string) => Promise<Buffer>
+  downloadRemoteFile: (config: WebAppPushConfig, fileId: string) => Promise<Buffer>,
+  plan?: FileOperationPlan
```
Raina451: why optional?

packages/cli/src/core/webapp-file-handler/push-plan.ts
```diff
@@ -98,29 +103,40 @@ export async function computeExecutionPlan(
...
+      filePath === PUSH_METADATA_REMOTE_PATH ||
+      filePath === PUSH_METADATA_RELATIVE_PATH ||
+      filePath === PUSH_METADATA_FILENAME
     ) {
       continue;
     }
     if (processedFileIds.has(remoteFile.id)) continue;
-    const normalizedRemote = filePath.replace(/\\/g, '/');
+    const normalizedRemote = normalizePathForComparison(filePath);
     if (!localRemotePaths.has(normalizedRemote)) {
```
Raina451: are we storing in localRemotePaths normalisedpaths too?

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -1,87 +1,254 @@
...
+  return {
+    schemaVersion: DEFAULT_SCHEMA_VERSION,
+    projectId: config.projectId,
+    description: '',
+    lastPushDate: new Date().toISOString(),
+    lastPushAuthor: getPushAuthorEmail(config.envConfig.accessToken, config.logger),
+  };
 }
 
+/** Applies codeVersion from parsed remote metadata (bump patch or set default). Used to avoid duplicate download. */
```
Raina451: what is codeversion? same as app version?

deepeshrai-tech: codeVersion was created to handle backward compatibility but we decided its wont be needed anymore so will remove its usage all togehter

packages/cli/src/core/webapp-file-handler/push-plan.ts
deepeshrai-tech: we’re not storing normalized paths in localRemotePaths. We’re storing the raw remotePath from localPathToRemotePath. We only normalize when we look up

PR #198, opened 2026-02-10
Doc update

Sarath1018:
Restructured Entities documentation to use MkDocs Material's navigation index feature
Added Choice Sets as a sub-item under Entities in the documentation navigation
Updated contributing guidelines to reference develop branch

image
image


swati354 (approved):

Raina451 (changes requested):
from screenshot, when entities is clicked, the heading is showing as "index" , can we fix that?

PR #201, opened 2026-02-10
fix: post process cross-platform compatible

vnaren23:
Previous command was Unix commands, moving to cross platform compatible scripts


Raina451 (approved):
FYI, we want to get rid of docs:post-process, there should be a better way in typedoc/mkdocs to acheieve the same.
@Sarath1018 please create a ticket to track the same

PR #205, opened 2026-02-10
chore: set up oxlint with recommended defaults

ruudandriessen:
## Summary
- Add oxlint configuration with `correctness` and `suspicious` rule categories enabled at deny level
- Install oxlint as devDependency and add `lint` script to package.json
- Fix 46 linting violations across the codebase (unused variables/imports, redundant constructors, escape sequences, etc)

## Details
The oxlint setup provides a solid foundation for code quality without being overly strict. Rules are disabled for intentional architectural patterns (static-only utility classes, async Express handlers) and incompatibilities (Error cause syntax requires ES2022+ but project targets ES2020).

All tests pass (341/341), build succeeds cleanly, and oxlint reports 0 warnings and 0 errors.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -32,6 +32,9 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Typecheck
+        run: npm run typecheck
+
       - name: Run Tests
         run: npm run test:coverage -- --run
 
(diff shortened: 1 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
```
swati354: nit : could add `cache: 'npm'`, to speed up CI runs

ruudandriessen: if we want actual speed - we should just use bun instead tbh 😛  install with cache doesn't even improve the speed that much

samples/process-app/src/components/ProcessList.tsx
```diff
@@ -216,7 +216,7 @@ export const ProcessList = () => {
       ) : (
         <div className="space-y-3">
           {processes
-            .sort((a, b) => {
+            .toSorted((a, b) => {
```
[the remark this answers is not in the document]
ruudandriessen: I reverted this change for now - will make a separate PR to add typechecks for future 😄

PR #206, opened 2026-02-10
chore: replace axios with fetch for file uploads

ruudandriessen:
## Summary
Removes the unused axios dependency and replaces the single axios.put call with the native fetch API. All other HTTP calls in the SDK already use fetch via ApiClient, completing the migration to native HTTP clients.

## Changes
- Replaced `axios.put()` with `fetch()` in BucketService
- Updated tests to mock `fetch` instead of axios
- Removed axios from dependencies

## Test Plan
- All 341 unit tests pass
- All 28 bucket service tests pass, including upload scenarios with Buffer, Blob, and token refresh

🤖 Generated with Claude Code


tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -388,17 +388,19 @@ describe('BucketService Unit Tests', () => {
...
-          headers: expect.any(Object) // Headers should be empty object since Keys/Values are empty
+          method: 'PUT',
+          body: bufferContent,
+          headers: expect.any(Object)
         })
       );
 
       // Verify the success logic: status >= 200 && status < 300
       expect(result.success).toBe(true); // 201 is in range [200, 300)
+      mockFetch.mockRestore();
```
Raina451: mockFetch.mockRestore() won't run if the test fails early. Move cleanup to afterEach /beforeeach

PR #208, opened 2026-02-11
Add typecheck to CI pipeline

ruudandriessen:
## Summary
Add TypeScript type checking as an explicit step in the PR checks workflow. This catches type errors earlier in the CI pipeline and ensures code quality before tests and build.

## Changes
- Added `typecheck` npm script using `tsc --noEmit`
- Added typecheck step to PR checks workflow before test execution
- Updated CI summary to reflect typecheck status

🤖 Generated with Claude Code

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -32,6 +32,9 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Typecheck
+        run: npm run typecheck
+
       - name: Run Tests
         run: npm run test:coverage -- --run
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Typecheck
+        run: npm run typecheck
```
Raina451: @rollup/plugin-typescript already typechecks during the build step. Just want to confirm this adds earlier failure signal and a cleaner developer experience, is that the intent?

Sarath1018 (approved):

.github/workflows/pr-checks.yml
swati354: > I think my previous PR didn't fail since it was targeted at main instead of develop.

No, it was because `samples` are not part of the SDK build.

PR #217, opened 2026-02-15
feat: update push command to incude source code as welll

deepeshrai-tech:
Associated Jira Ticket- PLT-96945

# Summary

Extends the web app push flow with source code, remote metadata, project-root manifest updates, and non-fatal handling of create conflicts so push completes successfully

### 1. Push metadata (`push_metadata.json`)

- **Local**: Push metadata is written under `.uipath/push_metadata.json` (schema version, project id, last push date/author, **buildDir**).
- **Remote**: Uploaded to `source/push_metadata.json` (not under the build dir). Created or updated as part of push; uses same diff/plan as other files.
- **buildDir**: Stored from `--build-dir` / config so the current build output folder (e.g. `dist`, `output`) is recorded and can be used by the platform.

### 2. Remote web app manifest (`webAppManifest.json`)

- **Location**: The manifest lives at **project root** on the remote (not under `source/`).
- **Behavior**: When the remote has `webAppManifest.json`, push updates it so `config.bundlePath` is set to `"source/<buildDirName>"` (e.g. `"source/dist"`), keeping existing `type` and `config.isCompiled` intact.
- **Failure**: If the file is missing or update fails, the error is logged and push still succeeds.

## Push scope 
- **Build output**: Files under the build dir (e.g. `dist/`) are pushed to `source/<buildDir>/...`.
- **Source files**: Other project files (excluding build dir and ignore patterns) are pushed under `source/...` with the same plan (add/update/delete).
- **Exclusions**: Git-like patterns (e.g. `node_modules/`, `.uipath/`, `.env*`, `.gitignore`) and dotfiles remain excluded via `constants/push.ts` and the local file collector.

### DIFF
--- a/packages/cli/src/core/webapp-file-handler/folder-ops.ts
+++ b/packages/cli/src/core/webapp-file-handler/folder-ops.ts
@@ -11,8 +11,14 @@ import {
 } from './structure.js';
 import * as api from './api.js';
 
-/** Returns true if folderPath exists in the map (by exact or normalized key). Backend may return different casing. */
-function hasFolderByPath(foldersMap: Map<string, ProjectFolder>, folderPath: string): boolean {
+/**
+ * Returns true if folderPath exists in the map (by exact or normalized key).
+ * Used for "folder exists on remote" checks; backend may return different casing (e.g. Source vs source).
+ */
+export function hasFolderByPath(
+  foldersMap: Map<string, ProjectFolder>,
+  folderPath: string
+): boolean {
   const norm = normalizeFolderPath(folderPath);
   return (
     foldersMap.has(folderPath) ||
@@ -118,8 +124,7 @@ function resolveFolderIdAfterCreate(
   const pathId = byPath?.id;
   if (pathId) return pathId;
   const byName = afterCreate.get(folderName);
-  const nameId = byName?.id;
-  return nameId ?? undefined;
+  return byName?.id ?? undefined;
 }
 
 /**
@@ -173,8 +178,8 @@ export async function ensureFoldersCreated(
       if (id) folderIdMap.set(norm(folder.path), id);
     }
 
-    const hadNull = results.some((r) => !r.id);
-    if (hadNull) {
+    const someFolderCreatesReturnedNull = results.some((r) => !r.id);
+    if (someFolderCreatesReturnedNull) {
       cliTelemetryClient.track('Cli.Push.FolderCreateConflict', {
         message: 'One or more create folder returned null; refetching to resolve parent ids',
         depth: String(depth),
@@ -188,9 +193,9 @@ export async function ensureFoldersCreated(
     }
   }
 
-  const next = await api.fetchRemoteStructure(config);
-  setStructure(next);
-  const afterCreate = getRemoteFoldersMap(next);
+  const structureAfterFolderCreation = await api.fetchRemoteStructure(config);
+  setStructure(structureAfterFolderCreation);
+  const afterCreate = getRemoteFoldersMap(structureAfterFolderCreation);
   for (const folder of plan.createFolders) {
     if (folderIdMap.has(norm(folder.path))) continue;
     const folderName = folder.path.split('/').filter(Boolean).pop()!;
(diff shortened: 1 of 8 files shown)


packages/cli/src/core/webapp-file-handler/file-ops.ts
```diff
@@ -3,15 +3,14 @@ import { MESSAGES } from '../../constants/index.js';
 import type { WebAppPushConfig, FileOperationPlan, FileOpsResult } from './types.js';
 import * as api from './api.js';
 
-/** Max concurrent file upload/update requests per batch. Keeps push fast without overloading the server or hitting rate limits. */
-const WORKER_POOL_SIZE = 8;
```
swati354: We can still keep the batching right? Why remove it?
We also had a progress logging because of this, which is now removed.

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -206,3 +213,49 @@ export async function uploadPushMetadataToRemote(
...
+  lockKey: string | null
+): Promise<void> {
+  const remoteFile = fullRemoteFiles.get(WEB_APP_MANIFEST_FILENAME);
+  if (!remoteFile) return;
+
+  try {
+    const content = await api.downloadRemoteFile(config, remoteFile.id);
+    const parsed = JSON.parse(content.toString('utf-8')) as Record<string, unknown>;
+    if (typeof parsed !== 'object' || parsed === null) return;
+    const buildDirName = bundlePath.replace(/\\/g, '/').replace(/^\/+|\/+$/g, '') || bundlePath;
```
swati354: can we use 
https://github.com/UiPath/uipath-typescript/blob/f880763e3b703c9ced7dffc69eb820d207f61b38/packages/cli/src/core/webapp-file-handler/structure.ts#L18

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -206,3 +213,49 @@ export async function uploadPushMetadataToRemote(
...
+      content: newContent,
+    };
+    await api.updateFile(
+      config,
+      WEB_APP_MANIFEST_FILENAME,
+      localFile,
+      remoteFile.id,
+      lockKey
+    );
+  } catch (err) {
```
swati354: if there is try catch here already then in the `handler.ts` where this method is called, the handler's catch is dead code

packages/cli/src/core/webapp-file-handler/folder-ops.ts
```diff
@@ -31,95 +58,145 @@ export async function ensureContentRootExists(
...
+      const afterCreate = await refetchAndMergeFolderIds(config, folderIdMap, setStructure);
+      for (const { folder, folderName } of results) {
+        if (folderIdMap.has(norm(folder.path))) continue;
+        const resolvedId = resolveFolderIdAfterCreate(afterCreate, folder.path, folderName);
+        if (resolvedId) folderIdMap.set(norm(folder.path), resolvedId);
+      }
     }
   }
+
+  const next = await api.fetchRemoteStructure(config);
```
NishankSiddharth: same here, let's use better variable naming which are self explainatory

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -206,3 +213,49 @@ export async function uploadPushMetadataToRemote(
...
+  bundlePath: string,
+  fullRemoteFiles: Map<string, ProjectFile>,
+  lockKey: string | null
+): Promise<void> {
+  const remoteFile = fullRemoteFiles.get(WEB_APP_MANIFEST_FILENAME);
+  if (!remoteFile) return;
+
+  try {
+    const content = await api.downloadRemoteFile(config, remoteFile.id);
+    const parsed = JSON.parse(content.toString('utf-8')) as Record<string, unknown>;
```
NishankSiddharth: do we need 'utf-8' explicitly?

Raina451 (commented):
@deepeshrai-tech Is there a size limit on the remote? pushing whole source code can be quite large

PR #218, opened 2026-02-16
fix: migrate npm publish to trusted publishing

swati354:
npm [revoked all classic tokens](https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/). This PR migrates to [npm trusted publishing](https://docs.npmjs.com/trusted-publishers/), eliminating the need for any stored npm tokens.

- Added `id-token: write` permission ( required for GitHub Actions to generate the OIDC token npm uses for auth )

- Upgraded Node from 18 to 24 ( trusted publishing requires npm CLI 11.5.1+, which ships with Node 24 natively )

- Added `--provenance` to npm publish ( signs the package with a verifiable link back to this repo and workflow )

- Removed Configure Git step ( nothing in the workflow does git commits )

- Removed npm pack step ( npm publish packs internally, artifact wasn't being used )

- Fixed Generate Summary ( replaced undefined `$NEW_VERSION` / `$CURRENT_VERSION` vars with actual version from package.json

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -22,7 +22,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '24'
-          registry-url: 'https://registry.npmjs.org'
 
       - name: Install Dependencies
         run: npm ci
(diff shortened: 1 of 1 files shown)


vnaren23 (approved):

.github/workflows/publish.yml
```diff
@@ -54,40 +51,27 @@ jobs:
       - name: Build
         run: npm run build
 
-      - name: Pack
-        run: npm pack
-
-      - name: publish to github packages
-        run: npm publish
+      - name: Publish to GitHub Packages
+        run: npm publish --registry=https://npm.pkg.github.com
```
Raina451: `.npmrc` auth is scoped to `registry.npmjs.org`, not `npm.pkg.github.com`. You need a second `setup-node` step here i belive with `registry-url:
  'https://npm.pkg.github.com'` before this step. See GitHubs docs [publishing to multiple
  registries](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages).

PR #229, opened 2026-02-16
Task events service for coded escalation apps

Sandeepan-Ghosh-0312:
## Description
- Added new package coded-action-apps which developers will call using new CodedActionAppsService()
- This service handles bidirectional communication between Action Center and coded app loaded from 'host' domain.
- External client authentication is taken care of by Action Center. Token (including refresh) is sent to the app through an event.
- When an sdk api call is made, it checks the expiry time of current token which if not found (or less than current time) it requests for a token from Action Center, receives it and updates token along with expiry time in TokenManager.

https://github.com/user-attachments/assets/7c67088d-89b0-483c-9242-eda6b276ace2

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/packages/coded-action-apps/README.md
+++ b/packages/coded-action-apps/README.md
@@ -1,0 +1,184 @@
+# @uipath/coded-action-apps
+
+An SDK enabling coded apps to be the UI for tasks within UiPath Action Center. SDK handles bi-directional communication between the coded app and UiPath Action Center host using window.postMessage events.
+
+## Installation
+
+```bash
+npm install @uipath/coded-action-apps
+```
+
+## Overview
+
+Action Center renders a coded action app within an iframe. @uipath/coded-action-apps provides service CodedActionAppsService which offers below capabilities:
+
+- **Receive** - `getTask()` - On app load, UiPath Action Center provides the task details.
+- **Notify** - `setTaskData()` - Notify Action Center when task data changes (e.g. to enable the Save button).
+- **Complete** - `completeTask()` - Completes the task when user clicks on submit buttons (e.g. 'Approve' or 'Reject' buttons)
+- **Display** - `showMessage()` - Displays toast messages inside Action Center.
+
+## Usage
+
+### Initialise the service
+
+```ts
+import { CodedActionAppsService } from '@uipath/coded-action-apps';
+
+const service = new CodedActionAppsService();
+```
+
+The class is also exported under the alias `CodedActionApps` for convenience:
+
+```ts
+import { CodedActionApps } from '@uipath/coded-action-apps';
+
+const service = new CodedActionApps();
+```
+
+---
+
+### Get task details from Action Center
+
+Call this once when the app loads. It sends an `AC.init` event to Action Center and waits for the task data to be posted back. (**Recommended**: Do not call it more than once in your app)
+
+```ts
+const taskData = await service.getTask();
+
+console.log(taskData.taskId);     // number
+console.log(taskData.title);      // string
+console.log(taskData.status);     // TaskStatus enum
+console.log(taskData.isReadOnly); // boolean
+console.log(taskData.data);       // the task's form data
+console.log(taskData.folderId);   // number
+console.log(taskData.folderName); // string
+console.log(taskData.theme);      // Theme enum — the UI theme Action Center is currently using
+```
+
+> **Note:** This call will reject with an error if Action Center does not respond within 3 seconds, or if the parent origin is not trusted.
+
+---
+
+### Notify Action Center of data changes
+
+Call this whenever the user modifies the task form data. Action Center uses this signal to enable its Save button.
+
+```ts
+service.setTaskData({ field: 'updatedValue' });
+```
+
+---
+
+### Complete a task
+
+Call this when the user submits the form to mark the task as complete. Provide the action taken (e.g. `"Approve"` or `"Reject"`) and the final data payload. The call returns a `Promise<TaskCompleteResponse>` that resolves once Action Center confirms the completion.
+
+```ts
+const result = await service.completeTask('Approve', { approved: true, notes: 'Looks good' });
+
+if (!result.success) {
+  console.error(result.errorMessage);
+}
+```
+
+> **Note:** Only one `completeTask` call may be in flight at a time. Calling it again before the previous call resolves will throw an error: `"A completeTask call is already in progress"`.
+
+---
+
+### Display a message in Action Center
+
+Show a toast notification inside Action Center using one of the four severity levels.
+
+```ts
+import { MessageSeverity } from '@uipath/coded-action-apps';
+
+service.showMessage('Validation successful', MessageSeverity.Success);
+service.showMessage('Validation failed', MessageSeverity.Error);
+service.showMessage('Please review the details', MessageSeverity.Warning);
+service.showMessage('User information', MessageSeverity.Info);
+```
+
+---
+
+## API Reference
+
+### `CodedActionAppsService`
+
+| Method | Description |
+|---|---|
+| `getTask()` | Returns a `Promise<Task>` with the current task's metadata and form data. |
+| `setTaskData(data)` | Sends updated data to Action Center to signal that the form has changed. |
+| `completeTask(actionTaken, data)` | Marks the task as complete with the given action label and final data. Returns `Promise<TaskCompleteResponse>`. Throws if a call is already in progress. |
+| `showMessage(msg, type)` | Displays a toast message in Action Center with the given severity. |
+
+---
+
+### `Task`
+
+```ts
+type Task = {
+  taskId: number;
+  title: string;
+  status: TaskStatus;
+  isReadOnly: boolean;
+  action: string | null;
+  data: unknown;
+  folderId: number;
+  folderName: string;
+  theme: Theme;
+};
+```
+
+---
+
+### `TaskStatus`
+
+```ts
+enum TaskStatus {
+  Unassigned = 'Unassigned',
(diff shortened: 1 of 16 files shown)


src/core/auth/token-manager.ts
```diff
@@ -3,6 +3,8 @@ import { isBrowser } from '../../utils/platform';
 import { AuthToken, TokenInfo } from './types';
 import { hasOAuthConfig } from '../config/sdk-config';
 import { Config } from '../config/config';
+import type { TaskEventsServiceInternal } from '../../services';
```
Raina451: this is creating a circular dependency

src/core/auth/service.ts
```diff
@@ -192,7 +193,7 @@ export class AuthService {
    */
   public authenticateWithSecret(secret: string): boolean {
     try {
-      this.updateToken({ token: secret, type: 'secret' });
+      this.updateToken({ token: isInActionCenter() ? '' : secret, type: 'secret' });
```
Raina451: lets not touch authenticateWithSecret() at all. Instead, TaskEventsServic should handle its own token initialization internally, it already has access to tokenManager via SDKInternalsRegistry, so it can call `tokenManager.setToken({ token: '', type: 'secret' })` directly in its own constructor/init flow. check this possibility.

Raina451: Same logic applies to the isInActionCenter() checks added in token-manager.ts and api-client.ts , those core files shouldn't branch on Action Center context. 
should be injected by TaskEventsService, not hardcoded in core.

src/models/action-center/tasks.types.ts
```diff
@@ -244,3 +244,53 @@ export type TaskGetByIdOptions = BaseOptions
...
+  LOADAPP = 'AC.loadApp',
+  ERROR = 'AC.error',
+  TOKENREFRESHED = 'AC.tokenRefreshed',
+  LANGUAGECHANGED = 'AC.languageChanged',
+  THEMECHANGED = 'AC.themeChanged',
+  REFRESHTOKEN = 'AC.refreshToken',
+  DISPLAYMESSAGE = 'AC.displayMessage',
+}
+
+enum Theme {
```
Raina451: export this also so that it will be visible in docs

src/models/action-center/taskEvents.models.ts
```diff
@@ -0,0 +1,179 @@
...
+   * });
+   *
+   * // Complete task with rejection
+   * taskEventsService.completeTask('Reject', {
+   *   approved: false,
+   *   reason: 'Missing information'
+   * });
+   * ```
+   */
+  completeTask(actionTaken: string, data: unknown): void;
```
Raina451: to complete a task, users have to use completetask from taskEventsService and cannot do tasks.complete from `@uipath/uipath-typescript/tasks`?

package.json
```diff
@@ -137,7 +137,8 @@
         "test:coverage": "vitest --coverage"
     },
     "dependencies": {
-        "@opentelemetry/sdk-logs": "^0.204.0"
+        "@opentelemetry/sdk-logs": "^0.204.0",
+        "jwt-decode": "^4.0.0"
```
Raina451: this adds a dependency in pkg for a trivial operation only. check if we can parse the
  JWT payload manually: `JSON.parse(atob(token.split('.')[1])).exp` maybe? this taskeventservice is meant for browser only so should be doable

src/services/action-center/taskEvents.ts
```diff
@@ -0,0 +1,217 @@
...
+        window.parent.postMessage(
+          { eventType, content },
+          this.parentOrigin!,
+        );
+      } catch (error) {
+        window.parent.postMessage(
+          {
+            eventType: ActionCenterEventNames.ERROR,
+            content: {
+              errorData: error,
```
Raina451: The error handler calls postMessage again which could also throw making an infinite loop?

src/core/auth/token-manager.ts
```diff
@@ -26,7 +29,13 @@ export class TokenManager {
...
     private config: Config,
     private isOAuth: boolean = false
-  ) {}
+  ) {
+    if (isInActionCenter()) {
+      const acTokenManager = new ActionCenterTokenManager(config, (tokenInfo) => this.setToken(tokenInfo));
+      this.actionCenterTokenRefreshFn = (tokenInfo: TokenInfo) => acTokenManager.refreshAccessToken(tokenInfo);
+      this.isOAuth = false;
+    }
+  }
```
Raina451: TokenManager is part of the core bundle (`@uipath/uipath-typescript/core`) which every SDK consumer imports. With this change, it now imports ActionCenterTokenManager from src/services... - a service-layer module.

This is a problem because the SDK uses modular exports (/core, /tasks, etc.) so users only bundle what they need. Since token-manager.ts is in core, the Action Center code now gets bundled into core for every consumer. It also creates a circular dependency chain: core -> action-center service => core (since action-center-token-manager.ts imports from src/core/errors and src/core/config).

Instead of having TokenManager auto-detect and instantiate ActionCenterTokenManager internally, accept an optional token refresh function only as a constructor parameter. Then have AuthService (which already creates the TokenManager) do the environment check and inject the refresh strategy from outside. 
for eg in token manager, have : `private externalTokenRefreshFn: TokenRefreshFn | null =null` (optional param) and in constructor  do not do env detection, just store what was given 
```
if (externalTokenRefreshFn) {
        this.externalTokenRefreshFn = externalTokenRefreshFn;
        this.isOAuth = false;
      }
```
and in auth service (which already creates the token manager ) inject the externalRefreshFn. and authservice contructor can detect the isinactioncenter and creates ActionCenterTokenManager . basically build and refresh strategy outside tokenmanager

packages/coded-action-apps/src/coded-action-apps-service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /** 
+   * Validates that the given origin is a known UiPath environment or a local development server,
+   * guarding against cross-origin message spoofing.
+   *
+   * @param origin - The origin string to validate, sourced from the `basedomain` query parameter.
+   * @returns `true` if the origin is trusted, `false` otherwise.
+   */
+  private isValidOrigin(origin: string | null): boolean {
+    const ALLOWED_ORIGINS = ['https://alpha.uipath.com', 'https://staging.uipath.com', 'https://cloud.uipath.com'];
```
swati354: Should we make this a module level constant?

packages/coded-action-apps/README.md
swati354: needs to be updated.

packages/coded-action-apps/src/coded-action-apps-service.ts
```diff
@@ -0,0 +1,181 @@
...
+      return false;
+    }
+
+    if (ALLOWED_ORIGINS.includes(origin)) {
+      return true;
+    }
+
+    try {
+      const url = new URL(origin);
+      return url.hostname === 'localhost';
```
Raina451: nit: Consider restricting localhost validation to http/https protocols, not just hostname.

Raina451 (commented):
also have we decided on how docs will be discoverable for this pkg?

packages/coded-action-apps/src/coded-action-apps-service.ts
```diff
@@ -0,0 +1,181 @@
...
+import { CodedActionAppsServiceModel } from './coded-action-apps.models';
+import { ActionCenterEventNames, ActionCenterEventResponsePayload } from './types.internal';
+
+const INIT_TIMEOUT = 3000;
+
+/**
+ * Service for bi-directional communication between coded action apps and Action Center
+ */
+export class CodedActionAppsService implements CodedActionAppsServiceModel {
+  private readonly parentOrigin = new URLSearchParams(window.location.search).get('basedomain');
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: this should always be running in an iframe which should have a window object

packages/coded-action-apps/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Installation
+
+```bash
+npm install @uipath/coded-action-apps
+```
+
+## Overview
+
+When Action Center opens a coded action app, it renders it inside an iframe. This package provides a service — `CodedActionAppsService` — that the app uses to:
```
vnaren23: Action Center renders a coded action app within an iframe. `@uipath/coded-action-apps` provides service `CodedActionAppsService` which offers below capabilities

- **Receive** - `getTask()` - On app load, UiPath Action Center provides the task details.
- **Complete** - `completeTask` - Provide a better explanation
- **Notify** -  Action Center when task data changes (e.g. to enable the Save button).
- **Display** toast messages inside Action Center.

packages/coded-action-apps/README.md
```diff
@@ -0,0 +1,197 @@
...
+## Security
+
+The service reads the trusted parent origin from the `basedomain` query parameter in the iframe URL. All outbound `postMessage` calls are targeted exclusively at that origin. Incoming messages are validated against the same origin before being processed. Allowed origins are:
+
+- `https://cloud.uipath.com`
+- `https://staging.uipath.com`
+- `https://alpha.uipath.com`
+- `http://localhost` / `https://localhost` (for local development)
+
+Any message from an origin outside this list is silently discarded.
```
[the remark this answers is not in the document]
Raina451: i dont think user need to know this

swati354 (approved):

PR #230, opened 2026-02-16
feat: add sdk logout method - [PLT-97974]

Raina451:
will updates sample once we release package with this change. because right now users are cloning the samples from main branch, then app wont work. 
#235

### DIFF
--- a/tests/integration/auth-errors.integration.test.ts
+++ b/tests/integration/auth-errors.integration.test.ts
@@ -14,6 +14,9 @@ function isForbiddenError(error: any, keywords: string[] = []): boolean {
     error.status === 403 ||
     error.statusCode === 401 ||
     error.status === 401 ||
+    // Server may return HTML (redirect/error page) for invalid org/tenant,
+    // causing a JSON parse error — this still indicates rejection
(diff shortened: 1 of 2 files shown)


swati354:
How does this work for secret based authentication?
calling logout would clear the token from memory, and then calling initialize() would return return early without setting the token, and the api calls would fail?
https://github.com/UiPath/uipath-typescript/blob/4499d9d30e530f9352ab9f1d11af73f9364d17cd/src/core/uipath.ts#L108

The user would have to create an entirely new UiPath instance in this case?

Raina451:
> How does this work for secret based authentication? calling logout would clear the token from memory, and then calling initialize() would return return early without setting the token, and the api calls would fail?
> 
> https://github.com/UiPath/uipath-typescript/blob/4499d9d30e530f9352ab9f1d11af73f9364d17cd/src/core/uipath.ts#L108
> 
> The user would have to create an entirely new UiPath instance in this case?

logout doesn't really apply to secret-based auth since the token is just a static, there's no session to end. We already tell users  that "you don't need to call initialize() for secret auth" - so it doesn't make sense to then require them to call it again after logout to recover. I've just made logout() silently skip for secret  auth, it won't clear anything or change state.

PR #234, opened 2026-02-18
feat(cli) : added pull cmd

deepeshrai-tech:
Associated Jira Ticket : PLT-96946

## Summary

Adds a **pull** command to sync project files from Studio Web into the local workspace (mirror of push), plus a **schema version check** on push so users are prompted to pull first when their local code version is behind the remote. 
---

### Behavior

- **Source of truth**: Remote Studio Web project. Only files under remote `source/` are pulled; they are written at the **target directory root**
- **Target directory**: Current working directory by default, or `--target-dir <path>`.
- **Project type validation**: Pull is allowed only for:
  - **Web App**: `webAppManifest.json` with `"type": "App_ProCode"`, and
  - **Automation**: a `.uiproj` file with `"ProjectType": "WebApp"`.
  Validation is done by downloading the manifest and checking these keys.
- **Overwrite protection**:
  - If local files would be overwritten and `--overwrite` is not set: in an interactive terminal, the user is prompted **"This pull will overwrite one or more local files. Do you want to continue? (Y/n)"**. Otherwise the command fails and suggests `--overwrite` or moving files.
  - With `--overwrite`, overwrite is allowed without prompting.
- **Build output excluded**: The folder indicated by `buildDir` in remote `push_metadata.json` (e.g. `output/`, `dist/`) is **not** pulled (no files and no folder skeleton). This keeps the local tree to source code only.
- **`push_metadata.json`**: Pulled from remote `source/push_metadata.json` and written locally to **`.uipath/push_metadata.json`** (same location as push uses).
- **Progress**: A single updating line (e.g. `[pull] Pulling files... 3/15 completed`). Failed downloads are still logged and reported at the end.
- **Concurrency**: Up to 10 files are downloaded (and written) in parallel.

### CLI

- **Command**: `uipath pull [project-id]`
- **Flags**: `--target-dir`, `--overwrite`, `--baseUrl`, `--orgId`, `--tenantId`, `--tenantName`, `--accessToken`
- **Project ID**: Argument or `UIPATH_PROJECT_ID` env var.

---

## 2. Push: schema version check (pull first when behind)

- **Problem**: When multiple people contribute, someone can push and then another person (with an older local `schemaVersion`) pushes and overwrites remote with older code.
- **Solution**: Before push, the CLI compares local `schemaVersion` (from `.uipath/push_metadata.json`) with remote (from `source/push_metadata.json`). If **remote > local**, push is **blocked** and the user is told to run `uipath pull` first, then push again.
- **Edge cases**:
  - **First push** (no remote metadata): Check is skipped; push proceeds.
  - **Local metadata missing, remote exists**: Check is skipped; push uses remote metadata as base and recreates local (existing behavior).
- **Implementation**: `ensureSchemaVersionNotBehindRemote` and `PushBehindRemoteError` in `metadata.ts`; semantic version comparison (major.minor.patch). Telemetry continues to use existing push flow; no new events required for this check.

---

## 3. Config clarity

- **WebAppPushConfig** is used by both push and pull. Push uses all fields; pull uses `projectId`, `rootDir` (target dir), `envConfig`, `logger` and passes placeholder `bundlePath`/`manifestFile` so the API layer accepts the same type.
- **Comments added**: JSDoc on `WebAppPushConfig` in `types.ts`; inline comment in `commands/pull.ts` where the config object is built.

---

---

## 4. README

- **New section**: "Pull (Studio Web)" with prerequisites, examples, arguments, flags, and behavior (including overwrite prompt and buildDir exclusion).
- **Commands list**: Added `uipath pull`.
- **Workflow**: Added "Pull (Studio Web)" steps (pull → edit → push).
- **Intro**: Updated to mention "pushing or pulling" and Studio Web.

---

## Testing suggestions

- **Pull**: From a Studio Web project with `source/` content and `push_metadata.json` (with `buildDir`). Verify only source files are pulled, `buildDir` folder is not created, `.uipath/push_metadata.json` is written, progress line updates, and overwrite prompt appears when local files would be overwritten (without `--overwrite`).
- **Push when behind**: Set local `schemaVersion` lower than remote (e.g. edit `.uipath/push_metadata.json`), run push → should fail with message to pull first. After pull, push should succeed.
- **First push / local metadata missing**: Unchanged behavior (push succeeds; no version check when remote has no metadata; local metadata recreated from remote when missing).

### DIFF
--- a/packages/cli/src/core/webapp-file-handler/pull-utils.ts
+++ b/packages/cli/src/core/webapp-file-handler/pull-utils.ts
@@ -7,7 +7,11 @@ import * as path from 'path';
 import { PUSH_METADATA_RELATIVE_PATH, PUSH_METADATA_FILENAME } from '../../constants/api.js';
 import { AUTH_CONSTANTS } from '../../constants/index.js';
 import { PULL_WEB_APP_MANIFEST, PACKAGE_JSON_FILENAME } from '../../constants/pull.js';
-import { REMOTE_SOURCE_FOLDER_NAME, normalizePathToForwardSlashes } from './structure.js';
+import {
+  REMOTE_SOURCE_FOLDER_NAME,
+  normalizePathToForwardSlashes,
+  normalizeBundlePath,
+} from './structure.js';
 import type { ProjectFile } from './types.js';
 
 /** Markers that indicate a directory is likely the root of an app project (for pull target-dir soft check). */
@@ -80,12 +84,11 @@ export function getLocalRelativePath(relativePath: string): string {
 /**
  * Returns true if relativePath is exactly buildDir or under buildDir/.
  * Used to exclude build output from pull (no files and no folder skeleton).
- * Normalizes buildDir (forward slashes, no trailing slash) for consistent behavior.
- * Shared with push (local-files) for the same "under build dir" check.
+ * Uses normalizeBundlePath from structure (same as push) for consistent behavior; ReDoS-safe.
  */
 export function isPathUnderBuildDir(relativePath: string, buildDir: string | null): boolean {
   if (!buildDir) return false;
-  const norm = buildDir.replace(/\\/g, '/').replace(/\/+$/, '');
+  const norm = normalizeBundlePath(buildDir);
   if (!norm) return false;
   return relativePath === norm || relativePath.startsWith(`${norm}/`);
 }
--- a/packages/cli/src/core/webapp-file-handler/structure.ts
+++ b/packages/cli/src/core/webapp-file-handler/structure.ts
@@ -19,9 +19,12 @@ export const PUSH_METADATA_REMOTE_PATH = `${REMOTE_SOURCE_FOLDER_NAME}/push_meta
 /** Remote project-root manifest filename (at root, not under source). Updated on push with config.bundlePath. */
 export const WEB_APP_MANIFEST_FILENAME = 'webAppManifest.json';
 
-/** Normalize bundle path (strip leading/trailing slashes, use forward slashes). */
+/** Normalize bundle path (strip leading/trailing slashes, use forward slashes). No regex (ReDoS-safe). */
 export function normalizeBundlePath(bundlePath: string): string {
-  return bundlePath.replace(/\\/g, '/').replace(/^\/+|\/+$/g, '') || bundlePath;
+  let s = bundlePath.split('\\').join('/');
+  while (s.length > 0 && s.startsWith('/')) s = s.slice(1);
+  while (s.length > 0 && s.endsWith('/')) s = s.slice(0, -1);
+  return s || bundlePath;
 }
 
 /** Remote root for pushed content: source/<buildDir> (e.g. source/dist). All build dir files go under this path. */
@@ -39,11 +42,12 @@ export function localPathToRemotePath(
   remoteContentRoot: string
 ): string {
   const normalized = localPath.replace(/\\/g, '/');
-  const prefix = bundlePath.replace(/\\/g, '/').replace(/\/+$/, '') + '/';
-  if (normalized.startsWith(prefix)) {
+  const buildDirNorm = normalizeBundlePath(bundlePath);
+  const prefix = buildDirNorm ? buildDirNorm + '/' : '';
+  if (buildDirNorm && normalized.startsWith(prefix)) {
     return remoteContentRoot + '/' + normalized.slice(prefix.length);
   }
-  if (normalized === bundlePath.replace(/\\/g, '/').replace(/\/+$/, '')) {
+  if (buildDirNorm && normalized === buildDirNorm) {
     return remoteContentRoot;
   }
   return remoteContentRoot + '/' + normalized;
@@ -68,7 +72,7 @@ export function getRemotePathForLocalPath(
   remoteContentRoot: string
 ): string {
   const normalized = localPath.replace(/\\/g, '/');
-  const buildDirNorm = bundlePath.replace(/\\/g, '/').replace(/\/+$/, '');
+  const buildDirNorm = normalizeBundlePath(bundlePath);
   const prefix = buildDirNorm ? buildDirNorm + '/' : '';
   if (buildDirNorm && (normalized === buildDirNorm || normalized.startsWith(prefix))) {
     return localPathToRemotePath(localPath, bundlePath, remoteContentRoot);


packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,125 @@
...
+    }),
+  };
+
+  static override flags = {
+    help: Flags.help({ char: 'h' }),
+    overwrite: Flags.boolean({
+      description: 'Allow overwriting existing local files',
+      default: false,
+    }),
+    'target-dir': Flags.string({
```
Raina451: why 'target-dir' and not target-dir (without string)?
also follow camelCase

packages/cli/src/core/webapp-file-handler/handler.ts
```diff
@@ -103,6 +104,13 @@ export class WebAppFileHandler {
     const remoteContentRoot = getRemoteContentRoot(this.config.bundlePath);
     const contentRootExists = hasFolderByPath(fullRemoteFolders, remoteContentRoot);
 
+    this.config.logger.log(chalk.gray('[push] Checking code version (local vs remote)...'));
+    await ensureSchemaVersionNotBehindRemote(
+      this.config,
+      fullRemoteFiles,
+      (cfg, fileId) => api.downloadRemoteFile(cfg, fileId)
```
Raina451: ensureSchemaVersionNotBehindRemote runs after the lock is acquired. If it throw push behind rmote error, the lock remains held until it auto-expires. Is that acceptable?

packages/cli/src/core/webapp-file-handler/pull-validation.ts
```diff
@@ -0,0 +1,81 @@
...
+    const content = await api.downloadRemoteFile(config, file.id);
+    let parsed: { ProjectType?: string };
+    try {
+      parsed = JSON.parse(content.toString('utf8')) as { ProjectType?: string };
+    } catch {
+      throw new ProjectPullError(MESSAGES.ERRORS.PULL_UIPROJ_PROJECT_TYPE_INVALID);
+    }
+    if (parsed.ProjectType !== PULL_UIPROJ_PROJECT_TYPE) {
+      throw new ProjectPullError(MESSAGES.ERRORS.PULL_UIPROJ_PROJECT_TYPE_INVALID);
+    }
```
Raina451: these are almost doing the same thing,  can we move the common logic out

packages/cli/src/core/webapp-file-handler/pull-errors.ts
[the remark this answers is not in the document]
deepeshrai-tech: removed it

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,139 @@
...
+          process.exit(0);
+        }
+      }
+    }
+
+    // Same config shape as push (WebAppPushConfig); bundlePath and manifestFile are placeholders so runPull and the API layer accept the same type.
+    const config = {
+      projectId,
+      rootDir,
+      bundlePath: 'dist',
```
NishankSiddharth: what does it mean to pass `dist` in bundlepath for pull command?

packages/cli/src/constants/messages.ts
```diff
@@ -95,6 +95,26 @@ export const MESSAGES = {
...
+    // Pull command
+    PULL_PROJECT_ID_REQUIRED:
+      'Project ID is required. Use: uipath pull <project-id> or set UIPATH_PROJECT_ID in the .env file and use uipath pull directly',
+    PULL_VALIDATION_FAILED: 'Validation failed',
+    PULL_FAILED_PREFIX: 'Pull failed: ',
+    PULL_TARGET_DIR_NOT_FOUND: 'Target directory does not exist.',
+    PULL_TARGET_DIR_NOT_DIRECTORY: 'Target path is not a directory.',
+    PULL_PROJECT_NOT_FOUND: 'Project not found or empty in Studio Web.',
+    PULL_PROJECT_TYPE_UNSUPPORTED: 'Project type is not supported. Expected a Web App project (webAppManifest.json with type: "App_ProCode") or .uiproj (with ProjectType: "WebApp").',
+    PULL_WEB_APP_MANIFEST_TYPE_INVALID: 'webAppManifest.json must have "type": "App_ProCode" to pull.',
```
NishankSiddharth: this is too low level and internal detail of manifest, in case of such failure we should show then project we are pulling is not supported etc, we should use some user friendly message.

packages/cli/src/constants/pull.ts
```diff
@@ -0,0 +1,21 @@
...
+/** Root-level manifest that indicates a Web App project (Studio Web). Must match structure.WEB_APP_MANIFEST_FILENAME. */
+export const PULL_WEB_APP_MANIFEST = 'webAppManifest.json';
+
+/** Required value in webAppManifest.json "type" field for pull to be allowed. */
+export const PULL_WEB_APP_MANIFEST_TYPE = 'App_ProCode';
+
+/** File extension that indicates a classic automation project (.uiproj). */
+export const PULL_UIPROJ_EXTENSION = '.uiproj';
+
+/** Required value in .uiproj "ProjectType" field for pull to be allowed. */
```
NishankSiddharth: why do we need this?
WebAppManifest validation should be enough

packages/cli/src/core/webapp-file-handler/api.ts
```diff
@@ -98,11 +98,34 @@ export async function fetchRemoteStructure(
   return (await response.json()) as ProjectStructure;
 }
 
+/**
+ * Releases the project lock. Call after push completes (success or failure) so the project is not
+ * left locked until backend timeout. DELETE /Project/{projectId}/Lock/{lockKey}.
+ */
+export async function releaseLock(config: WebAppPushConfig, lockKey: string): Promise<void> {
```
NishankSiddharth: did we check if this is the right API to call to release the lock?

packages/cli/src/core/webapp-file-handler/pull-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+import { PUSH_METADATA_RELATIVE_PATH, PUSH_METADATA_FILENAME } from '../../constants/api.js';
+import { PULL_WEB_APP_MANIFEST, PULL_UIPROJ_EXTENSION } from '../../constants/pull.js';
+import { REMOTE_SOURCE_FOLDER_NAME, normalizePathToForwardSlashes } from './structure.js';
+import type { ProjectFile } from './types.js';
+
+/** Markers that indicate a directory is likely the root of an app project (for pull target-dir soft check). */
+const PROJECT_ROOT_MARKERS = [
+  'package.json',
+  PULL_WEB_APP_MANIFEST, // webAppManifest.json
+  '.uipath',
```
NishankSiddharth: do we have any const for this directory name which we can reuse ?

packages/cli/src/core/webapp-file-handler/pull-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+] as const;
+
+/**
+ * Returns true if dir looks like a project root (has package.json, webAppManifest.json, .uipath, or a .uiproj file).
+ * Used to warn when pulling into the current directory and it does not look like a project root.
+ */
+export function looksLikeProjectRoot(dir: string): boolean {
```
NishankSiddharth:
```suggestion
export function isProjectRootDirectory(dir: string): boolean {
```

packages/cli/src/core/webapp-file-handler/pull-validation.ts
```diff
@@ -0,0 +1,92 @@
...
+  const exact = filesMap.get(PULL_WEB_APP_MANIFEST);
+  if (exact) return PULL_WEB_APP_MANIFEST;
+  for (const p of filesMap.keys()) {
+    if (p.endsWith(`/${PULL_WEB_APP_MANIFEST}`)) return p;
+  }
+  return null;
+}
+
+/** Finds the first remote path ending with .uiproj (case-insensitive). */
+export function findUiprojPath(filesMap: Map<string, ProjectFile>): string | null {
```
NishankSiddharth: we should skip reading uiproj

packages/cli/src/core/webapp-file-handler/pull-validation.ts
```diff
@@ -0,0 +1,92 @@
...
+      file,
+      MESSAGES.ERRORS.PULL_WEB_APP_MANIFEST_TYPE_INVALID
+    );
+    if (parsed['type'] !== PULL_WEB_APP_MANIFEST_TYPE) {
+      throw new Error(MESSAGES.ERRORS.PULL_WEB_APP_MANIFEST_TYPE_INVALID);
+    }
+    return;
+  }
+
+  if (uiprojPath) {
```
NishankSiddharth: remove

packages/cli/src/core/webapp-file-handler/run-pull.ts
```diff
@@ -0,0 +1,240 @@
...
+  logger.log(chalk.gray(`[pull] Recreating ${allFolderPaths.size} folder(s) under target-dir...`));
+  ensureLocalFolders(rootDir, allFolderPaths);
+
+  const failedPaths: Array<{ path: string; error: string }> = [];
+  const entries = [...filesMap.entries()];
+  const total = entries.length;
+  let completedCount = 0;
+  const progressPrefix = MESSAGES.INFO.PULL_PROGRESS;
+
+  const processOne = async ([remotePath, file]: [string, ProjectFile]): Promise<void> => {
```
NishankSiddharth: let's name this method better

packages/cli/src/core/webapp-file-handler/types.ts
```diff
@@ -1,5 +1,9 @@
 import type { EnvironmentConfig } from '../../types/index.js';
 
+/**
+ * Config used by both push and pull. Push uses all fields (rootDir = project root, bundlePath/manifestFile for build and metadata).
+ * Pull uses projectId, rootDir (target dir), envConfig, logger; bundlePath and manifestFile are passed as placeholders so the same type is accepted by the API layer.
+ */
 export interface WebAppPushConfig {
```
NishankSiddharth: if it is used for both let's rename this interface

packages/cli/src/commands/push.ts
```diff
@@ -28,11 +28,12 @@ export default class Push extends Command {
...
     help: Flags.help({ char: 'h' }),
-    'build-dir': Flags.string({
-      description: 'Relative path to the build output directory (e.g. dist, build, out). Default: dist',
+    buildDir: Flags.string({
+      description:
+        'Relative path to the build output directory; should be the root of the build output (e.g. dist, build, out). Default: dist',
       default: 'dist',
     }),
-    'ignore-resources': Flags.boolean({
+    ignoreResources: Flags.boolean({
```
swati354: we should update the examples as well ?
https://github.com/UiPath/uipath-typescript/blob/2243b06f9a4eff82769b156cea25e1b2db1201f2/packages/cli/src/commands/pull.ts#L16

And here - https://github.com/UiPath/uipath-typescript/blob/2243b06f9a4eff82769b156cea25e1b2db1201f2/packages/cli/src/constants/messages.ts#L72

packages/cli/src/core/webapp-file-handler/handler.ts
```diff
@@ -88,6 +89,23 @@ export class WebAppFileHandler {
...
+          await api.releaseLock(this.config, this.lockKey);
+          this.config.logger.log(chalk.gray('[push] Lock released.'));
+        } catch (err) {
+          const msg = err instanceof Error ? err.message : String(err);
+          this.config.logger.log(chalk.yellow(`[push] Could not release lock: ${msg}`));
+        }
+        this.lockKey = null;
+      }
+    }
+  }
```
swati354: similar to this, should `importReferencedResources` also release the lock?

packages/cli/src/core/webapp-file-handler/metadata.ts
```diff
@@ -93,6 +117,41 @@ function createNewMetadataPayload(config: WebAppPushConfig): PushMetadata {
...
+  downloadRemoteFile: (config: WebAppPushConfig, fileId: string) => Promise<Buffer>
+): Promise<void> {
+  const remoteEntry = remoteFiles.get(PUSH_METADATA_REMOTE_PATH);
+  if (!remoteEntry) return;
+
+  let localVersion: string | null = null;
+  const metadataPath = path.join(config.rootDir, config.manifestFile);
+  try {
+    const content = fs.readFileSync(metadataPath, 'utf-8');
+    const parsed = JSON.parse(content) as { schemaVersion?: unknown };
```
swati354: shouldn't `JSON.parse` be inside a try-catch?

packages/cli/src/core/webapp-file-handler/structure.ts
```diff
@@ -3,9 +3,14 @@ import type { ProjectFile, ProjectFolder, ProjectStructure } from './types.js';
 /** Remote root folder name for pushed content (build + source). All pushed files live under this. */
 export const REMOTE_SOURCE_FOLDER_NAME = 'source';
 
+/** Normalize path to forward slashes. Use for parsing or comparing remote paths (getRemoteFilesMap / getRemoteFoldersMap use /). */
+export function normalizePathToForwardSlashes(p: string): string {
```
swati354: nit : we have extracted a method but are still using manual replace at a lot of places in the file
eg, line 56, 70, etc

packages/cli/src/commands/pull.ts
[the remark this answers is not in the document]
deepeshrai-tech: its just that wanted to use same interface for both push and pull cmds, will make sense once I change the name of the common interface

packages/cli/README.md
```diff
@@ -237,6 +237,71 @@ This command will:
...
+src/drafts/
+```
+
+### Pull (Studio Web)
+
+Pull project files from a Studio Web project into your local workspace. Syncs only files under the remote `source/` folder and writes them at the target directory root. Build output (`buildDir` from remote `push_metadata.json`, e.g. `output/` or `dist/`) is excluded so you get source code only.
+
+**Prerequisites:**
+- Environment variables set (see [Prerequisites](#prerequisites)), or pass `--orgId`, `--tenantId`, `--accessToken` (and optionally `--baseUrl`, `--tenantName`) for non-interactive use. `UIPATH_PROJECT_ID` can be used instead of passing `<project-id>`.
+- The remote project must be a supported type: Web App (`webAppManifest.json` with `type: "App_ProCode"`) and automation project (`.uiproj` with `ProjectType: "WebApp"`).
```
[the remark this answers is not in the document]
deepeshrai-tech: have removed the .uiproj check wasn't necessary so only have one single check now

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,139 @@
...
+      manifestFile: PUSH_METADATA_RELATIVE_PATH,
+      envConfig,
+      logger: this as { log: (message: string) => void },
+    };
+
+    const promptOverwrite =
```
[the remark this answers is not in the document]
deepeshrai-tech: Screenshot 2026-02-24 at 12 50 01 PM

NishankSiddharth (approved):

packages/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,139 @@
...
+import { PUSH_METADATA_RELATIVE_PATH } from '../constants/api.js';
+import { AUTH_CONSTANTS, MESSAGES } from '../constants/index.js';
+import { PULL_OVERWRITE_LIST_MAX_DISPLAY } from '../constants/pull.js';
+import { getEnvironmentConfig } from '../utils/env-config.js';
+import { runPull, isProjectRootDirectory } from '../core/webapp-file-handler/index.js';
+import { track } from '../telemetry/index.js';
+
+export default class Pull extends Command {
+  static override description =
+    'Pull project files from Studio Web into the local workspace';
```
Raina451: can we inform user to run this command from root of the project?

PR #244, opened 2026-02-23
chore: add retry to publish npm step in publish sdk worfkflow

Raina451:
in one of recent run, npm publish failed with error: "package with this version already exists" which was false positive. but re-runing worked. 
Adding retry to npm publish step

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,9 +57,14 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Publish to npm
-        run: |
-          echo "@uipath:registry=https://registry.npmjs.org" > .npmrc
-          npm publish --provenance --access public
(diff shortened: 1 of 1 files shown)


maninder-uipath (approved):

.github/workflows/publish.yml
```diff
@@ -57,9 +57,14 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Publish to npm
-        run: |
-          echo "@uipath:registry=https://registry.npmjs.org" > .npmrc
-          npm publish --provenance --access public
+        uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3.0.2
```
NishankSiddharth: ideally we should not need a retry logic, can we check why it failed in the first place?
Same version exist error might be pointing to some other issue?

Raina451:
not needed

PR #249, opened 2026-02-25
test: test coverage improvement

vnaren23:
1. Dont skip tests, rather fail.
2. Improve coverage on entities and choice sets

### DIFF
--- a/tests/integration/shared/data-fabric/attachment.integration.test.ts
+++ b/tests/integration/shared/data-fabric/attachment.integration.test.ts
@@ -83,13 +83,13 @@ describe.skipIf(!hasAttachmentConfig).each(modes)(
         );
 
         // Now delete the attachment
-        await expect(
-          entities.deleteAttachment(
-            ATTACHMENT_CONFIG.entityId,
-            ATTACHMENT_CONFIG.recordId,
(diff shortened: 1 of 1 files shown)


tests/integration/shared/data-fabric/choicesets.integration.test.ts
```diff
@@ -1,60 +1,72 @@
 import { describe, it, expect } from 'vitest';
-import { getServices, setupUnifiedTests, InitMode } from '../../config/unified-setup';
+import { getServices, getTestConfig, setupUnifiedTests, InitMode } from '../../config/unified-setup';
 
 const modes: InitMode[] = ['v0', 'v1'];
 
 describe.each(modes)('Data Fabric ChoiceSets - Integration Tests [%s]', (mode) => {
   setupUnifiedTests(mode);
+  const testConfig = getTestConfig();
```
Raina451: gettestconfig runs before setupUnifiedTests, better move to before all to esnure config is fully intiialised

vnaren23: Not necessarily. We have getTestConfig called even within methods. Whats do we lose by calling it after setup?

deepeshrai-tech (approved):

PR #331, opened 2026-03-26
[PLT-100513] feat(codedapp-tool): add CI publish workflow

deepeshrai-tech:
- Add publish-codedapp-tool.yml workflow that auto-publishes to both npmjs.org and GitHub Packages on merge to main
- Version check prevents re-publishing existing versions
- Switch publishConfig from GitHub Packages to npmjs.org (public)
- Bump version to 0.1.6 to match CLI major.minor for auto-install


packages/codedapp-tool/package.json
maninder-uipath: package-lock missing

PR #337, opened 2026-03-30
feat(jobs): add Jobs.stop() method [PLT-100574]

ninja-shreyash:
## Method Added

| Service | `jobs.stop()` | `stop(jobKeys: string[], folderId: number, options?: JobStopOptions): Promise<void>` |
| Bound | `job.stop()` | `stop(options?: JobStopOptions): Promise<void>` |

## Endpoint Called

| `stop()` (resolve keys) | GET | `/odata/Jobs` (via `getAll` with `$filter=key in (...)`) | `OR.Jobs` |
| `stop()` (stop jobs) | POST | `/odata/Jobs/UiPath.Server.Configuration.OData.StopJobs` | `OR.Jobs` |

- Extends `FolderScopedService` — sets `X-UIPATH-OrganizationUnitId` header
- Composite method: resolves UUID keys → integer IDs via `getAll()`, then calls the StopJobs OData action
- Keys are chunked (max 50 per query) and resolved in parallel via `Promise.all`
- Returns `Promise<void>` — succeeds silently or throws typed errors on failure

## Example Usage

```typescript
import { Jobs, StopStrategy } from '@uipath/uipath-typescript/jobs';

const jobs = new Jobs(sdk);

// Stop a single job with default soft stop
await jobs.stop([<jobKey>], <folderId>);

// Force-kill multiple jobs
await jobs.stop(
  [<jobKey1>, <jobKey2>],
  <folderId>,
  { strategy: StopStrategy.Kill }
);

// Stop via bound method on a job response
const allJobs = await jobs.getAll({ folderId: <folderId> });
const runningJob = allJobs.items.find(j => j.state === 'Running');
if (runningJob) {
  await runningJob.stop();
}
```

### Composition flow

```
stop(jobKeys, folderId, options?)
  │
  ├─ empty array? → return immediately
  │
  ├─ validate folderId
  │
  ├─ resolveJobKeys(jobKeys, folderId)
  │   ├─ deduplicate keys
  │   ├─ chunk into batches of 50
  │   ├─ Promise.all: getAll({ filter: "key in (...)", select: "id,key" }) per chunk
  │   ├─ build keyToIdMap from responses
  │   └─ throw ValidationError if any keys not found
  │
  └─ stopJobsByIds(jobIds, strategy, headers)
      └─ POST /odata/Jobs/UiPath.Server.Configuration.OData.StopJobs
         body: { jobIds: [...], strategy: "SoftStop" | "Kill" }
```

### Types

| Type | Purpose |
|------|---------|
| `JobStopOptions` | Optional strategy param (`SoftStop` or `Kill`) |
| `StopStrategy` | Enum — re-exported from `jobs.types.ts` for subpath import |

## Additional Changes

- **`api-client.ts`**: Replaced `content-length: 0` header check with body-based empty response detection (`response.text()` → check empty → `JSON.parse`)
- **`StopStrategy` re-export**: Re-exported from `jobs.types.ts` so the barrel `export *` picks it up — removed direct export from `jobs/index.ts`

## Files

| Area | Files |
|------|-------|
| Endpoint | `src/utils/constants/endpoints/orchestrator.ts` |
| Types | `src/models/orchestrator/jobs.types.ts` |
| Constants | `src/models/orchestrator/jobs.constants.ts` |
| Models | `src/models/orchestrator/jobs.models.ts` |
| Service | `src/services/orchestrator/jobs/jobs.ts` |
| Core | `src/core/http/api-client.ts` |
| Unit tests | `tests/unit/services/orchestrator/jobs.test.ts` (33 tests), `tests/unit/models/orchestrator/jobs.test.ts` (8 tests) |
| Integration tests | `tests/integration/shared/orchestrator/jobs.integration.test.ts` (10 tests) |
| Test utils | `tests/utils/constants/jobs.ts` |
| Docs | `docs/oauth-scopes.md`, `agent_docs/rules.md` |

Refs PLT-100574


ninja-shreyash:
@claude pls add OR.Jobs.Write along with OR.Jobs for Jobs stop() method

tests/integration/shared/orchestrator/jobs.integration.test.ts
```diff
@@ -188,6 +188,56 @@ describe.each(modes)('Orchestrator Jobs - Integration Tests [%s]', (mode) => {
...
+
+    it('should return empty result when called with empty array', async () => {
+      const { jobs, folderId } = getJobsService();
+
+      if (!folderId) {
+        console.warn('INTEGRATION_TEST_FOLDER_ID not configured, skipping stop test.');
+        return;
+      }
+
+      const result = await jobs.stop([], folderId);
```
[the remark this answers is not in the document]
ninja-shreyash: fixed

tests/unit/services/orchestrator/jobs.test.ts
```diff
@@ -494,4 +495,135 @@ describe('JobService Unit Tests', () => {
...
+      });
+
+      const error = createMockError(TEST_CONSTANTS.ERROR_MESSAGE);
+      mockApiClient.post.mockRejectedValueOnce(error);
+
+      await expect(
+        jobService.stop([JOB_TEST_CONSTANTS.JOB_KEY], TEST_CONSTANTS.FOLDER_ID)
+      ).rejects.toThrow(TEST_CONSTANTS.ERROR_MESSAGE);
+    });
+  });
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/models/orchestrator/jobs.models.ts
```diff
@@ -100,6 +101,43 @@ export interface JobServiceModel {
...
+   * const result = await jobs.stop(
+   *   ['c80c3b30-f010-4eb8-82d4-b67bc615e137'],
+   *   123
+   * );
+   *
+   * // Force-kill multiple jobs
+   * const killResult = await jobs.stop(
+   *   ['c80c3b30-f010-4eb8-82d4-b67bc615e137', '24ef1040-454d-4184-b994-c641ee32318d'],
+   *   123,
+   *   { strategy: StopStrategy.Kill }
```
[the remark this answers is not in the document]
ninja-shreyash: done

tests/integration/shared/orchestrator/jobs.integration.test.ts
```diff
@@ -91,6 +91,56 @@ describe.each(modes)('Orchestrator Jobs - Integration Tests [%s]', (mode) => {
     });
   });
 
+  describe('stop', () => {
+    it('should stop a running job with soft stop strategy', async () => {
+      const { jobs, folderId } = getJobsService();
+
+      if (!folderId) {
+        console.warn('INTEGRATION_TEST_FOLDER_ID not configured, skipping stop test.');
+        return;
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -222,4 +258,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+
+    for (let i = 0; i < uniqueKeys.length; i += JOB_KEY_RESOLUTION_CHUNK_SIZE) {
+      const chunk = uniqueKeys.slice(i, i + JOB_KEY_RESOLUTION_CHUNK_SIZE);
+      const filterValues = chunk.map((key) => `'${key}'`).join(',');
+      const filter = `Key in (${filterValues})`;
+
+      const response = await this.get<CollectionResponse<{ Key: string; Id: number }>>(
+        JOB_ENDPOINTS.GET_ALL,
+        {
+          params: { $filter: filter, $select: 'Id,Key' },
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -222,4 +258,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+      );
+
+      for (const job of response.data.value) {
+        keyToIdMap.set(job.Key, job.Id);
+      }
+    }
+
+    const missingKeys = uniqueKeys.filter((key) => !keyToIdMap.has(key));
+    if (missingKeys.length > 0) {
+      throw new Error(`Jobs not found for keys: ${missingKeys.join(', ')}`);
```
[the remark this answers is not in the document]
ninja-shreyash: done

tests/integration/shared/orchestrator/jobs.integration.test.ts
```diff
@@ -188,6 +188,51 @@ describe.each(modes)('Orchestrator Jobs - Integration Tests [%s]', (mode) => {
...
+      const { jobs, folderId } = getJobsService();
+
+      if (!folderId) {
+        throw new Error('INTEGRATION_TEST_FOLDER_ID not configured — cannot run stop test.');
+      }
+
+      // Find a running job to stop
+      const runningJobs = await jobs.getAll({
+        folderId,
+        filter: "State eq 'Running'",
```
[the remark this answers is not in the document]
ninja-shreyash: @claude make this change in the PR

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -211,6 +216,55 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+   */
+  @track('Jobs.Stop')
+  async stop(
+    jobKeys: string[],
+    folderId: number,
+    options?: JobStopOptions
+  ): Promise<OperationResponse<JobStopData>> {
+    if (jobKeys.length === 0) {
+      return { success: true, data: { jobIds: [] } };
+    }
```
swati354, line 256: add folderId validation as well.

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -211,6 +216,55 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+
+    const headers = createHeaders({ [FOLDER_ID]: folderId });
+    const strategy = options?.strategy ?? StopStrategy.SoftStop;
+
+    const jobIds = await this.resolveJobKeys(jobKeys, headers);
+
+    await this.stopJobsByIds(jobIds, strategy, headers);
+
+    return { success: true, data: { jobIds } };
```
swati354, line 265: what happens if we fail to stop a job?

tests/integration/shared/orchestrator/jobs.integration.test.ts
```diff
@@ -188,6 +188,51 @@ describe.each(modes)('Orchestrator Jobs - Integration Tests [%s]', (mode) => {
...
+      const runningJobs = await jobs.getAll({
+        folderId,
+        filter: "state eq 'Running'",
+        pageSize: 1,
+      });
+
+      if (runningJobs.items.length === 0) {
+        console.warn('No running jobs available to test stop. Skipping.');
+        return;
+      }
```
swati354, line 209: why are we not throwing error here?

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -252,4 +306,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+        keyToIdMap.set(job.Key, job.Id);
+      }
+    }
+
+    const missingKeys = uniqueKeys.filter((key) => !keyToIdMap.has(key));
+    if (missingKeys.length > 0) {
+      throw new ValidationError({ message: `Jobs not found for keys: ${missingKeys.join(', ')}` });
+    }
+
+    return jobKeys.map((key) => keyToIdMap.get(key)!);
```
swati354, line 344: we should use `uniqueKeys` here?

src/models/orchestrator/jobs.models.ts
```diff
@@ -129,6 +130,41 @@ export interface JobServiceModel {
...
+   *
+   * // Force-kill multiple jobs
+   * const result = await jobs.stop(
+   *   [<jobKey1>, <jobKey2>],
+   *   <folderId>,
+   *   { strategy: StopStrategy.Kill }
+   * );
+   * ```
+   */
+  stop(
```
swati354, line 163: why are we not binding this method to a job. Yes it is capable of operating on multiple jobs, but if users want to stop a single job returned from a response, it would still be useful?

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -10,12 +10,17 @@ import { ValidationError, ServerError } from '../../../core/errors';
...
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../../utils/pagination';
 import { PaginationHelpers } from '../../../utils/pagination/helpers';
 import { PaginationType } from '../../../utils/pagination/internal-types';
 import { track } from '../../../core/telemetry';
 import type { IUiPath } from '../../../core/types';
+import { OperationResponse, CollectionResponse } from '../../../models/common/types';
+import { StopStrategy } from '../../../models/orchestrator/processes.types';
+
+/** Maximum number of job keys to resolve in a single OData filter query */
+const JOB_KEY_RESOLUTION_CHUNK_SIZE = 50;
```
swati354, line 23: keep in `jobs.constants.ts`. 
Also is this number randomly chosen by us?

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -252,4 +306,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+   */
+  private async stopJobsByIds(
+    jobIds: number[],
+    strategy: StopStrategy,
+    headers: Record<string, string>
+  ): Promise<void> {
+    await this.post(
+      JOB_ENDPOINTS.STOP,
+      { jobIds, strategy },
+      { headers, responseType: RESPONSE_TYPES.TEXT }
```
swati354, line 358: the api returns text? Or is it just to handle an empty 200 response? If it is the later, are there other ways to handle this at api-client?

src/models/orchestrator/jobs.models.ts
```diff
@@ -129,6 +130,41 @@ export interface JobServiceModel {
    * ```
    */
   getOutput(jobKey: string, folderId: number): Promise<Record<string, unknown> | null>;
+
+  /**
+   * Stops one or more jobs by their UUID keys.
+   *
+   * Resolves the provided job UUID keys to integer IDs, then sends a stop request to the Orchestrator.
+   * Keys are processed in chunks of 50 to avoid URL length limits. Throws if any keys cannot be resolved.
```
swati354, line 138: Why do the end users need to know about our handling of keys?

docs/oauth-scopes.md
```diff
@@ -16,6 +16,7 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `getAll()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getById()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getOutput()` | `OR.Jobs` or `OR.Jobs.Read` |
+| `stop()` | `OR.Jobs` or `OR.Jobs.Write` |
```
swati354, line 19: would just `OR.Jobs.Write` work or would they also need `OR.Jobs.Read` since we are doing a get call internally?

src/models/orchestrator/jobs.types.ts
```diff
@@ -171,3 +171,23 @@ export type JobGetAllOptions = RequestOptions & PaginationOptions & {
  */
 export interface JobGetByIdOptions extends BaseOptions {}
 
+/**
+ * Options for stopping jobs
+ */
+export interface JobStopOptions {
+  /**
+   * The stop strategy to use.
+   * - `SoftStop` — sends a cancellation request and waits for the job to finish gracefully
```
swati354, line 180: thought - we don't really wait for the job to stop right? 
should we modify the comment to be 
`requests graceful cancellation; the job completes its current activity before stopping`

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: this is how python sdk is also doing. so the number is coming from there

docs/oauth-scopes.md
ninja-shreyash: Yes. Lets mention `OR.Jobs` scope only

src/models/orchestrator/jobs.models.ts
ninja-shreyash: removed

src/models/orchestrator/jobs.models.ts
ninja-shreyash: Agreed. Done

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: done

tests/integration/shared/orchestrator/jobs.integration.test.ts
ninja-shreyash: done

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: Yes done

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: The error from the API propagates naturally. If stopJobsByIds fails, the promise rejects with a typed error (e.g., NotFoundError, AuthenticationError).

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: Yes we can handle this in api-client itself. Done

src/models/orchestrator/jobs.types.ts
ninja-shreyash: Yes we dont wait in case of softStop. Job transitions to "Stopping"

ninja-shreyash: And Kill transitions job to "Terminating" So both are fire and forget. Updated the doc accordingly

src/core/http/api-client.ts
```diff
@@ -107,6 +107,12 @@ export class ApiClient {
         return blob as T;
       }
 
+      // Handle text response type for endpoints that return empty or plain text bodies
+      if (options.responseType === RESPONSE_TYPES.TEXT) {
```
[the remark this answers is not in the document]
ninja-shreyash: done

tests/utils/constants/jobs.ts
```diff
@@ -29,4 +31,8 @@ export const JOB_TEST_CONSTANTS = {
 
   // Error Messages
   ERROR_JOB_NOT_FOUND: 'Job not found',
+  ERROR_JOBS_NOT_FOUND_FOR_KEYS: 'Jobs not found for keys',
+
+  // Stop resolution
+  KEY_RESOLUTION_CHUNK_SIZE: 50,
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/services/orchestrator/jobs/index.ts
```diff
@@ -23,3 +23,4 @@ export { JobService as Jobs, JobService } from './jobs';
 export * from '../../../models/orchestrator/jobs.types';
 export * from '../../../models/orchestrator/jobs.models';
 export { JobState } from '../../../models/common/types';
+export { StopStrategy } from '../../../models/orchestrator/processes.types';
```
Sarath1018: question: why not export *?

docs/oauth-scopes.md
```diff
@@ -16,6 +16,7 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `getAll()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getById()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getOutput()` | `OR.Jobs` or `OR.Jobs.Read`, `OR.Folders` or `OR.Folders.Read` |
+| `stop()` | `OR.Jobs` |
```
Sarath1018, line 19:
```suggestion
| `stop()` | `OR.Jobs` or `OR.Jobs.Read` and `OR.Jobs.Write` |
```

Sarath1018: Just mentioning the alternative. what do you think @swati354

src/core/http/api-client.ts
```diff
@@ -97,7 +97,7 @@ export class ApiClient {
         throw ErrorFactory.createFromHttpStatus(response.status, errorInfo);
       }
 
-      if (response.status === 204) {
+      if (response.status === 204 || response.headers.get('content-length') === '0') {
```
Raina451: is it because response code is 200, but there is no content?

Raina451: If a server sends content-length: 0 on a response that has meaning in its status code alone that will get silently swallowed. better check if we could do somehting like this: 

```
const text = await response.text();
  if (!text) {
    return undefined as T;
  }
  return json parse
```
checking a header isbit  less reliable than checking the actual body content.

src/models/orchestrator/jobs.constants.ts
```diff
@@ -1,3 +1,6 @@
+/** Maximum number of job keys to resolve in a single OData filter query (matches Python SDK) */
```
Raina451: remove "(matches Python SDK) "

src/models/orchestrator/jobs.models.ts
```diff
@@ -129,6 +130,40 @@ export interface JobServiceModel {
...
+   * // Force-kill multiple jobs
+   * const result = await jobs.stop(
+   *   [<jobKey1>, <jobKey2>],
+   *   <folderId>,
+   *   { strategy: StopStrategy.Kill }
+   * );
+   * ```
+   */
+  stop(
+    jobKeys: string[],
```
Raina451, line 162: we used id as a naming, better use the same here instead of key.

src/models/orchestrator/jobs.types.ts
```diff
@@ -171,3 +171,23 @@ export type JobGetAllOptions = RequestOptions & PaginationOptions & {
...
+   */
+  strategy?: StopStrategy;
+}
+
+/**
+ * Data returned from a stop operation
+ */
+export interface JobStopData {
+  /** The resolved integer IDs of the jobs that were stopped */
+  jobIds: number[];
```
Raina451: here id is needed and not key?

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -252,4 +306,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
       throw new ServerError({ message: 'Failed to parse job output file as JSON' });
     }
   }
+
+  /**
+   * Resolves job UUID keys to integer IDs via OData filter queries.
+   * Chunks keys into batches to avoid URL length limits.
+   */
+  private async resolveJobKeys(
```
Raina451, line 311: why is this needed?
api isnt expecting the guid key itself?

src/models/orchestrator/jobs.types.ts
```diff
@@ -171,3 +171,23 @@ export type JobGetAllOptions = RequestOptions & PaginationOptions & {
...
+ * Options for stopping jobs
+ */
+export interface JobStopOptions {
+  /**
+   * The stop strategy to use.
+   * - `SoftStop` — requests graceful cancellation; the job completes its current activity before stopping
+   * - `Kill` — requests immediate termination of the job
+   * @default StopStrategy.SoftStop
+   */
+  strategy?: StopStrategy;
```
Raina451, line 186: Consider re-exporting StopStrategy from jobs.types.ts . i think its only available in process types.

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -252,4 +306,56 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+
+    for (let i = 0; i < uniqueKeys.length; i += JOB_KEY_RESOLUTION_CHUNK_SIZE) {
+      const chunk = uniqueKeys.slice(i, i + JOB_KEY_RESOLUTION_CHUNK_SIZE);
+      const filterValues = chunk.map((key) => `'${key}'`).join(',');
+      const filter = `Key in (${filterValues})`;
+
+      const response = await this.get<CollectionResponse<{ Key: string; Id: number }>>(
+        JOB_ENDPOINTS.GET_ALL,
+        {
+          params: { $filter: filter, $select: 'Id,Key', $top: chunk.length },
```
Raina451: chunks are independent, should be parallelized, else it will be slow.

docs/oauth-scopes.md
Raina451: not sure if OR.Jobs.Read would work. but should be added OR.Jobs.Read or OR.Jobs.Write whatever is the correct one.

src/core/http/api-client.ts
ninja-shreyash: Yes exactly

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: API accepts array of job IDs only. We need to internally resolve the GUIDs and prepare list of unique job id's and call JOBS_STOP endpoint on these ids.

docs/oauth-scopes.md
ninja-shreyash: We actually need both read and write. That is why gave OR.Jobs in the scope

src/services/orchestrator/jobs/jobs.ts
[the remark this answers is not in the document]
ninja-shreyash: done

src/models/orchestrator/jobs.types.ts
ninja-shreyash: done

src/services/orchestrator/jobs/index.ts
ninja-shreyash: We want StopStrategy to be available at @uipath/uipath-typescript/jobs import path. Everything else is not desired to be available at this path.

src/models/orchestrator/jobs.models.ts
swati354: > we used id as a naming, better use the same here instead of key.

I think key would be better? the response for jobs returns both id and key, it could be misleading.
We use id naming only in getById and thats because the method name itself suggests that.
If you see getOutput, we use the key terminology there.

src/core/http/api-client.ts
ninja-shreyash: done

src/models/orchestrator/jobs.constants.ts
ninja-shreyash: done

src/models/orchestrator/jobs.models.ts
ninja-shreyash: +1

src/models/orchestrator/jobs.types.ts
ninja-shreyash: We can instead just return success. No point in just returning the keys/ids back

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: done

src/models/orchestrator/jobs.types.ts
Raina451: should be fine.
i am only against returning number id.

ninja-shreyash: done

src/models/orchestrator/jobs.models.ts
```diff
@@ -129,6 +129,40 @@ export interface JobServiceModel {
...
+   *
+   * @param jobKeys - Array of job UUID keys to stop (e.g., from {@link JobGetResponse}.key)
+   * @param folderId - The folder ID where the jobs reside (required)
+   * @param options - Optional {@link JobStopOptions} including stop strategy
+   * @returns Promise that resolves when the jobs are stopped successfully, or rejects on failure
+   *
+   * @example
+   * ```typescript
+   * // Stop a single job with default soft stop
+   * const result = await jobs.stop([<jobKey>], <folderId>);
```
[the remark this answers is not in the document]
ninja-shreyash: @claude make the suggested chnages

Sarath1018 (approved):

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -252,4 +303,61 @@ export class JobService extends FolderScopedService implements JobServiceModel {
       throw new ServerError({ message: 'Failed to parse job output file as JSON' });
     }
   }
+
+  /**
+   * Resolves job UUID keys to integer IDs via the getAll method.
+   * Chunks keys into batches to avoid URL length limits.
+   */
+  private async resolveJobKeys(
```
Raina451, line 311: this is not the actual jobid with which orchestrator calls. 
i dont think we should build ids from keys and call the stop api.

src/models/orchestrator/jobs.types.ts
```diff
@@ -171,3 +173,16 @@ export type JobGetAllOptions = RequestOptions & PaginationOptions & {
...
```
Raina451, line 186: i see there is batchExecutionKey also in request dto, what is that?

Image

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: it is the same id. When we call getAll and filter by the guid, then we can get the id from the job response and use the same integer id to call stop api.

ninja-shreyash: cc: @vnaren23

src/models/orchestrator/jobs.types.ts
ninja-shreyash: I checked that every job will have a batchExecutionKey. We get this in RawJobGetResponse as well. But I am still trying to figure out how can we execute jobs in a batch

ninja-shreyash: But the backend method only accepts JobIds and StopStrategy. So swagger is outdated.

https://github.com/UiPath/Orchestrator/blob/62a92d3fc4940fe1079f24fcc151225b363bdaf7/src/Web/Web/OData/Models/StopJobsRequest.cs

Raina451 (approved):

PR #345, opened 2026-04-06
feat:generate all types for all services [APPS-33351]

SunilYadav-uipath:
1. Bundled type definitions — a single .d.ts combining all service entries.
2. Dynamically generated from serviceEntries so new services are included automatically.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,7 +5,7 @@ import dts from 'rollup-plugin-dts';              // Generates TypeScript declar
 import json from '@rollup/plugin-json';           // Imports JSON files as ES6 modules
 import alias from '@rollup/plugin-alias';         // Path alias support (@/ -> src/)
 import builtins from 'builtin-modules';           // List of Node.js built-in modules (fs, crypto, etc.)
-import { readFileSync } from 'fs';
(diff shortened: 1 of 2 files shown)


vnaren23 (approved):

rollup.config.js
```diff
@@ -229,5 +224,33 @@ serviceEntries.forEach(({ name, input, output }) => {
...
+  );
+  writeFileSync(tempEntryPath, [
+    '// Auto-generated from serviceEntries in rollup.config.js — do not edit',
+    ...exportLines,
+  ].join('\n') + '\n');
+
+  return {
+    input: tempEntryPath,
+    output: { file: 'dist/types/index.d.ts', format: 'es' },
+    plugins: [
```
Raina451, line 245: Can we simplify this? Instead of generating a temp file at build time(writeFileSync -> build -> unlinkSync), 
just create a static `src/types/index.ts` that re-exports all service indexes:

 ```
  export * from '../services/data-fabric/index';
  export * from '../services/action-center/index';
  // ... all services
```

and then add a DTS-only rollup config pointing at it. should have same result and no temporary handling at rollup ?

Raina451 (changes requested):

vnaren23:
Closing stale PR

PR #352, opened 2026-04-07
Added DataFabric Sdk

AditiGoyalUipath:
Adds SDK for below DataFabric capabilities
   CREATE_ENTITY
    DELETE_ENTITY
    QUERY_BY_ID
    BULK_IMPORT_BY_ID

https://uipath.atlassian.net/browse/DS-7644
https://uipath.atlassian.net/wiki/spaces/CC/pages/90518585507/Design+Document+Data+Fabric+CLI+Tool+SDK+Integration+for+Coding+Agents

https://uipath.atlassian.net/browse/DS-7912
https://uipath.atlassian.net/browse/DS-7913
https://uipath.atlassian.net/browse/DS-7914


ninja-shreyash (commented):
Can you add new methods to oauth-scopes.md? Also please add integration tests for the newly onboarded methods

vitest.integration.config.ts
```diff
@@ -6,8 +6,8 @@ export default defineConfig({
     environment: 'node',
     include: ['tests/integration/**/*.integration.test.ts'],
     exclude: ['tests/integration/shared/maestro/**', 'tests/integration/shared/orchestrator/attachments.integration.test.ts'],
-    testTimeout: 30000,
-    hookTimeout: 30000,
+    testTimeout: 60000,
+    hookTimeout: 60000,
```
Sarath1018: Noticed that the test timeout was increased from 30s to 60s. Just wanted to understand the reason behind this change. Did you encounter any test cases that are genuinely taking more than 30 seconds to complete? Or was this increased due to intermittent timeouts (for example, API delays) rather than the test logic itself? I am suspecting some tests take more time because of ratelimit issue. I noticed that yesterday, I am working on it.

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +868,75 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+    fields: EntitySchemaField[],
+  ): Promise<void> {
+    const payload = {
+      description: (raw.description as string | undefined) ?? '',
+      displayName: (raw.displayName as string | undefined) ?? (raw.name as string | undefined) ?? '',
+      entityDefinition: {
+        id: entityId,
+        name: raw.name,
+        fields,
+        folderId: '00000000-0000-0000-0000-000000000000',
```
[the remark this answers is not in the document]
AditiGoyalUipath: Since we are onboarding tenant-level entities, the folderId is currently hardcoded as `00000000-0000-0000-0000-000000000000` in DataFabric. With folder-supported entities now released, we need to properly onboard and support them in the SDK across all relevant operations.
But still its better to handle here atleast.

vitest.integration.config.ts
AditiGoyalUipath: Yes few existing tests were taking more time. I increased to check if PR checks are pass. Will revert this and check if test build is getting successful.

docs/oauth-scopes.md
```diff
@@ -47,6 +47,12 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `downloadAttachment()` | `DataFabric.Data.Read` |
 | `uploadAttachment()` | `DataFabric.Data.Write` |
 | `deleteAttachment()` | `DataFabric.Data.Write` |
+| `queryRecords()` | `DataFabric.Data.Read` |
+| `bulkImport()` | `DataFabric.Data.Write` |
+| `createEntity()` | `DataFabric.Schema.Write` |
+| `deleteEntity()` | `DataFabric.Schema.Write` |
+| `addField()` | `DataFabric.Schema.Write` |
+| `removeField()` | `DataFabric.Schema.Write` |
```
swati354: We need to rename these methods 
1. `createEntity()` -> `create()`
2. `deleteEntity()` -> `deleteById()`
3. `addField()` -> `insertFieldById()`
4. `removeField()` -> `deleteFieldById`

src/services/data-fabric/entities.ts
```diff
@@ -437,6 +442,98 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+   * // Stop on first failure
+   * const result = await entities.bulkImport("<entityId>", csvContent, "import.csv", {
+   *   failOnFirst: true
+   * });
+   * ```
+   */
+  @track('Entities.BulkImport')
+  async bulkImport(entityId: string, csvContent: string, fileName: string, options: EntityBulkImportOptions = {}): Promise<EntityBulkImportResponse> {
+    const formData = new FormData();
+    formData.append('file', new Blob([csvContent], { type: 'text/csv' }), fileName);
```
swati354: The content-type for this endpoint is specified as `application/json`. Wouldn't formData send it as `multipart/form-data`?

https://docs.uipath.com/data-service/automation-cloud/latest/api-guide/bulk-upload

src/models/data-fabric/entities.constants.ts
```diff
@@ -28,6 +28,47 @@ export const EntityMap = {
...
+  defaultValue: string | null;
+  isRbacEnabled: boolean;
+  isUnique: boolean;
+  isEncrypted: boolean;
+}
+
+/**
+ * Maps friendly type names to the API field payload components
+ */
+export const EntitySchemaFieldTypeMap: Record<string, {
```
swati354: could this be typed to `Record<EntityFieldType, ...>`

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +505,42 @@ export interface EntityServiceModel {
...
+  addField(entityId: string, field: EntityFieldDefinition): Promise<void>;
+
+  /**
+   * Removes a field from a Data Fabric entity
+   *
+   * @param entityId - UUID of the entity
+   * @param fieldName - Name of the field to remove
+   * @returns Promise resolving when the field is removed
+   */
+  removeField(entityId: string, fieldName: string): Promise<void>;
```
swati354: Add @example blocks on all these methods.

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +121,114 @@ export interface EntityDeleteOptions {
...
 
 
+/**
+ * A single filter condition for querying entity records
+ */
+export interface EntityQueryFilter {
+  /** Name of the field to filter on */
+  fieldName: string;
+  /** Comparison operator (e.g., "=", "!=", ">", "<", ">=", "<=", "contains", "startswith", "endswith", "in", "not in") */
+  operator: string;
```
swati354: if this is a fixed set, please convert this to an enum.

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +121,114 @@ export interface EntityDeleteOptions {
...
+  /** Value to compare against */
+  value: string;
```
swati354: question : would even the numbers and boolean values be passed as strings?

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +121,114 @@ export interface EntityDeleteOptions {
...
+  /** Filter conditions to apply */
+  filterGroup?: EntityQueryFilterGroup;
+  /** List of field names to include in results (returns all fields if omitted) */
+  selectedFields?: string[];
+  /** Sort options for the results */
+  sortOptions?: EntityQuerySortOption[];
+  /** Number of records to skip (for pagination) */
+  start?: number;
+  /** Maximum number of records to return */
+  limit?: number;
```
swati354: The SDK already supports pagination. Can't we use that instead?

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +121,114 @@ export interface EntityDeleteOptions {
...
+  /** Name of the field to sort by */
+  fieldName: string;
+  /** Whether to sort in descending order (default: false) */
+  isDescending?: boolean;
+}
+
+/**
+ * Options for querying entity records with filters, sorting, and pagination
+ */
+export interface EntityQueryRecordsOptions {
```
swati354: question : I see that you have not added aggregates, joins or groupBy. Is that intentional?

src/services/data-fabric/entities.ts
```diff
@@ -605,9 +702,97 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+   * ```typescript
+   * const entityId = await entities.createEntity("Products", "Product catalog", [
+   *   { name: "ProductName", type: "text", isRequired: true },
+   *   { name: "Price", type: "decimal" },
+   * ]);
+   * ```
+   */
+  @track('Entities.CreateEntity')
+  async createEntity(name: string, description: string, fields: EntityFieldDefinition[]): Promise<string> {
+    const sanitizedName = name.replace(/\s+/g, '').toLowerCase();
```
swati354: Entity names and field names are being silently sanitized. Can't we require the user to pass a valid technical name?
Could we instead validate and require the user to pass a valid technical name (lowercase, no spaces)? And provide a separate displayName option for the human-readable label?

^ for both the entity name and field name.

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +866,75 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+      entityDefinition: {
+        id: entityId,
+        name: raw.name,
+        fields,
+        folderId: (raw.folderId as string | undefined) ?? '00000000-0000-0000-0000-000000000000',
+        isRbacEnabled: (raw.isRbacEnabled as boolean | undefined) ?? false,
+        externalFields: [],
+      },
+    };
+    await this.post(DATA_FABRIC_ENDPOINTS.ENTITY.CREATE_ENTITY, payload);
```
swati354: Why are we using the create entity endpoint for addField/removeField? I see dedicated endpoints in the DF swagger
`POST /api/Entity/{entityId}/field`
`DELETE /api/Entity/{entityId}/field/{fieldId}`
The current implementation reads the full schema, modifies the fields array, then writes the entire schema back via the create endpoint.

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +866,75 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+    const payload = {
+      description: (raw.description as string | undefined) ?? '',
+      displayName: (raw.displayName as string | undefined) ?? (raw.name as string | undefined) ?? '',
+      entityDefinition: {
+        id: entityId,
+        name: raw.name,
+        fields,
+        folderId: (raw.folderId as string | undefined) ?? '00000000-0000-0000-0000-000000000000',
+        isRbacEnabled: (raw.isRbacEnabled as boolean | undefined) ?? false,
+        externalFields: [],
```
swati354: won't this wipe out all external fields?

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +866,75 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+        isRequired: Boolean(f.isRequired ?? false),
+        fieldDisplayType: (f.fieldDisplayType as string | undefined) ?? 'Basic',
+        sqlType: (f.sqlType as { name: string; lengthLimit?: number }) ?? { name: 'NVARCHAR', lengthLimit: 200 },
+        choiceSetId: (f.choiceSetId as string | null) ?? null,
+        defaultValue: (f.defaultValue as string | null) ?? null,
+        isRbacEnabled: Boolean(f.isRbacEnabled ?? false),
+        isUnique: Boolean(f.isUnique ?? false),
+        isEncrypted: Boolean(f.isEncrypted ?? false),
+      }));
+  }
```
swati354: if we use `raw: RawEntityGetResponse`, there won't be any need for manually casting every field.

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: Yes, We are not onboarding aggregates

tests/integration/shared/data-fabric/attachment.integration.test.ts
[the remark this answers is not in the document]
AditiGoyalUipath: As discussed these we will add post some of this API onboarded to External app

src/services/data-fabric/entities.ts
AditiGoyalUipath: Post Datafabric External fields addField/removeField were deprecated . In datafabric We are using the Upsert API for Create and Update of Entity.

src/services/data-fabric/entities.ts
AditiGoyalUipath: Fixed it also currently We are not onboarding entity with externalFields so this won't be an issue

src/services/data-fabric/entities.ts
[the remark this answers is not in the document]
AditiGoyalUipath: We need to check the doc. The API supports multipart/form-data. Can verify form UI

docs/oauth-scopes.md
```diff
@@ -47,6 +47,11 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `downloadAttachment()` | `DataFabric.Data.Read` |
 | `uploadAttachment()` | `DataFabric.Data.Write` |
 | `deleteAttachment()` | `DataFabric.Data.Write` |
+| `queryRecords()` | `DataFabric.Data.Read` |
+| `importRecordsById()` | `DataFabric.Data.Write` |
+| `create()` | `DataFabric.Schema.Write` |
+| `deleteById()` | `DataFabric.Schema.Write` |
+| `updateById()` | `DataFabric.Schema.Write` |
```
swati354: 1. except the `queryRecords` method, remove others.
2. queryRecords to be named `queryRecordsById` and then to be listed here as : 
 `queryRecordsById()/ queryrecords` | `DataFabric.Data.Read` |

src/models/data-fabric/entities.constants.ts
```diff
@@ -25,23 +26,59 @@ export const EntityMap = {
...
   updateTime: 'updatedTime',
   sqlType: 'fieldDataType',
-  fieldDefinition: 'fieldMetaData' 
+  fieldDefinition: 'fieldMetaData'
+};
+
+/**
+ * Maps EntityFieldDataType values to the API field payload components for create/update operations
+ */
+export const EntitySchemaFieldTypeMap: Record<EntityFieldDataType, EntitySchemaFieldMapping> = {
```
swati354, line 18: these three maps - `EntitySchemaFieldTypeMap`, `FieldDisplayTypeToDataType` and `EntityFieldTypeMap` should stay in sync. Could we add a unit test for this?

docs/oauth-scopes.md
```diff
@@ -47,6 +47,7 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `downloadAttachment()` | `DataFabric.Data.Read` |
 | `uploadAttachment()` | `DataFabric.Data.Write` |
 | `deleteAttachment()` | `DataFabric.Data.Write` |
+| `queryRecordsById()` | `DataFabric.Data.Read` |
```
swati354: `| queryRecordsById() / queryRecords() | DataFabric.Data.Read |`

src/models/data-fabric/entities.internal-types.ts
```diff
@@ -1,15 +1,18 @@
...
 
-// Field Definition Interface
-interface FieldDefinition {
-  name: string;
-  sqlType: FieldType;
+/**
+ * Shape of each entry in EntitySchemaFieldTypeMap — internal only.
+ */
+export interface EntitySchemaFieldMapping {
+  sqlTypeName: string;
```
swati354: can we type the `sqlTypeName` as `SqlFieldType`?
it will create a circular dependency if you import it though, so move `SqlFieldType` to this file and then import it in constants file.

src/models/data-fabric/entities.models.ts
```diff
@@ -93,7 +99,7 @@ export interface EntityServiceModel {
    * const choicesets = new ChoiceSets(sdk);
    *
    * // Get entity metadata with methods
-   * const entity = await entities.getById(<entityId>);
+   * const entity = await entities.getById("<entityId>");
```
swati354, line 99: let's keep the Id unquoted for now(here and on the comments above the newly added methods as well). If we want to add the quotes, that would require changes at a lot of other places as well.

src/models/data-fabric/entities.models.ts
```diff
@@ -333,11 +334,52 @@ export interface EntityServiceModel {
    */
   deleteRecordsById(id: string, recordIds: string[], options?: EntityDeleteRecordsOptions): Promise<EntityDeleteResponse>;
 
+
   /**
-   * @deprecated Use {@link deleteRecordsById} instead.
-   * @hidden
+   * Queries entity records with filters, sorting, and pagination
+   *
+   * @param entityId - UUID of the entity
```
swati354: name the field as `id` as is done for other entity methods.

src/models/data-fabric/entities.models.ts
```diff
@@ -609,16 +724,47 @@ export interface EntityMethods {
...
+   * ```typescript
+   * const entity = await entities.getById("<entityId>");
+   * await entity.updateById({
+   *   displayName: "Updated Name",
+   *   addFields: [{ name: "notes", type: EntityFieldDataType.MULTILINE_TEXT }],
+   * });
+   * ```
+   * @internal
+   */
+  updateById(options: EntityUpdateByIdOptions): Promise<void>;
```
swati354: 1. bind the query and importRecords method as well.
2. bound methods should drop the `byId` prefix.
3. drop the deprecated delete and update methods from here as well.

update accordingly in the `createEntityMethods` as well.

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +121,114 @@ export interface EntityDeleteOptions {
...
+  /** Comparison operator (e.g., "=", "!=", ">", "<", ">=", "<=", "contains", "startswith", "endswith", "in", "not in") */
+  operator: string;
+  /** Value to compare against */
+  value: string;
+}
+
+/**
+ * A group of query filters combined with a logical operator
+ */
+export interface EntityQueryFilterGroup {
```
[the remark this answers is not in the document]
AditiGoyalUipath: yes they are optional .filterGroups if there is and/OR operation. queryFilters Will contain =, != certain operators

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: If its needed I can take this as part of follow up PR.

src/models/data-fabric/entities.models.ts
```diff
@@ -713,21 +849,30 @@ function createEntityMethods(entityData: RawEntityGetResponse, service: EntitySe
...
-  entityData: RawEntityGetResponse, 
+  entityMetadata: RawEntityGetResponse,
   service: EntityServiceModel
 ): EntityGetResponse {
-  const methods = createEntityMethods(entityData, service);
-  return Object.assign({}, entityData, methods) as EntityGetResponse;
-}
\ No newline at end of file
+  return Object.assign({}, entityMetadata, createEntityMethods(entityMetadata, service)) as EntityGetResponse;
+}
```
swati354, line 918: these changes aren't required.

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +503,79 @@ export interface EntityServiceModel {
...
   deleteAttachment(entityId: string, recordId: string, fieldName: string): Promise<EntityDeleteAttachmentResponse>;
+
+  /**
+   * Creates a new Data Fabric entity with the given schema
+   *
+   * @param name - Technical entity name — lowercase, starts with a letter, letters/numbers/underscores only
+   *   (e.g., `"product_catalog"`).
+   * @param description - Entity description
+   * @param fields - Array of field definitions
+   * @param displayName - Human-readable display name shown in the UI (defaults to `name`)
```
swati354: 1. no `displayName` field is added as a parameter on create.
2. `options` param is missing

src/models/data-fabric/entities.models.ts
```diff
@@ -339,6 +344,45 @@ export interface EntityServiceModel {
...
+   * @param options - Import options including failOnFirst
+   * @returns Promise resolving to {@link EntityBulkImportResponse} with record counts
+   * @example
+   * ```typescript
+   * const csvContent = "name,age\nJohn,30\nJane,25";
+   * const result = await entities.bulkImport(<entityId>, csvContent, "import.csv");
+   * console.log(`Inserted ${result.insertedRecords} of ${result.totalRecords} records`);
+   * ```
+   */
+  bulkImport(entityId: string, csvContent: string, fileName: string, options?: EntityBulkImportOptions): Promise<EntityBulkImportResponse>;
```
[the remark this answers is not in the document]
AditiGoyalUipath: Since they are internal currently skipping this

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,202 @@ export interface EntityDeleteOptions {
...
+  /** Human-readable display name shown in the UI (defaults to `name` if omitted) */
+  displayName?: string;
+  /** Folder ID for the entity (defaults to the tenant-level folder) */
+  folderId?: string;
+  /** Whether role-based access control is enabled for this entity (default: false) */
+  isRbacEnabled?: boolean;
+  /** Whether Insights integration is enabled for this entity (default: false) */
+  isInsightsEnabled?: boolean;
+  /** External field source definitions (default: empty) */
+  externalFields?: unknown[];
```
swati354: can this be typed to `ExternalSourceFields[]`?

src/models/data-fabric/entities.types.ts
swati354: Yes if we can use the existing pagination, we should use that. If this is to be taken up in a follow up PR, better raise the PR's against a different branch. We can then merge that branch to main once all changes are done.

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,202 @@ export interface EntityDeleteOptions {
...
+  /** New description for the entity */
+  description?: string;
+  /** Whether role-based access control is enabled for this entity */
+  isRbacEnabled?: boolean;
+}
+
+/**
+ * Response from a bulk import operation
+ */
+export interface EntityBulkImportResponse {
```
swati354: `EntityBulkImportResponse` -> `EntityImportRecordsResponse`

src/models/data-fabric/entities.models.ts
AditiGoyalUipath: most of the places it is entityId

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,202 @@ export interface EntityDeleteOptions {
...
+  /** Total number of records in the import file */
+  totalRecords?: number;
+  /** Number of records successfully inserted */
+  insertedRecords?: number;
+  /** Link to download the error file (if any records failed) */
+  errorFileLink?: string | null;
```
swati354: all three are optional? I think only `errorFileLink` should be truly optional. Could you please check this.

src/utils/constants/endpoints/data-fabric.ts
```diff
@@ -19,6 +26,11 @@ export const DATA_FABRIC_ENDPOINTS = {
     UPDATE_RECORD_BY_ID: (entityId: string, recordId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/update/${recordId}`,
     UPDATE_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/update-batch`,
     DELETE_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/delete-batch`,
+    UPSERT: `${DATAFABRIC_BASE}/api/Entity`,
+    DELETE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}/delete`,
```
swati354: not a DELETE call but a post call to the delete endpoint. Intentional?

src/services/data-fabric/entities.ts
```diff
@@ -437,6 +445,101 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+   *   console.log(`Error file link: ${result.errorFileLink}`);
+   * }
+   * ```
+   * @internal
+   */
+  @track('Entities.ImportRecordsById')
+  async importRecordsById(entityId: string, file: EntityFileType): Promise<EntityImportRecordsResponse> {
+    const formData = new FormData();
+    if (file instanceof Uint8Array) {
+      formData.append('file', new Blob([file.buffer as ArrayBuffer]));
```
swati354, line 537: > new Blob([file.buffer as ArrayBuffer]));

can we pass just the file? `new Blob([file]);`

src/models/data-fabric/entities.types.ts
```diff
@@ -258,32 +459,40 @@ export interface Field {
  * Detailed field definition
  */
 export interface FieldMetaData {
```
swati354, line 482: changing this interface would be a breaking change.
Better create a separate interface in internal-types.

src/models/data-fabric/entities.models.ts
```diff
@@ -609,16 +726,47 @@ export interface EntityMethods {
...
+   */
+  update(options: EntityUpdateByIdOptions): Promise<void>;
+
 }
 
 /**
  * Entity with methods combining metadata with operation methods
  */
-export type EntityGetResponse = RawEntityGetResponse & EntityMethods;
+export interface EntityGetResponse extends RawEntityGetResponse, EntityMethods {}
```
swati354: we don't need to convert it to an interface. it is only for non intersection types becasue they break TypeDoc docs.
We'll remove it from our `rules.md `as well.

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +950,39 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+
+  /**
+   * Validates that a name is a valid technical identifier:
+   * lowercase, starts with a letter, contains only letters, numbers, and underscores.
+   * @throws Error if the name is invalid
+   */
+  private static validateTechnicalName(name: string, context: string): void {
+    if (!/^[a-z][a-z0-9_]*$/.test(name)) {
+      throw new ValidationError({
+        message: `Invalid ${context} name '${name}'. Name must be lowercase, start with a letter, and contain only letters, numbers, and underscores (e.g., "${name.toLowerCase().replace(/\s+/g, '_').replace(/[^a-z0-9_]/g, '')}").`
```
swati354: the suggestion here itself could be invalid if the name starts with a digit?

src/utils/constants/endpoints/data-fabric.ts
AditiGoyalUipath: Yes this is supported method in DataFabric used by client.

AditiGoyalUipath: We do have both methods . Better to use DELETE for External Client. I will update this.

src/services/data-fabric/entities.ts
AditiGoyalUipath: No it is not supported and needs to be typecasted to ArrayBuffer. It is similar to other Attachment methods

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: We need the metadata to be consistent with DataFabric. It is needed mainly for Update Entity calls especially sqlType We can discuss how to handle it. And I kept the fields optional so that it won't break things

docs/pagination.md
```diff
@@ -121,6 +121,7 @@ console.log(`Total count: ${allAssets.totalCount}`);
 | Jobs | `getAll()` | ✅ Yes |
 | Entities | `getAll()` | ✅ Yes |
 | Entities | `getAllRecords()` | ✅ Yes |
+| Entities | `queryRecordsById()` | ❌ No (use `start`/`limit` params directly) |
```
Sarath1018: looks like it supports pagination please check

AditiGoyalUipath: This is new api I added. Here the pagination support is through API only not via SDK. This I will include incrementally if we need to add SDK pagination to be supported here

Sarath1018 (approved):

src/services/data-fabric/entities.ts
vnaren23, line 1: folderKey is a uuid and folderId is a number. It seems we want uuid but have argument and variable names as folderId. Lets change that throughout the PR.

src/services/data-fabric/entities.ts
```diff
@@ -681,4 +954,45 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+      ...(field.referenceEntityName !== undefined && { referenceEntityName: field.referenceEntityName }),
+      ...(field.referenceFieldName !== undefined && { referenceFieldName: field.referenceFieldName }),
+    };
+  }
+
+  private static readonly RESERVED_FIELD_NAMES = new Set([
+    'Id', 'CreatedBy', 'CreateTime', 'UpdatedBy', 'UpdateTime'
+  ]);
+
+  private validateName(name: string, context: 'entity' | 'field'): void {
```
vnaren23, line 991: Its okay to have these client side validations but I hope the same is validated at backend APIs as well.

src/models/data-fabric/entities.types.ts
```diff
@@ -363,13 +584,15 @@ export interface RawEntityGetResponse {
   entityType: EntityType;
   description: string;
   fields: FieldMetaData[];
+  folderId?: string;
```
vnaren23, line 596: UUID or number? Change accordingly.

src/models/data-fabric/entities.types.ts
```diff
@@ -363,13 +584,15 @@ export interface RawEntityGetResponse {
...
   externalFields?: ExternalSourceFields[];
   sourceJoinCriterias?: SourceJoinCriteria[];
   recordCount?: number;
   storageSizeInMB?: number;
   usedStorageSizeInMB?: number;
   attachmentSizeInByte?: number;
   isRbacEnabled: boolean;
+  isInsightsEnabled?: boolean;
```
vnaren23, line 604: This is internal, why do we need this property?

AditiGoyalUipath: Can an Entity be used for both scenarios?  internal as well as External ?

src/services/data-fabric/entities.ts
AditiGoyalUipath: Yes it is validated

src/services/data-fabric/entities.ts
```diff
@@ -590,24 +692,187 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+   *   { name: "product_name", type: EntityFieldDataType.STRING, isRequired: true, isUnique: true },
+   *   { name: "price", type: EntityFieldDataType.INTEGER, defaultValue: "0" },
+   * ], { displayName: "Product Catalog", isRbacEnabled: true });
+   * ```
+   * @internal
    */
-  async updateById(id: string, data: EntityRecord[], options: EntityUpdateOptions = {}): Promise<EntityUpdateResponse> {
-    return this.updateRecordsById(id, data, options);
+  @track('Entities.Create')
+  async create(name: string, description: string, fields: EntityCreateFieldOptions[], options?: EntityCreateOptions): Promise<string> {
```
vnaren23: Isnt description optional for an entity?

AditiGoyalUipath: Yes description is optional.

src/models/data-fabric/entities.types.ts
vnaren23: Better to rename as `isAnalyticsEnabled`. This is inline to what DF UI shows. 

Image

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: In datafabric We are using FolderId as UUID. Will keep it consistent here as well

vnaren23:
@Sarath1018 Since these are new endpoints we should probably whitelist these APIs in CloudFlare as well. Please raise a PR for the same. 
Add only external method APIs.

AditiGoyalUipath:
> @Sarath1018 Since these are new endpoints we should probably whitelist these APIs in CloudFlare as well. Please raise a PR for the same. Add only external method APIs.
@vnaren23 - We have added Here
https://github.com/UiPath/apps-dev-tools/pull/42

src/models/data-fabric/entities.types.ts
vnaren23: No, we should not be bringing DF intricacies into SDK. SDK is developer facing and forcing them to follow team specific conventions is not a good experience.

vnaren23: At SDK level external facing names should be `folderKey`. While making the API call you can change it to `folderId` using map.

src/services/data-fabric/entities.ts
vnaren23: Its better if we move the description into `options?: EntityCreateOptions`, that way your method signature is 

```
async create(name: string, fields: EntityCreateFieldOptions[], options?: EntityCreateOptions): Promise<string> {
```

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: Make Sense . Will add it as folderKey here. Following the existing  folderKey type adding it as string.

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: So this field will be received as part of Update Metadata. I have kept the name inline with how backend accepts it.

vnaren23: Okay may not be this specific file but whatever is external facing, better rename to `isAnalyticsEnabled`. Like `EntityCreateOptions ` interface in `src/models/data-fabric/entities.types.ts` file

src/services/data-fabric/entities.ts
AditiGoyalUipath: Yes .. I will adding it part of pagination PR

Sarath1018:
> @Sarath1018 Since these are new endpoints we should probably whitelist these APIs in CloudFlare as well. Please raise a PR for the same. Add only external method APIs.

@vnaren23 integration tests will not pass if we don't add there as they use `alpha.api.uipath.com`

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: Wouldn't it be confusing for user . When there will be difference between client API  response and SDK response for this field? If concern is that it is not understandable that isInsightEnabled is same as Support Analytical queries in UI. Wont adding the description about this field be enough?

vnaren23: Naming should itself be simple and to the point. If we can achieve via naming why even add description. TBH no one reads description. 
For the developer, every user facing surface (argument, property etc) can be `isAnalyticsEnabled`. SDK can rename it internally before hitting backend APIs. Since `insights` is our internal convention, we dont need to force it on developers as well.

AditiGoyalUipath: I have included these changes as part of this PR https://github.com/UiPath/uipath-typescript/pull/369/changes

AditiGoyalUipath:
Follow Up PR - https://github.com/UiPath/uipath-typescript/pull/369

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,203 @@ export interface EntityDeleteOptions {
...
+
+/**
+ * User-facing field definition for creating or updating entity schemas
+ */
+export interface EntityCreateFieldOptions extends EntityFieldBase {
+  /**
+   * Technical field name — must start with a letter and contain only
+   * letters, numbers, and underscores (e.g., `"productName"`).
+   */
+  name: string;
```
vnaren23: This has `name` whereas when we query an entity using `EntityQueryFilter`, we take the same input as `fieldName`. Lets rename here also to `fieldName` to match

src/models/data-fabric/entities.models.ts
```diff
@@ -333,11 +331,59 @@ export interface EntityServiceModel {
...
+   *
+   * const entities = new Entities(sdk);
+   *
+   * // Non-paginated query with a filter
+   * const result = await entities.queryRecordsById(<id>, {
+   *   filterGroup: {
+   *     logicalOperator: LogicalOperator.And,
+   *     queryFilters: [{ fieldName: "status", operator: QueryFilterOperator.Equals, value: "active" }]
+   *   },
+   *   sortOptions: [{ fieldName: "created_at", isDescending: true }],
```
vnaren23: nit: this `fieldName: "created_at"` is meant to refer to the system generated field? If yes then it should be `fieldName: "CreateTime"`

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +507,78 @@ export interface EntityServiceModel {
    * ```
    */
   deleteAttachment(entityId: string, recordId: string, fieldName: string): Promise<EntityDeleteAttachmentResponse>;
+
+  /**
+   * Creates a new Data Fabric entity with the given schema
+   *
+   * @param name - Technical entity name — must start with a letter, letters/numbers/underscores only
```
vnaren23: nit: Just call it entity name. Remove Technical.

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +507,78 @@ export interface EntityServiceModel {
...
+   * @example
+   * ```typescript
+   * await entities.deleteById(<id>);
+   * ```
+   * @internal
+   */
+  deleteById(id: string): Promise<void>;
+
+  /**
+   * Updates an existing Data Fabric entity — schema and/or metadata in a single call.
```
vnaren23: Remove `in a single call`

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,202 @@ export interface EntityDeleteOptions {
 export interface EntityDeleteRecordsOptions extends EntityDeleteOptions {}
 
 
+/**
+ * Logical operator for combining query filter groups
+ */
+export enum LogicalOperator {
```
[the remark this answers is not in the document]
vnaren23: @AditiGoyalUipath This seems like valid. You'll need to add `export { LogicalOperator, QueryFilterOperator } from '../../models/data-fabric/entities.types';` to `uipath-typescript\src\services\data-fabric\index.ts` file explicitly

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +507,78 @@ export interface EntityServiceModel {
...
+   * @param id - UUID of the entity to update
+   * @param options - Changes to apply ({@link EntityUpdateByIdOptions})
+   * @returns Promise resolving when the update is complete
+   *
+   * @example
+   * ```typescript
+   * // Schema-only: add a field and remove another
+   * await entities.updateById(<id>, {
+   *   addFields: [{ name: "notes", type: EntityFieldDataType.MULTILINE_TEXT }],
+   *   removeFields: ["old_field"],
```
vnaren23: Cant `removeFields` also take an object like `[{ name: "notes" }]`? Both `updateFields` and `addFields` are taking object arrays, whereas `removeFields` are string[]

AditiGoyalUipath: Since removeFields only requires a single string parameter, we kept it simple instead of converting it into an object. In contrast, addFields and updateFields require additional attributes, so they are designed to take objects.

src/models/data-fabric/entities.models.ts
```diff
@@ -461,6 +503,79 @@ export interface EntityServiceModel {
...
+   * });
+   *
+   * // Combined: update a field and rename at the same time
+   * await entities.updateById("<entityId>", {
+   *   updateFields: [{ id: "<fieldId>", displayName: "Unit Price", isRequired: true }],
+   *   displayName: "Price Catalog",
+   * });
+   * ```
+   * @internal
+   */
```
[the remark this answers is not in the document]
vnaren23: @AditiGoyalUipath This is valid. I tried with curl and can see we only need to pass name and id as mandatory arguments. Rest all are optional. 

Image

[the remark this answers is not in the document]
vnaren23: We should change the method signature to `updateById(id: string, name: string, options?: EntityUpdateByIdOptions): `

Remove name out of `EntityUpdateByIdOptions`

src/models/data-fabric/entities.models.ts
```diff
@@ -588,17 +706,26 @@ export interface EntityMethods {
...
 
-    /**
-   * @deprecated Use {@link updateRecords} instead.
-   * @hidden
+  /**
+   * Queries records in this entity with filters, sorting, and SDK-managed pagination
+   *
+   * @param options - Query options including filterGroup, selectedFields, sortOptions, and pagination
+   * @returns Promise resolving to {@link NonPaginatedResponse} without pagination options,
+   *   or {@link PaginatedResponse} when `pageSize`, `cursor`, or `jumpToPage` are provided
```
vnaren23, line 715: @example missing for this method

src/models/data-fabric/entities.models.ts
```diff
@@ -588,17 +706,26 @@ export interface EntityMethods {
...
+      ? PaginatedResponse<EntityRecord>
+      : NonPaginatedResponse<EntityRecord>
+  >;
 
   /**
-   * @deprecated Use {@link deleteRecords} instead.
-   * @hidden
+   * Imports records from a CSV file into this entity
+   *
+   * @param file - CSV file to import as a Blob, File, or Uint8Array
```
vnaren23, line 738: @example missing here

src/models/data-fabric/entities.models.ts
```diff
@@ -609,16 +726,47 @@ export interface EntityMethods {
...
+   * ```typescript
+   * const entity = await entities.getById(<id>);
+   * await entity.update({
+   *   displayName: "Updated Name",
+   *   addFields: [{ name: "notes", type: EntityFieldDataType.MULTILINE_TEXT }],
+   * });
+   * ```
+   * @internal
+   */
+  update(options: EntityUpdateByIdOptions): Promise<void>;
```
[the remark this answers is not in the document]
vnaren23: Same as above. options argument should be optional.

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,203 @@ export interface EntityDeleteOptions {
...
+export enum QueryFilterOperator {
+  Equals = '=',
+  NotEquals = '!=',
+  GreaterThan = '>',
+  LessThan = '<',
+  GreaterThanOrEqual = '>=',
+  LessThanOrEqual = '<=',
+  Contains = 'contains',
+  NotContains = 'not contains',
+  StartsWith = 'startswith',
```
vnaren23, line 152: Why does `not contains` have a space whereas `startswith` doesnt have space?

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,202 @@ export interface EntityDeleteOptions {
...
+  /** Nested filter groups for complex boolean expressions */
+  filterGroups?: EntityQueryFilterGroup[];
+}
+
+/**
+ * Sort option for query results
+ */
+export interface EntityQuerySortOption {
+  /** Name of the field to sort by */
+  fieldName?: string;
```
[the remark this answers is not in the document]
vnaren23: @AditiGoyalUipath valid point here

src/services/data-fabric/entities.ts
```diff
@@ -577,37 +766,250 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+    const raw = entityResponse.data;
+
+    // Carry forward existing non-system fields from GET response (skip system/primary-key fields)
+    let fields: FieldMetaData[] = (raw.fields ?? []).filter(
+      (f) => !f.isSystemField && !f.isPrimaryKey,
+    );
+
+    // Apply removals
+    if (options.removeFields?.length) {
+      const removeSet = new Set(options.removeFields);
```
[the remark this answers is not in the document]
vnaren23: Do check if this is needed @AditiGoyalUipath

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: Yes this is inconsistent but this is how it is supported in Datafabric backend

src/models/data-fabric/entities.types.ts
```diff
@@ -121,6 +126,203 @@ export interface EntityDeleteOptions {
...
+ *
+ * Schema changes (`addFields`, `removeFields`, `updateFields`) and metadata changes
+ * (`displayName`, `description`, `isRbacEnabled`) can be combined; each is applied
+ * only when the corresponding fields are provided.
+ */
+export interface EntityUpdateByIdOptions {
+  /** New fields to add */
+  addFields?: EntityCreateFieldOptions[];
+  /** Fields to remove, each identified by field name */
+  removeFields?: Array<{ fieldName: string }>;
```
vnaren23: Lets name it properly.


```
export interface EntityRemoveFieldOptions {
  /** Name of the field to remove*/
  fieldName: string;
}
...
...

removeFields?: EntityRemoveFieldOptions[];

```

src/models/data-fabric/entities.models.ts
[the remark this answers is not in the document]
AditiGoyalUipath: Name cannot be updated atleast from UI. Need to check the backend. So Name shouldn't be a parameter for update options. We have displayname part of options which can be updated

vnaren23 (approved):

src/services/data-fabric/index.ts
```diff
@@ -27,6 +27,7 @@ export { ChoiceSetService as ChoiceSets, ChoiceSetService } from './choicesets';
 
 // Re-export service-specific types
 export * from '../../models/data-fabric/entities.types';
+export { LogicalOperator, QueryFilterOperator } from '../../models/data-fabric/entities.types';
```
Sarath1018, line 30: This is redundant. Line 29 already does `export * from '../../models/data-fabric/entities.types'`

PR #363, opened 2026-04-13
feat: get user details from token

vnaren23:
<details>
  <summary>Prompt used</summary>
This is the Typescript SDK repo. We want to add a util method to into src/core/uipath.ts named getTokenIdentity()

The util method should be able to retrieve details of current logged in user from the jwt token stored. We already expose isAuthenticated method. Use that to validate if user has already been authenticated. If not authenticated throw an error. We already expose getToken() method to get the token. Use the same method internally to get the token, decode the jwt token. Response should be consisting of below 5 fields. 
email
first_name
last_name
preferred_username
name

Follow coding and naming conventions mentioned in repo.
</details>

<details>
  <summary>Generated claude plan</summary>

# Plan: Add `getTokenIdentity()` to UiPath core class

## Context

Consumers of the SDK currently have no first-class way to read identity claims (email, name, etc.) for the authenticated user. The only hook is `getToken()`, which returns the raw JWT string — forcing every app to pull in a JWT library and implement base64url decoding themselves. This util exposes the identity claims directly from the token already held in memory, so coded-apps (and any other SDK consumer) can render the current user without an extra round-trip or dependency.

The method is a local, read-only utility — no API call, just decode the JWT payload we already have.

### Method signature

```typescript
public getTokenIdentity(): TokenIdentity
```

Synchronous, no parameters. Mirrors the style of the existing `isAuthenticated()` / `getToken()` methods on [`src/core/uipath.ts`](src/core/uipath.ts) — plain public method, JSDoc block, no `@track` decorator (consistent with other UiPath core class methods — decorators are only used on service methods).

### Return type

New interface `TokenIdentity` added to [src/core/auth/types.ts](src/core/auth/types.ts) (existing home for `TokenInfo`, `AuthToken`, `OAuthContext` — no need for a new file):

```typescript
/**
 * Identity claims decoded from the current JWT access token.
 */
export interface TokenIdentity {
  email?: string;
  firstName?: string;
  lastName?: string;
  preferredUsername?: string;
  name?: string;
}
```

All fields optional — JWT claims are not guaranteed to be present, and marking them required would cause runtime `undefined` access. Raw JWT claims use snake_case (`first_name`, `last_name`, `preferred_username`) and are mapped to camelCase to match the rest of the SDK's public surface.

### Behavior

1. If `!this.isAuthenticated()` → throw `AuthenticationError` with message `'User is not authenticated. Call initialize() before getTokenIdentity().'`
2. Read `this.getToken()`. If undefined (defensive — shouldn't happen when `isAuthenticated()` is true, but possible with secret-auth empty-token path at [`src/core/uipath.ts:120`](src/core/uipath.ts#L120)) → throw `AuthenticationError`.
3. Split the JWT on `.` — must be 3 segments. Otherwise → throw `ValidationError` with message `'Invalid JWT token format.'`
4. Base64url-decode the payload segment (replace `-`→`+`, `_`→`/`, pad with `=` to multiple of 4), then `decodeBase64()` from [src/utils/encoding/base64.ts](src/utils/encoding/base64.ts) and `JSON.parse()`. Any failure in this chain → throw `ValidationError` with message `'Failed to decode JWT token payload.'`
5. Map the 5 snake_case claims to camelCase and return.

### Error type choices

- **Not authenticated** → `AuthenticationError` from [src/core/errors/authentication.ts](src/core/errors/authentication.ts). The error type is designed for "missing authentication" (see class JSDoc).
- **Malformed token / decode failure** → `ValidationError` from [src/core/errors/validation.ts](src/core/errors/validation.ts). Per `agent_docs/conventions.md` error-types section: `ValidationError` is for invalid data format (user-supplied input), `ServerError` is for failed parsing of *API response* data. A malformed JWT sitting in auth state is closer to invalid input than a server response issue. This is a judgment call; using `ServerError` here would also be defensible, but `ValidationError` matches the "invalid data format" description more directly.

### Files to modify

| File | Change |
|------|--------|
| [src/core/auth/types.ts](src/core/auth/types.ts) | Add `TokenIdentity` interface |
| [src/core/uipath.ts](src/core/uipath.ts) | Add `getTokenIdentity()` method after `getToken()` (line 239); add imports for `TokenIdentity`, `AuthenticationError`, `ValidationError`, `decodeBase64` |
| [src/core/index.ts](src/core/index.ts) | Add `export type { TokenIdentity }` alongside existing `TokenInfo` export (line 48) |

No changes needed to [src/index.ts](src/index.ts) — it doesn't explicitly re-export `TokenInfo` either, so `TokenIdentity` follows the same pattern (available via `@uipath/uipath-typescript/core`).

No changes to `package.json`, `rollup.config.js`, `mkdocs.yml` — this is a method on an existing class exported through an existing path, not a new service.

No `docs/oauth-scopes.md` entry — this method makes no API call, so no OAuth scope is required.

No `docs/pagination.md` entry — not a paginated method.

### Reused utilities

- `decodeBase64()` — [src/utils/encoding/base64.ts:33](src/utils/encoding/base64.ts#L33). Cross-platform (browser `atob` / Node `Buffer`).
- `AuthenticationError` — [src/core/errors/authentication.ts](src/core/errors/authentication.ts).
- `ValidationError` — [src/core/errors/validation.ts](src/core/errors/validation.ts).
- `isAuthenticated()` / `getToken()` — existing methods on `UiPath` class.

### JSDoc

Example to include in the method's JSDoc:

```typescript
const sdk = new UiPath({ /* ... */ });
await sdk.initialize();

const identity = sdk.getTokenIdentity();
console.log(identity.email, identity.name);
```

## Tests

Add a new `describe('getTokenIdentity')` block to [tests/unit/core/uipath.test.ts](tests/unit/core/uipath.test.ts). The existing test file already mocks `AuthService` with configurable `hasValidToken` and `getToken` — extend the mock so individual tests can override the returned token.

Cases to cover:
1. Returns all 5 mapped fields when JWT payload contains the claims (uses a fixture JWT assembled with base64url-encoded payload `{ email, first_name, last_name, preferred_username, name }`).
2. Returns fields as `undefined` when claims are missing from the payload.
3. Throws `AuthenticationError` when `isAuthenticated()` is false.
4. Throws `ValidationError` when token is not a valid JWT structure (e.g., `'not.a.jwt'` has 3 segments but the middle is not valid base64url JSON; also test 2-segment and 1-segment strings).
5. Throws `ValidationError` when payload base64 decodes but is not valid JSON.

Use `TEST_CONSTANTS` style; build the fixture JWT inline with a small helper in the test file (no new test util needed for a single use).

## Verification

```bash
npm run typecheck     # must be clean
npm run lint          # 0 errors
npm run test:unit     # all unit tests pass, new tests green
npm run build         # dist/ produced
```

Manual end-to-end check (optional, relies on a live auth setup):

```typescript
const sdk = new UiPath({ /* oauth config */ });
await sdk.initialize();
console.log(sdk.getTokenIdentity());
```

Confirm the console output shows the 5 camelCase fields populated from the real token claims.

</details>
<details>
  <summary>Local Testing</summary>

image
  

  
</details>

### DIFF
--- a/src/core/auth/types.ts
+++ b/src/core/auth/types.ts
@@ -31,3 +31,14 @@ export interface OAuthContext {
   tenantName: string;
   scope: string;
 }
+
+/**
+ * Identity claims decoded from the current JWT access token.
+ */
+export interface TokenIdentity {
+  email?: string;
+  firstName?: string;
+  lastName?: string;
+  preferredUsername?: string;
+  name?: string;
+}
--- a/src/core/uipath.ts
+++ b/src/core/uipath.ts
@@ -1,7 +1,7 @@
 import { UiPathConfig } from './config/config';
 import { ExecutionContext } from './context/execution';
 import { AuthService } from './auth/service';
-import { TokenInfo } from './auth/types';
+import { TokenInfo, TokenIdentity } from './auth/types';
 import { UiPathSDKConfig, PartialUiPathConfig, BaseConfig, hasOAuthConfig, hasSecretConfig } from './config/sdk-config';
 import { validateConfig, normalizeBaseUrl, isCompleteConfig } from './config/config-utils';
 import { telemetryClient, trackEvent } from './telemetry';
@@ -9,6 +9,8 @@ import { SDKInternalsRegistry } from './internals';
 import { loadFromMetaTags } from './config/runtime';
 import type { IUiPath } from './types';
 import { isInActionCenter } from '../utils/platform';
+import { decodeBase64 } from '../utils/encoding/base64';
+import { AuthenticationError, ValidationError } from './errors';
 
 /**
  * UiPath - Core SDK class for authentication and configuration management.
@@ -236,6 +238,65 @@ export class UiPath implements IUiPath {
    */
   public getToken(): string | undefined {
     return this.#authService?.getToken();
+  }
+
+  /**
+   * Retrieves identity claims of the currently authenticated user by decoding
+   * the JWT access token held in memory. Does not make an API call.
+   *
+   * Returns the following camelCase claims when present on the token:
+   * `email`, `firstName`, `lastName`, `preferredUsername`, `name`.
+   *
+   * @returns The {@link TokenIdentity} extracted from the JWT payload.
+   * @throws {@link AuthenticationError} If the user is not authenticated.
+   * @throws {@link ValidationError} If the token is malformed or its payload cannot be decoded.
+   *
+   * @example
+   * ```typescript
+   * const sdk = new UiPath({ ...config });
+   * await sdk.initialize();
+   *
+   * const identity = sdk.getTokenIdentity();
+   * console.log(identity.email, identity.name);
+   * ```
+   */
+  public getTokenIdentity(): TokenIdentity {
+    if (!this.isAuthenticated()) {
+      throw new AuthenticationError({
+        message: 'User is not authenticated. Call initialize() before getTokenIdentity().'
+      });
+    }
+
+    const token = this.getToken();
+    if (!token) {
+      throw new AuthenticationError({
+        message: 'User is not authenticated. Call initialize() before getTokenIdentity().'
+      });
+    }
+
+    const segments = token.split('.');
+    if (segments.length !== 3) {
+      throw new ValidationError({ message: 'Invalid JWT token format.' });
+    }
+
+    let claims: Record<string, unknown>;
+    try {
+      // Convert base64url to base64 and pad to a multiple of 4.
+      let payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
+      const paddingLength = (4 - (payload.length % 4)) % 4;
+      payload = payload + '='.repeat(paddingLength);
+      claims = JSON.parse(decodeBase64(payload));
+    } catch {
+      throw new ValidationError({ message: 'Failed to decode JWT token payload.' });
+    }
+
+    return {
+      email: claims.email as string | undefined,
+      firstName: claims.first_name as string | undefined,
+      lastName: claims.last_name as string | undefined,
+      preferredUsername: claims.preferred_username as string | undefined,
+      name: claims.name as string | undefined
+    };
   }
 
   /**
--- a/tests/unit/core/uipath.test.ts
+++ b/tests/unit/core/uipath.test.ts
@@ -443,6 +443,19 @@ describe('UiPath Core', () => {
       expect(() => sdk.getTokenIdentity()).toThrow(ValidationError);
     });
 
+    it('should throw ValidationError when token has more than 3 segments', () => {
+      mockAuthState.token = 'header.payload.sig.extra';
+
+      const sdk = new UiPath({
+        baseUrl: TEST_CONSTANTS.BASE_URL,
+        orgName: TEST_CONSTANTS.ORGANIZATION_ID,
+        tenantName: TEST_CONSTANTS.TENANT_ID,
+        secret: TEST_CONSTANTS.CLIENT_SECRET
+      });
+
+      expect(() => sdk.getTokenIdentity()).toThrow(ValidationError);
+    });
+
     it('should throw ValidationError when payload is not valid JSON', () => {
       const header = Buffer.from(JSON.stringify({ alg: 'none' })).toString('base64url');
       const invalidPayload = Buffer.from('not-json-content').toString('base64url');
--- a/tests/unit/core/uipath.test.ts
+++ b/tests/unit/core/uipath.test.ts
@@ -3,13 +3,19 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { UiPath } from '../../../src/core/uipath';
 import { UiPathConfig } from '../../../src/core/config/config';
 import { ExecutionContext } from '../../../src/core/context/execution';
+import { AuthenticationError, ValidationError } from '../../../src/core/errors';
 import { getConfig, getContext, getTokenManager, getPrivateSDK } from '../../utils/setup';
 import { TEST_CONSTANTS } from '../../utils/constants/common';
 
 // ===== MOCKING =====
+const mockAuthState = {
+  token: 'mock-access-token' as string | undefined,
+  hasValidToken: true
+};
+
 const mockTokenManager = {
-  getToken: () => 'mock-access-token',
-  hasValidToken: () => true
+  getToken: () => mockAuthState.token,
+  hasValidToken: () => mockAuthState.hasValidToken
 };
 
 const mockLogout = vi.fn();
@@ -17,8 +23,8 @@ const mockLogout = vi.fn();
 vi.mock('../../../src/core/auth/service', () => {
   const AuthService: any = vi.fn().mockImplementation(() => ({
     getTokenManager: () => mockTokenManager,
-    hasValidToken: () => true,
-    getToken: () => 'mock-access-token',
+    hasValidToken: () => mockAuthState.hasValidToken,
+    getToken: () => mockAuthState.token,
     authenticateWithSecret: vi.fn(),
     authenticate: vi.fn().mockResolvedValue(true),
     logout: mockLogout
@@ -33,6 +39,13 @@ vi.mock('../../../src/core/auth/service', () => {
 });
 
 vi.mock('../../../src/core/http/api-client');
+
+// ===== TEST HELPERS =====
+const createJwt = (payload: Record<string, unknown>): string => {
+  const header = Buffer.from(JSON.stringify({ alg: 'none', typ: 'JWT' })).toString('base64url');
+  const body = Buffer.from(JSON.stringify(payload)).toString('base64url');
+  return `${header}.${body}.signature`;
+};
 
 // ===== TEST SUITE =====
 describe('UiPath Core', () => {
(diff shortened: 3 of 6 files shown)


src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+   * @example
+   * ```typescript
+   * const sdk = new UiPath({ ...config });
+   * await sdk.initialize();
+   *
+   * const identity = sdk.getTokenIdentity();
+   * console.log(identity.email, identity.name);
+   * ```
+   */
+  public getTokenIdentity(): TokenIdentity {
```
maninder-uipath, line 263: `getTokenIdentity` seems very technical name. IMO it should be `getUserDetails`

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+      let payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
+      const paddingLength = (4 - (payload.length % 4)) % 4;
+      payload = payload + '='.repeat(paddingLength);
+      claims = JSON.parse(decodeBase64(payload));
+    } catch {
+      throw new ValidationError({ message: 'Failed to decode JWT token payload.' });
+    }
+
+    return {
+      email: claims.email as string | undefined,
```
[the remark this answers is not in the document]
vnaren23: None of these values are ever a non string.

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+   */
+  public getTokenIdentity(): TokenIdentity {
+    if (!this.isAuthenticated()) {
+      throw new AuthenticationError({
+        message: 'User is not authenticated. Call initialize() before getTokenIdentity().'
+      });
+    }
+
+    const token = this.getToken();
+    if (!token) {
```
maninder-uipath, line 271: Is there any scenario possible where `this.isAuthenticated()` is true but token is undefined/null? If yes, then we should handle that scenario.

src/core/uipath.ts
vnaren23: `getUserDetails` might seem like we are querying our backend for user info. Since all we are doing is decoding a token, used `getTokenIdentity`

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+      throw new ValidationError({ message: 'Invalid JWT token format.' });
+    }
+
+    let claims: Record<string, unknown>;
+    try {
+      // Convert base64url to base64 and pad to a multiple of 4.
+      let payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
+      const paddingLength = (4 - (payload.length % 4)) % 4;
+      payload = payload + '='.repeat(paddingLength);
+      claims = JSON.parse(decodeBase64(payload));
```
maninder-uipath, line 288: Is this logic future proof? Should we rely on a well known lib for decoding & parsing token?

src/core/uipath.ts
maninder-uipath: then by this logic, even `getToken` method can also be interpreted by someone in the same manner

src/core/uipath.ts
vnaren23: The if condition is checking the exact scenario you are saying isnt it?

src/core/uipath.ts
vnaren23: jwt tokens are base64encoded. No need to complicate with separate libs for it. Will only bloat SDK.

src/core/uipath.ts
vnaren23: Yes, even this method is on top of `getToken`. Just a helper for easier SDK usage, nothing more.

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+    } catch {
+      throw new ValidationError({ message: 'Failed to decode JWT token payload.' });
```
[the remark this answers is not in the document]
vnaren23: Not necessarily. SDK takes both OAuth (JWT) and PAT token (not JWT). Since we cannot decode a PAT token, this can remain as validation error.

tests/unit/core/uipath.test.ts
```diff
@@ -352,6 +365,100 @@ describe('UiPath Core', () => {
...
+        baseUrl: TEST_CONSTANTS.BASE_URL,
+        orgName: TEST_CONSTANTS.ORGANIZATION_ID,
+        tenantName: TEST_CONSTANTS.TENANT_ID,
+        secret: TEST_CONSTANTS.CLIENT_SECRET
+      });
+
+      expect(() => sdk.getTokenIdentity()).toThrow(AuthenticationError);
+    });
+
+    it('should throw ValidationError when token has fewer than 3 segments', () => {
```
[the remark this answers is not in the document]
vnaren23: Done.

src/core/uipath.ts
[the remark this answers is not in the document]
vnaren23: Done.

src/core/uipath.ts
maninder-uipath: I meant to ask how will this scenario even occur? It should never be the case where this.isAuthenticated() is true but token is undefined/null

src/core/uipath.ts
maninder-uipath: from a user's POV, I dont like this name

vnaren23: I'm open to changing. Maybe `getTokenUserDetails` ?

swati354: How about `getLoggedInUser`/ `getAuthenticatedUser`?
Since we have `isAuthenticated` method already, `getAuthenticatedUser` makes sense.

swati354: > getUserDetails might seem like we are querying our backend for user info. Since all we are doing is decoding a token, used getTokenIdentity

@vnaren23 question - Why does the user have to care about how the details are being fetched. If anything, shouldn't the name describe what the user gets, and not how they get it?
The "users might think it's a network call" worry could be solved with a JSDoc comment?

src/core/auth/types.ts
```diff
@@ -31,3 +31,14 @@ export interface OAuthContext {
   tenantName: string;
   scope: string;
 }
+
+/**
+ * Identity claims decoded from the current JWT access token.
+ */
+export interface TokenIdentity {
```
swati354, line 38: 1. Can we rename this as well?
2. Is there any other info we get from the token that we can put here as well - like user-id ?
3. Why are all fields optional?

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
     return this.#authService?.getToken();
   }
 
+  /**
+   * Retrieves identity claims of the currently authenticated user by decoding
+   * the JWT access token held in memory. Does not make an API call.
+   *
+   * Returns the following camelCase claims when present on the token:
+   * `email`, `firstName`, `lastName`, `preferredUsername`, `name`.
```
swati354, line 248: 1. No need for this comment, the `@returns` tag takes care of this.
2. Since these comments will surface on the docs, can we add `Does not work with PAT tokens` info here? or to be more specific `only supported for OAuth user sessions` ?

src/core/uipath.ts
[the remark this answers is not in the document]
swati354: the JWT being decoded didn't come from the user, it comes from the identity. The caller of the method passes nothing. So how is validation error a right fit?

src/core/uipath.ts
swati354: Is this check for when the sdk is used inside Action Center? Probably then the user is authenticated and does not have a token until the api call is made.

src/core/uipath.ts
```diff
@@ -238,6 +240,65 @@ export class UiPath implements IUiPath {
...
+    if (segments.length !== 3) {
+      throw new ValidationError({ message: 'Invalid JWT token format.' });
+    }
+
+    let claims: Record<string, unknown>;
+    try {
+      // Convert base64url to base64 and pad to a multiple of 4.
+      let payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
+      const paddingLength = (4 - (payload.length % 4)) % 4;
+      payload = payload + '='.repeat(paddingLength);
```
swati354, line 287: should we create a helper for this in `base64.ts` ?

src/core/uipath.ts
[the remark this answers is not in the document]
swati354: nit/thought: if you're using `unknown`, do the work `unknown` exists to make you do. The whole point of it is that it forces a check before narrowing.

src/core/uipath.ts
vnaren23: There actually could be a getUserDetails API call which we might be masking due to this implementation. Hence my hesitance in using `getUserDetails`. `getAuthenticatedUser` is more in tune to getting details from token

src/core/uipath.ts
vnaren23: Thats one possible scenario yes

src/core/uipath.ts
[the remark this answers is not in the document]
vnaren23: We can set token, could be a PAT token or anything random.

src/core/uipath.ts
vnaren23: We are not doing any decoding in these lines.

PR #367, opened 2026-04-13
feat(cas): agent inputs in conversations

scottcmg:
## Description
Conversational agents allow for agent inputs set by the user to be passed inline when conversations are created and updated. This change is to add support for this inline input feature, and to keep the implementation flexible for future external value support.

### DIFF
--- a/src/models/conversational-agent/conversations/types/common.types.ts
+++ b/src/models/conversational-agent/conversations/types/common.types.ts
@@ -135,6 +135,12 @@ export interface ExternalValue {
 export type InlineOrExternalValue<T> = InlineValue<T> | ExternalValue;
 
 /**
+ * Input arguments passed to the agent on each exchange. Only inline values are currently supported by the API.
+ * Using the InlineOrExternalValue type to make future external value support easier.
+ */
+export type AgentInput = InlineOrExternalValue<JSONObject>;
+
+/**
  * Tool call input value type.
  */
 export type ToolCallInputValue = JSONObject;
--- a/src/models/conversational-agent/conversations/types/common.types.ts
+++ b/src/models/conversational-agent/conversations/types/common.types.ts
@@ -135,10 +135,11 @@ export interface ExternalValue {
 export type InlineOrExternalValue<T> = InlineValue<T> | ExternalValue;
 
 /**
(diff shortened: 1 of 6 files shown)


maninder-uipath (approved):

maxduu (approved):

src/models/conversational-agent/conversations/types/common.types.ts
```diff
@@ -134,6 +134,12 @@ export interface ExternalValue {
  */
 export type InlineOrExternalValue<T> = InlineValue<T> | ExternalValue;
 
+/**
+ * Input arguments passed to the agent on each exchange. Only inline values are currently supported by the API.
+ * Using the InlineOrExternalValue type to make future external value support easier.
```
swati354, line 139: When are we expecting the external value support to come? If the external support is "someday, maybe", it would be worth narrowing it down to inline?

src/models/conversational-agent/conversations/types/common.types.ts
```diff
@@ -134,6 +134,12 @@ export interface ExternalValue {
  */
 export type InlineOrExternalValue<T> = InlineValue<T> | ExternalValue;
 
+/**
+ * Input arguments passed to the agent on each exchange. Only inline values are currently supported by the API.
```
maxduu: @scottcmg Could we add comments similar to the [swagger](https://alpha.uipath.com/9568beba-50a8-49d1-801e-f2d1711089fd/77de6ae2-0968-4a9d-82a4-9ae4b67f4da3/autopilotforeveryone_/api/v1/ui) which has:
```
Input arguments passed in to the execution of the agent on each exchange. The input arguments are expected to match the input-schema defined in the Agent definition. Currently, only inline values are supported and the total serialized JSON payload must be less than 4,000 characters.
```

src/models/conversational-agent/conversations/types/common.types.ts
maxduu: External value support is not currently planned. Hm, if we just use `AgentInput = InlineValue<JSONObject>`, I guess it would be forward compatible here? If so then yes that's a good point. @scottcmg

swati354 (approved):

PR #368, opened 2026-04-14
fix: isolate attachment integration tests with per-run record creation [PLT-101734]

Sarath1018:
### Problem

The attachment integration tests used a fixed DATA_FABRIC_TEST_RECORD_ID env var, meaning all tests — across every concurrent runner — operated on the same record and the same attachment field slot. Since the API stores exactly one file per field per record (upload replaces), concurrent runs caused race conditions:
- A deleteAttachment test in one runner could delete the file that a downloadAttachment/deleteAttachment test in another runner had just uploaded, causing 404s.

### Changes

- Per-run record isolation: a fresh record is inserted in beforeAll and its ID stored in a scoped recordId variable. Every test in the suite uses that record — concurrent runners each get their own, so they never share state.
- Cleanup in afterAll: the created record is deleted after all tests finish. registerResource also registers it as a safety net for crash scenarios.
- Removed DATA_FABRIC_TEST_RECORD_ID env var: no longer needed. Only DATA_FABRIC_TEST_ENTITY_ID and DATA_FABRIC_TEST_ATTACHMENT_FIELD are required.
- Removed retryWithBackoff: the retry existed because a download in one test could race with a delete in another. With per-test uploads immediately before each download within the same test body, the file is guaranteed to exist.


tests/integration/shared/data-fabric/attachment.integration.test.ts
```diff
@@ -33,16 +33,40 @@ describe.skipIf(!hasAttachmentConfig).each(modes)(
   (mode) => {
     setupUnifiedTests(mode);
 
+    let recordId: string;
```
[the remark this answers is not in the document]
Sarath1018: `let recordId!: string;`
Is better as it will guarantee this is assigned before use @claude

tests/integration/shared/data-fabric/attachment.integration.test.ts
```diff
@@ -68,105 +90,89 @@ describe.skipIf(!hasAttachmentConfig).each(modes)(
...
+          recordId,
           ATTACHMENT_CONFIG.fieldName,
         );
 
-        expect(downloadedFile).toBeDefined();
+        expect(result).toBeDefined();
       });
 
-      it('should upload and then download an attachment via entity method', async () => {
+      it('should upload and then delete an attachment via entity method', async () => {
```
vnaren23: Why did this change from upload-download to upload-delete?

vnaren23 (commented):
Minor clarification, rest LGTM

vnaren23 (approved):

tests/integration/shared/data-fabric/attachment.integration.test.ts
Sarath1018: I swapped positions of download testcase and delete while testing. Reverted it as it is causing unnecessary confusion

deepeshrai-tech (approved):

PR #378, opened 2026-04-16
docs: add Attachments scopes to Jobs.getOutput()

ninja-shreyash:
## Summary
- `Jobs.getOutput()` internally calls `Attachments.getById()` for blob downloads
- Added `OR.Folders` / `OR.Folders.Read` scopes to the `getOutput()` row in oauth-scopes docs so users know the full scope requirements

## Test plan
- [ ] Verify oauth-scopes.md renders correctly on docs site

### DIFF
--- a/docs/oauth-scopes.md
+++ b/docs/oauth-scopes.md
@@ -15,7 +15,7 @@ This page lists the specific OAuth scopes required in external app for each SDK 
 |--------|-------------|
 | `getAll()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getById()` | `OR.Jobs` or `OR.Jobs.Read` |
-| `getOutput()` | `OR.Jobs` or `OR.Jobs.Read` |
+| `getOutput()` | `OR.Jobs` or `OR.Jobs.Read`, `OR.Folders` or `OR.Folders.Read` |
 
 ## Attachments
 


vnaren23 (approved):

docs/oauth-scopes.md
```diff
@@ -15,7 +15,7 @@ This page lists the specific OAuth scopes required in external app for each SDK
 |--------|-------------|
 | `getAll()` | `OR.Jobs` or `OR.Jobs.Read` |
 | `getById()` | `OR.Jobs` or `OR.Jobs.Read` |
-| `getOutput()` | `OR.Jobs` or `OR.Jobs.Read` |
+| `getOutput()` | `OR.Jobs` or `OR.Jobs.Read`, `OR.Folders` or `OR.Folders.Read` |
```
Raina451, line 18: if both OR.Jobs and OR.Folders are needed, then better add:

`OR.Jobs OR.Folders` or `OR.Jobs.Read OR.Folders.Read`

else it;s not clear

Raina451: similar to what we have in : https://github.com/UiPath/uipath-typescript/blob/6d3393c956048baec769f229ef4cc957088c6973/docs/oauth-scopes.md#maestro-case-instances

ninja-shreyash: But everywhere else - https://github.com/UiPath/uipath-typescript/blob/6d3393c956048baec769f229ef4cc957088c6973/docs/oauth-scopes.md#conversational-agent, we have this convention and it makes more sense to be honest.

Since either one of the first pair and either one of second pair is enough.

Raina451 (approved):

PR #379, opened 2026-04-16
fix: Added POST request param as part of Body same as non paginated r…

AditiGoyalUipath:
Fixed the paginated request for QueryById—the filter parameters were being sent as string in the POST request query params. Updated it to pass them in the request body, consistent with the non-paginated request.

### DIFF
--- a/src/utils/pagination/helpers.ts
+++ b/src/utils/pagination/helpers.ts
@@ -185,13 +185,16 @@ export class PaginationHelpers {
     const endpoint = getEndpoint(folderId);
     const headers = folderId ? createHeaders({ [FOLDER_ID]: folderId }) : {};
 
+    const isPost = method?.toUpperCase() === HTTP_METHODS.POST.toUpperCase();
     const paginatedResponse = await serviceAccess.requestWithPagination<T>(
       method,
       endpoint,
       paginationParams,
       {
         headers,
-        params: additionalParams,
+        // POST: body
+        // GET: query params 
+        ...(isPost ? { body: additionalParams, params: {} } : { params: additionalParams }),
         pagination: {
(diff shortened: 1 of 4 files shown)


tests/unit/utils/pagination/helpers.test.ts
```diff
@@ -525,6 +526,32 @@ describe('PaginationHelpers Unit Tests', () => {
...
+        options: { paginationType: PaginationType.OFFSET },
+      });
+
+      expect(mockServiceAccess.requestWithPagination).toHaveBeenCalledWith(
+        HTTP_METHODS.POST,
+        PAGINATION_TEST_CONSTANTS.ENDPOINT_API_ITEMS,
+        { pageSize: PAGINATION_TEST_CONSTANTS.PAGE_SIZE },
+        expect.objectContaining({
+          body: { filterGroup },
+          params: {},
```
Sarath1018, line 551: Also verify that whether base.ts merges the pagination params (limit, start) into the body

src/utils/pagination/helpers.ts
```diff
@@ -185,13 +185,16 @@ export class PaginationHelpers {
...
     const paginatedResponse = await serviceAccess.requestWithPagination<T>(
       method,
       endpoint,
       paginationParams,
       {
         headers,
-        params: additionalParams,
+        // POST: body
+        // GET: query params 
+        ...(isPost ? { body: additionalParams, params: {} } : { params: additionalParams }),
```
Sarath1018, line 197: should we use `params: undefined` instead of empty object?

vnaren23:
This is not required anymore. Fixed as part of https://github.com/UiPath/uipath-typescript/pull/382

PR #382, opened 2026-04-20
fix: reset options.params to an empty object

vnaren23:
Fix for DF query method which was leaking params as query params in POST call

### DIFF
--- a/src/services/base.ts
+++ b/src/services/base.ts
@@ -164,7 +164,7 @@ export class BaseService {
     // Prepare request parameters based on pagination type
     const requestParams = this.preparePaginationRequestParams(paginationType, params, options.pagination);
 
-    // For POST requests, merge pagination params into body; for GET, use query params
+    // For POST requests, merge pagination params into body and set params to undefined; for GET, use query params
     if (method.toUpperCase() === 'POST') {
       const existingBody = (options.body && typeof options.body === 'object') ? options.body as Record<string, unknown> : {};
       options.body = {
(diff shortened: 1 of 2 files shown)


src/services/base.ts
```diff
@@ -172,6 +172,7 @@ export class BaseService {
         ...options.params,
         ...requestParams
       };
+      options.params = {};
```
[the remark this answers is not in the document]
vnaren23: Done.

maninder-uipath (approved):

src/services/base.ts
```diff
@@ -164,14 +164,15 @@ export class BaseService {
...
-    // For POST requests, merge pagination params into body; for GET, use query params
+    // For POST requests, merge pagination params into body and set params to undefined; for GET, use query params
     if (method.toUpperCase() === 'POST') {
       const existingBody = (options.body && typeof options.body === 'object') ? options.body as Record<string, unknown> : {};
       options.body = {
         ...existingBody,
         ...options.params,
         ...requestParams
       };
+      options.params = undefined;
```
Raina451, line 175: One suggestion: instead of modifying the options object directly (options.params = undefined), consider passing a clean copy to request.

Right now we are changing the original options object that the caller passed in. This works today because no caller checks `options.params` after this call, but it could cause issue later if someone does.

vnaren23: Not clear on how creating a new request would help. We were already modifying `options.body` and `options.params`.

vnaren23: One could point another way too. Currently caller at least knows what changes are being made to `options`. If we make a new object for caller to know what changes were done, we'll need to return the new object as well isnt it?

Raina451 (approved):

PR #384, opened 2026-04-21
feat(cas): add surface data via headers to identify origin of the requests in telemetry

scottcmg:
## Description
Add headers to CAS requests to allow for surface name and surface version to be passed on request to identify the origin of the requests. Information is set in the consumer's code and passed via these headers.


maxduu:
Overall looks good to me but this should be merged after CAS reaches prod with the changes and we've verified there, correct?

scottcmg:
> Overall looks good to me but this should be merged after CAS reaches prod with the changes and we've verified there, correct?

I actually think it's safe to merge in ahead of the backend changes. We can test with curl/postman to confirm what happens when the unknown headers are sent but I think the backend will just ignore them.

src/models/conversational-agent/conversational-agent.types.ts
```diff
@@ -10,6 +10,21 @@ export interface ConversationalAgentOptions {
    * external app client; omit for standard UiPath user tokens.
    */
   externalUserId?: string;
+  /**
```
ninja-shreyash, line 13: How is the end user going to pass the `surfaceName` and `surfaceVersion` fields? 
```
const conversationalAgent = new ConversationalAgent(sdk, {surfaceName: "name", surfaceVersion: "version"});
```

Is there any other way apart from constructor?

Asking because none of the other services expect any options to be passed as part of constructor

src/models/conversational-agent/conversational-agent.types.ts
```diff
@@ -10,6 +10,21 @@ export interface ConversationalAgentOptions {
    * external app client; omit for standard UiPath user tokens.
    */
   externalUserId?: string;
```
ninja-shreyash, line 12: Why is this not internal? Does the user have to pass this always?

src/services/conversational-agent/conversations/exchanges.ts
```diff
@@ -67,7 +72,12 @@ export class ExchangeService extends BaseService implements ExchangeServiceModel
    * @param options - Optional configuration (e.g. externalUserId for external app auth)
    */
   constructor(instance: IUiPath, options?: ConversationalAgentOptions) {
-    super(instance, options?.externalUserId ? { [EXTERNAL_USER_ID]: options.externalUserId } : undefined);
+    const defaultHeaders = createHeaders({
+      [EXTERNAL_USER_ID]: options?.externalUserId,
+      [CONVERSATIONAL_SURFACE_NAME]: options?.surfaceName,
+      [CONVERSATIONAL_SURFACE_VERSION]: options?.surfaceVersion,
+    });
+    super(instance, Object.keys(defaultHeaders).length > 0 ? defaultHeaders : undefined);
```
ninja-shreyash: We should extract this duplicated logic within `src/services/conversational-agent/helpers/header.ts` and use it within service constructors - 

```
super(instance, buildCasDefaultHeaders(options));
```

src/models/conversational-agent/conversational-agent.types.ts
scottcmg: This is required on all requests using a client credential grant type token (issued to app-scoped external apps). CAS requires a user id for conversational access/ownership storage and checks and typically uses the user id from the token. CC tokens though don't have user claims so we currently require all requests using that token type to pass the user id in a header. This is required for any consumer that uses a non-user token type so not strictly internal but also not required by all consumers.

src/models/conversational-agent/conversational-agent.types.ts
scottcmg: I see the asymmetry comparing to other services, but not sure if there are any clean alternatives that don't require some bigger restructuring/changes either here or in the widget sdk. Doing this per-method is a big change in the widget sdk, and for direct consumers it can become difficult to always remember to include the external user id (with app tokens without this header the call will fail in our backend). Lifting it into the UiPath object puts service-specific config on the top-level SDK object, which as you've pointed out already other services don't need. There could be other options that I'm overlooking, happy to discuss this further if you have any ideas for improvement or you think one of the above alternatives is a better option.

src/services/conversational-agent/conversations/conversations.ts
```diff
@@ -44,6 +43,7 @@ import { transformData, transformRequest, arrayDictionaryToRecord } from '@/util
 // Local imports
 import {
   ConversationEventHelperManagerImpl,
+  buildCasDefaultHeaders,
```
deepeshrai-tech: CAS is the internal abbreviation; the user-facing name in this SDK is Conversational Agent. Since this helper is internal-only this is borderline — but other helpers in the same folder all use full names. Maybe use buildConversationalAgentDefaultHeaders or simply buildDefaultHeaders .

deepeshrai-tech (approved):

src/models/conversational-agent/conversational-agent.types.ts
ninja-shreyash: UiPath object shouldn't carry service-specific config and per-method is also not clean. For now this is fine

ninja-shreyash (approved):

src/models/conversational-agent/conversational-agent.types.ts
```diff
@@ -10,6 +10,21 @@ export interface ConversationalAgentOptions {
...
+  /**
+   * Optional identifier used in UiPath logs to identify the implementing service
+   * of requests. External consumers do not need to set it; the server logs
+   * missing values as "unknown".
+   *
+   * @internal Intended for UiPath first-party surfaces.
```
vnaren23: Lets just tag as @internal. No need to comment as "Intended for UiPath first-party surfaces."

src/services/conversational-agent/helpers/header.ts
```diff
@@ -0,0 +1,31 @@
...
+  EXTERNAL_USER_ID,
+} from '@/utils/constants/headers';
+import { createHeaders } from '@/utils/http/headers';
+
+/**
+ * Builds the default request headers for Conversational Agent service calls
+ * from the provided options. Returns `undefined` when no headers apply, so
+ * the SDK does not pass an empty `headers` bag down to the HTTP layer.
+ */
+export function buildConversationalAgentDefaultHeaders(
```
vnaren23: Lets drop the `Default` and name it `buildConversationalAgentHeaders`

vnaren23 (commented):
Minor comments, rest LGTM

vnaren23 (approved):

PR #385, opened 2026-04-22
fix: pick user for task assignment

vnaren23:
Tests were blindly picking 1st element from array. Rather pick user specifically for tests.

### DIFF
--- a/src/models/action-center/tasks.types.ts
+++ b/src/models/action-center/tasks.types.ts
@@ -9,6 +9,7 @@ export interface UserLoginInfo {
   emailAddress: string;
   displayName: string;
   id: number;
+  type: TaskUserType;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/models/action-center/tasks.types.ts
```diff
@@ -268,3 +269,25 @@ export interface TaskGetByIdOptions extends BaseOptions {
...
+
+    /**  A user of this type is automatically created when adding a robot, is associated with Robot role and it is used by a robot when communicating with Orchestrator. */
+    Robot = 'Robot',
+
+    /**  A user of type Directory User */
+    DirectoryUser = 'DirectoryUser',
+
+    /** A user of type Directory Group */
+    DirectoryGroup = 'DirectoryGroup',
+
```
Raina451: why did this new enum needed only for tests and not for sdk users?

vnaren23: Its not only tests. enum is present in SDK types only. 

We needed a way to filter. I've introduced the enum so that we can filter.

Raina451: ok so does this mean, getusers method returns not only users but robot, externap app etc also?

Raina451: please check if method description here needs update: https://uipath.github.io/uipath-typescript/api/interfaces/TaskServiceModel/#getusers

vnaren23: > ok so does this mean, getusers method returns not only users but robot, externap app etc also?

Yes.

Will update method description

vnaren23: Done.

src/models/action-center/tasks.types.ts
```diff
@@ -2,13 +2,29 @@ import { BaseOptions, RequestOptions } from "../common/types";
...
+}
+
 export interface UserLoginInfo {
   name: string;
   surname: string;
   userName: string;
   emailAddress: string;
   displayName: string;
   id: number;
+  type: TaskUserType;
```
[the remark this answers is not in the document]
vnaren23: All references in Orchestrator use the same dto
image

ninja-shreyash (approved):

deepeshrai-tech (approved):

Raina451 (approved):

PR #386, opened 2026-04-22
feat(orchestrator): getByName for Assets + Processes, encoded folder header, meta-tag fallback [PLT-92768][PLT-92817][PLT-92813]

deepeshrai-tech:
## Summary

Prerequisite SDK work for the coded-apps bindings/overwrites initiative (PLT-92768 / PLT-93599). Adds name-based resource lookup for two Orchestrator services and the folder-context plumbing the runtime overwrite flow depends on.

- **`assets.getByName(name, options?)`** — OData `$filter=Name eq '…'` + `$top=1` on the folder-scoped Assets endpoint.
- **`processes.getByName(name, options?)`** — same shape against `/odata/Releases`.
- **`FolderScopedOptions`** — new shared type in `models/common/types.ts`. Three explicit fields:
  - `folderId?: number` → `X-UIPATH-OrganizationUnitId`
  - `folderKey?: string` → `X-UIPATH-FolderKey`
  - `folderPath?: string` → `X-UIPATH-FolderPath-Encoded`
  When more than one is supplied, all corresponding headers are forwarded and the server resolves precedence.
- **Folder header resolver** — `resolveFolderHeaders()` in `src/utils/folder/folder-headers.ts`. Centralizes folder-context routing so future folder-scoped methods (Queues, Buckets, Jobs, …) reuse one implementation. Each option field maps directly to its header — no auto-detection.
- **Encoding utility** — `encodeFolderPathHeader()` in `src/utils/encoding/folder-path.ts`. Folder paths are base64-of-UTF-16-LE bytes (matching `Encoding.Unicode` in Orchestrator). Cross-platform safe via `DataView.setUint16(..., true)`.
- **Shared lookup helper** — `FolderScopedService.getByNameLookup<TRaw, T>(...)` encapsulates name validation → header resolution → OData filter → empty-result `NotFoundError`. Both `getByName` implementations are ~10 lines each on top of it.
- **Init-time folder context via meta tag** — SDK reads `<meta name="uipath:folder-key">` (auto-injected by jamjam on coded-app deployments). Used as the default folder context when the caller doesn't supply one. `folderKey` is **not** accepted via the SDK constructor; it lives on the runtime registry (`PrivateSDK.folderKey`), populated from the meta tag only.
- **Validation** — `validateName()` in `src/utils/validation/name-validator.ts` trims and rejects empty / whitespace-only names with a `ValidationError`.
- **NotFoundError** messages now include folder context (id / key / path) for easier debugging.

## Usage

```typescript
import { Assets } from '@uipath/uipath-typescript/assets';
import { Processes } from '@uipath/uipath-typescript/processes';

const assets = new Assets(sdk);
const processes = new Processes(sdk);

// By folder ID
await assets.getByName('ApiKey', { folderId: 123 });

// By folder key (GUID)
await assets.getByName('ApiKey', { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });

// By folder path
await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });

// With OData query parameters
await assets.getByName('ApiKey', { folderPath: 'Shared/Finance', expand: 'KeyValueList' });

// Coded-app — folder resolved automatically from `<meta name="uipath:folder-key">`
await assets.getByName('ApiKey');

// Same convention for processes
await processes.getByName('MyProcess', { folderPath: 'Shared/Finance' });
```

## Test plan

- [x] `npm run test:unit` — 1372 pass
- [x] `npm run build` — rollup outputs produced

**New unit tests:**
- `tests/unit/utils/folder/folder-headers.test.ts` (15) — every resolver branch (each field's routing, multi-field forwarding, fallback, error)
- `tests/unit/utils/encoding/folder-path.test.ts` (3) — verifies base64-of-UTF-16-LE output against `Buffer.from(<path>, 'utf16le').toString('base64')`
- `tests/unit/services/orchestrator/assets.test.ts` — adds 9 cases for `getByName` (success + transform validation, header routing for each folder field, OData query passthrough, quote-escape, NotFoundError, ValidationError for empty name, meta-tag fallback, fallback suppression)
- `tests/unit/services/orchestrator/processes.test.ts` — same 9 cases for processes

**New integration tests** (`tests/integration/shared/orchestrator/`):
- `getByName` with `folderKey` → live API
- `getByName` with `folderPath` → live API
- Transform validation against the live API (camelCase fields present, PascalCase originals absent)
- `NotFoundError` for a nonexistent name

Requires three env vars for integration runs:
- `INTEGRATION_TEST_FOLDER_ID` (existing)
- `INTEGRATION_TEST_FOLDER_KEY` (new — same folder, GUID)
- `INTEGRATION_TEST_FOLDER_PATH` (new — same folder, FullyQualifiedName)

The CI workflow (`pr-checks.yml`) is wired to read these from the corresponding repo secrets.

## Out of scope (deferred)

- **`processes.start()`** — extending it for `folderKey`/`folderPath` is going out as its own PR (overload pattern, deprecate the existing positional shape). Tracked separately so the v1.x positional shape isn't touched here.
- Buckets / Queues / Jobs `getByName` — the resolver + helper are reusable; those will land as separate onboardings.

### DIFF
--- a/src/core/config/runtime.ts
+++ b/src/core/config/runtime.ts
@@ -3,6 +3,13 @@ import { UiPathMetaTags } from '../../utils/runtime/constants';
 import { isBrowser } from '../../utils/platform';
 
 /**
+ * Meta-tag-derived config. Extends {@link PartialUiPathConfig} with
+ * `folderKey`, which is sourced only from `<meta name="uipath:folder-key">`
+ * (not accepted via the public SDK constructor).
+ */
+export type MetaTagConfig = PartialUiPathConfig & { folderKey?: string };
+
+/**
  * Get the content of a meta tag by name.
  * Returns undefined if not in browser environment or meta tag is not found.
  */
@@ -18,16 +25,17 @@ export function getMetaTagContent(name: string): string | undefined {
  *
  * Returns partial config with values found, or null if no meta tags present.
  */
-export function loadFromMetaTags(): PartialUiPathConfig | null {
+export function loadFromMetaTags(): MetaTagConfig | null {
   if (!isBrowser) return null;
 
-  const config: PartialUiPathConfig = {
+  const config: MetaTagConfig = {
     clientId: getMetaTagContent(UiPathMetaTags.CLIENT_ID),
     scope: getMetaTagContent(UiPathMetaTags.SCOPE),
     orgName: getMetaTagContent(UiPathMetaTags.ORG_NAME),
     tenantName: getMetaTagContent(UiPathMetaTags.TENANT_NAME),
     baseUrl: getMetaTagContent(UiPathMetaTags.BASE_URL),
     redirectUri: getMetaTagContent(UiPathMetaTags.REDIRECT_URI),
+    folderKey: getMetaTagContent(UiPathMetaTags.FOLDER_KEY),
   };
 
   const hasAnyValue = Object.values(config).some(Boolean);
--- a/src/core/internals/types.ts
+++ b/src/core/internals/types.ts
@@ -18,4 +18,11 @@ export interface PrivateSDK {
   context: ExecutionContext;
   /** Token manager for authentication */
   tokenManager: TokenManager;
+  /**
+   * Default folder key (GUID), sourced only from `<meta name="uipath:folder-key">`
+   * (injected during coded-app deployment). Used by folder-scoped services
+   * as a fallback when the caller doesn't supply folder context.
+   * Not user-settable via the SDK constructor.
+   */
+  folderKey?: string;
 }
--- a/src/models/common/types.ts
+++ b/src/models/common/types.ts
@@ -45,3 +45,18 @@ export interface RequestOptions extends BaseOptions {
   filter?: string;
   orderby?: string;
 }
+
+/**
+ * Options that scope a name-based lookup (e.g. `getByName`) to a folder.
+ * Provide one of `folderId`, `folderKey`, or `folderPath`. When more than
+ * one is supplied, all are forwarded; the server applies precedence
+ * `folderPath` > `folderKey` > `folderId`.
+ */
+export interface FolderScopedOptions extends BaseOptions {
+  /** Numeric folder ID. */
+  folderId?: number;
+  /** Folder key (GUID-formatted string). */
+  folderKey?: string;
+  /** Slash-delimited folder path, e.g. `'Shared/Finance'`. */
+  folderPath?: string;
+}
--- a/src/models/orchestrator/assets.models.ts
+++ b/src/models/orchestrator/assets.models.ts
@@ -1,4 +1,4 @@
-import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions } from './assets.types';
+import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions, AssetGetByNameOptions } from './assets.types';
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
 
 /**
@@ -66,4 +66,28 @@ export interface AssetServiceModel {
    * ```
    */
   getById(id: number, folderId: number, options?: AssetGetByIdOptions): Promise<AssetGetResponse>;
-} 
+
+  /**
+   * Retrieves a single asset by name.
+   *
+   * @param name - Asset name to search for
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) and optional query parameters (`expand`, `select`)
+   * @returns Promise resolving to a single asset
+   * {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // By folder ID
+   * await assets.getByName('ApiKey', { folderId: 123 });
+   *
+   * // By folder key (GUID)
+   * await assets.getByName('ApiKey', { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // By folder path
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });
+   *
+   * // With expand
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance', expand: 'keyValueList' });
+   * ```
+   */
+  getByName(name: string, options?: AssetGetByNameOptions): Promise<AssetGetResponse>;
+}
--- a/src/models/orchestrator/processes.models.ts
+++ b/src/models/orchestrator/processes.models.ts
@@ -1,5 +1,5 @@
 import { RequestOptions } from '../common/types';
-import { ProcessGetAllOptions, ProcessGetResponse, ProcessStartRequest, ProcessStartResponse, ProcessGetByIdOptions } from './processes.types';
+import { ProcessGetAllOptions, ProcessGetResponse, ProcessStartRequest, ProcessStartResponse, ProcessGetByIdOptions, ProcessGetByNameOptions } from './processes.types';
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
 
 /**
@@ -78,10 +78,34 @@ export interface ProcessServiceModel {
    * ```
    */
   getById(id: number, folderId: number, options?: ProcessGetByIdOptions): Promise<ProcessGetResponse>;
-  
+
+  /**
+   * Retrieves a single process by name.
+   *
+   * @param name - Process name to search for
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) and optional query parameters (`expand`, `select`)
+   * @returns Promise resolving to a single process
+   * {@link ProcessGetResponse}
+   * @example
+   * ```typescript
+   * // By folder ID
+   * await processes.getByName('MyProcess', { folderId: 123 });
+   *
+   * // By folder key (GUID)
+   * await processes.getByName('MyProcess', { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // By folder path
+   * await processes.getByName('MyProcess', { folderPath: 'Shared/Finance' });
+   *
+   * // With expand
+   * await processes.getByName('MyProcess', { folderPath: 'Shared/Finance', expand: 'entryPoints' });
+   * ```
+   */
+  getByName(name: string, options?: ProcessGetByNameOptions): Promise<ProcessGetResponse>;
+
   /**
    * Starts a process with the specified configuration
-   * 
+   *
    * @param request - Process start configuration
    * @param folderId - Required folder ID
    * @param options - Optional request options
@@ -101,4 +125,4 @@ export interface ProcessServiceModel {
    * ```
    */
   start(request: ProcessStartRequest, folderId: number, options?: RequestOptions): Promise<ProcessStartResponse[]>;
-} 
+}
--- a/src/models/orchestrator/processes.types.ts
+++ b/src/models/orchestrator/processes.types.ts
@@ -1,4 +1,4 @@
-import { JobState, RequestOptions, BaseOptions } from '../common/types';
+import { JobState, RequestOptions, BaseOptions, FolderScopedOptions } from '../common/types';
 import { PaginationOptions } from '../../utils/pagination';
 
 /**
@@ -380,3 +380,9 @@ export type ProcessGetAllOptions = RequestOptions & PaginationOptions & {
  * Options for getting a single process by ID
  */
 export interface ProcessGetByIdOptions extends BaseOptions {}
+
+/**
+ * Options for getting a single process by name
+ */
+export interface ProcessGetByNameOptions extends FolderScopedOptions {}
+
--- a/src/services/base.ts
+++ b/src/services/base.ts
@@ -40,6 +40,12 @@ export class BaseService {
   #apiClient: ApiClient;
 
   /**
+   * SDK configuration (read-only). Available to subclasses so they can
+   * fall back to init-time defaults like `folderKey`.
+   */
+  protected readonly config: { folderKey?: string };
+
+  /**
    * Creates a base service instance with dependency injection.
    *
    * Extracts configuration, execution context, and token manager from the UiPath instance
@@ -70,8 +76,9 @@ export class BaseService {
    * ```
    */
   constructor(instance: IUiPath, headers?: Record<string, string>) {
-    const { config, context, tokenManager } = SDKInternalsRegistry.get(instance);
+    const { config, context, tokenManager, folderKey } = SDKInternalsRegistry.get(instance);
     this.#apiClient = new ApiClient(config, context, tokenManager, headers ? { headers } : {});
+    this.config = { folderKey };
   }
 
   /**
--- a/src/services/folder-scoped.ts
+++ b/src/services/folder-scoped.ts
@@ -1,9 +1,18 @@
 import { BaseService } from './base';
-import { CollectionResponse } from '../models/common/types';
+import { CollectionResponse, FolderScopedOptions } from '../models/common/types';
 import { createHeaders } from '../utils/http/headers';
 import { FOLDER_ID } from '../utils/constants/headers';
 import { ODATA_PREFIX } from '../utils/constants/common';
 import { addPrefixToKeys } from '../utils/transform';
+import { NotFoundError } from '../core/errors';
+import { validateName } from '../utils/validation/name-validator';
+import { resolveFolderHeaders } from '../utils/folder/folder-headers';
+
+/**
+ * Matches single-quote characters in OData string literals — escaped to `''`
+ * inside the `$filter=Name eq '…'` clause built by `getByNameLookup`.
+ */
+const SINGLE_QUOTE_RE = /'/g;
 
 /**
  * Base service for services that need folder-specific functionality.
@@ -18,7 +27,7 @@ import { addPrefixToKeys } from '../utils/transform';
 export class FolderScopedService extends BaseService {
   /**
    * Gets resources in a folder with optional query parameters
-   * 
+   *
    * @param endpoint - API endpoint to call
    * @param folderId - required folder ID
    * @param options - Query options
@@ -26,8 +35,8 @@ export class FolderScopedService extends BaseService {
    * @returns Promise resolving to an array of resources
    */
   protected async _getByFolder<T, R = T>(
-    endpoint: string, 
-    folderId: number, 
+    endpoint: string,
+    folderId: number,
     options: Record<string, any> = {},
     transformFn?: (item: T) => R
   ): Promise<R[]> {
@@ -35,10 +44,10 @@ export class FolderScopedService extends BaseService {
 
     const keysToPrefix = Object.keys(options);
     const apiOptions = addPrefixToKeys(options, ODATA_PREFIX, keysToPrefix);
-    
+
     const response = await this.get<CollectionResponse<T>>(
       endpoint,
-      { 
+      {
         params: apiOptions,
         headers
       }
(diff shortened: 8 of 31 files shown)


src/core/config/runtime.ts
```diff
@@ -2,6 +2,13 @@ import { PartialUiPathConfig } from './sdk-config';
 import { UiPathMetaTags } from '../../utils/runtime/constants';
 import { isBrowser } from '../../utils/platform';
 
+/**
+ * Meta-tag-derived config. Extends {@link PartialUiPathConfig} with
+ * `folderKey`, which is sourced only from `<meta name="uipath:folder-key">`
+ * (not accepted via the public SDK constructor).
+ */
+export type MetaTagConfig = PartialUiPathConfig & { folderKey?: string };
```
swati354, line 10: why not allow passing the folder-key as well via constructor? Every other meta tag has a constructor field.

src/models/common/types.ts
```diff
@@ -44,4 +44,23 @@ export interface BaseOptions {
...
+export interface FolderScopedOptions extends BaseOptions {
+  /**
+   * Folder path (e.g. 'Shared/Finance'). Sent as the
+   * `X-UIPATH-FolderPath-Encoded` header (encoded automatically as
+   * base64-of-UTF-16-LE per the Orchestrator contract).
+   */
+  folderPath?: string;
+  /**
+   * Folder key (GUID). Sent as the `X-UIPATH-FolderKey` header.
+   */
```
swati354: I don't think we need to surface the detail of the header or encoding here. (not in the docs)?
What should be surfaced is:

- That folderPath and folderKey scope the lookup to a folder.
- Format hint: folderPath is a slash-delimited path ('Shared/Finance'), folderKey is a GUID.
- Precedence when both are supplied.

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,31 @@ export interface AssetServiceModel {
    * ```
    */
   getById(id: number, folderId: number, options?: AssetGetByIdOptions): Promise<AssetGetResponse>;
-} 
\ No newline at end of file
+
+  /**
+   * Retrieves a single asset by name, optionally scoped to a folder
```
swati354: > optionally scoped to a folder

Why say optionally scoped? Named entities are always scoped to a folder?

src/services/orchestrator/assets/assets.ts
```diff
@@ -115,7 +115,50 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+   * // Get asset by name (uses default folder context)
+   * const asset = await assets.getByName('ApiKey');
+   * ```
+   */
+  @track('Assets.GetByName')
+  async getByName(name: string, options: AssetGetByNameOptions = {}): Promise<AssetGetResponse> {
+    const { folderPath, folderKey, ...queryOptions } = options;
+    return this.getResourceByName<AssetGetResponse>({
+      resourceType: 'Asset',
+      endpoint: ASSET_ENDPOINTS.GET_ALL,
```
swati354: why are we not using the folder scoped endpoint here?
https://github.com/UiPath/uipath-typescript/blob/db340aab1ee57fef8fe8510c533d7f109fdcefb5/src/utils/constants/endpoints/orchestrator.ts#L71

src/services/base.ts
```diff
@@ -147,6 +161,83 @@ export class BaseService {
...
+    /** Resource label used in validation + error messages (e.g. 'Asset', 'Process') */
+    resourceType: string;
+    /** Folder-scoped OData collection endpoint */
+    endpoint: string;
+    /** Resource name to search for */
+    name: unknown;
+    /** Optional folder path (sent as X-UIPATH-FolderPath-Encoded) */
+    folderPath?: unknown;
+    /** Optional folder key (sent as X-UIPATH-FolderKey) */
+    folderKey?: unknown;
```
swati354: shouldn't these be strings?

src/services/base.ts
```diff
@@ -147,6 +161,83 @@ export class BaseService {
...
+    const headers = createHeaders({
+      [FOLDER_PATH_ENCODED]: validated.folderPath ? encodeFolderPathHeader(validated.folderPath) : undefined,
+      [FOLDER_KEY]: effectiveFolderKey,
+    });
+
+    const queryOptions = params.queryOptions ?? {};
+    const apiOptions = {
+      ...addPrefixToKeys(queryOptions, ODATA_PREFIX, Object.keys(queryOptions)),
+      '$filter': `Name eq '${validated.name.replace(/'/g, "''")}'`,
+      '$top': '1',
```
swati354: this method only works for odata endpoints?

src/services/base.ts
```diff
@@ -147,6 +161,83 @@ export class BaseService {
...
+   * - `X-UIPATH-FolderPath-Encoded` / `X-UIPATH-FolderKey` header construction
+   * - Empty-result → `NotFoundError` with folder context in the message
+   *
+   * The transform step is caller-provided because each resource has its own
+   * PascalCase → camelCase field mapping.
+   *
+   * @param params - Resource-specific parameters
+   * @throws ValidationError when inputs are malformed; NotFoundError when no match
+   */
+  protected async getResourceByName<T>(params: {
```
swati354: can this method be put in folderScopedService - https://github.com/UiPath/uipath-typescript/blob/main/src/services/folder-scoped.ts, this service isn't being used right now, but seems like a better fit for this purpose?

swati354: can this method be put in folderScopedService - https://github.com/UiPath/uipath-typescript/blob/main/src/services/folder-scoped.ts, this service isn't being used right now, but seems like a better fit for this purpose?

swati354: Could we have the method as 
```
protected async getByNameLookup<TRaw, T>(
  resourceType: string,
  endpoint: string,
  name: string,
  options: FolderScopedOptions,
  transform: (raw: TRaw) => T,
): Promise<T> {
  ......
```
`<TRaw, T>` also kills the cast inside each caller's transform lambda

src/utils/validation/name-validator.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param resourceType - Resource label used in error messages (e.g. 'Asset',
+ *                       'Process'). Keeps errors contextual for the caller.
+ * @throws ValidationError when any argument fails validation
+ * @returns Trimmed, type-checked copies of the inputs ready to be passed to HTTP layers
+ */
+export function validateGetByNameArgs(
+  resourceType: string,
+  name: unknown,
+  folderPath?: unknown,
+  folderKey?: unknown,
```
swati354: type these to string

swati354: The two methods seem redundant.  They do the same thing.

src/core/config/runtime.ts
deepeshrai-tech: we agreed right in our last discussion to not pass it in the constructor ? @Raina451

src/models/common/types.ts
deepeshrai-tech: fixed

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: fixed

src/services/base.ts
deepeshrai-tech: yes

src/services/base.ts
deepeshrai-tech: it is used in assets,buckets,queues and jobs but yea keeping it in folderScopedService makes sense

src/core/config/runtime.ts
Raina451: let's pass in constructor also, 
priority order:
1. folderKey passed in method
2. constructor
3. meta tag

but it's only folderKey and not folderPath i believe. but lets discuss once

src/models/common/types.ts
```diff
@@ -44,4 +44,22 @@ export interface BaseOptions {
...
+
+/**
+ * Options that scope a name-based lookup (e.g. `getByName`) to a folder.
+ * Provide `folderPath` or `folderKey`. When both are supplied, `folderPath`
+ * wins.
+ */
+export interface FolderScopedOptions extends BaseOptions {
+  /**
+   * Slash-delimited folder path that scopes the lookup, e.g. `'Shared/Finance'`.
+   * When both `folderPath` and `folderKey` are supplied, `folderPath` wins.
```
ninja-shreyash: The same precedence is mentioned at 3 places. We can keep it at one place within the same JSDoc

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,26 @@ export interface AssetServiceModel {
    * ```
    */
   getById(id: number, folderId: number, options?: AssetGetByIdOptions): Promise<AssetGetResponse>;
-} 
\ No newline at end of file
+
+  /**
+   * Retrieves a single asset by name, scoped to a folder via `folderPath` or
```
ninja-shreyash: Do we need to mention "scoped to a folder" here? We don't mention "scoped to folder" for other getById methods - even within this service assets itself even though it takes folderId as required parameter

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,26 @@ export interface AssetServiceModel {
...
+   * @example
+   * ```typescript
+   * // Get asset by name with folder path
+   * const asset = await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });
+   *
+   * // Get asset by name with folder key
+   * const asset = await assets.getByName('ApiKey', { folderKey: 'folder-guid' });
+   *
+   * // Get asset by name (uses default folder context)
+   * const asset = await assets.getByName('ApiKey');
```
ninja-shreyash: this is unlike the other getById methods where folderId is a required param. what is going to happen if no folder is identified from the meta tags/constructor as well?

src/models/orchestrator/processes.models.ts
```diff
@@ -78,27 +78,83 @@ export interface ProcessServiceModel {
...
+   * const process = await processes.getByName('MyProcess', { folderKey: 'folder-guid' });
+   *
+   * // Get process by name (uses default folder context)
+   * const process = await processes.getByName('MyProcess');
+   * ```
+   */
+  getByName(name: string, options?: ProcessGetByNameOptions): Promise<ProcessGetResponse>;
+
+  /**
+   * Starts a process execution (job).
```
ninja-shreyash: There would be many more methods taking folderId as a required param across services.. are we going to overload all those as well?

src/models/orchestrator/assets.models.ts
ninja-shreyash: I think

`@param options - Optional folder scoping and query parameters` should be enough

src/models/orchestrator/processes.types.ts
```diff
@@ -379,4 +379,23 @@ export type ProcessGetAllOptions = RequestOptions & PaginationOptions & {
...
+export interface ProcessGetByIdOptions extends BaseOptions {}
+
+/**
+ * Options for getting a single process by name
+ */
+export interface ProcessGetByNameOptions extends FolderScopedOptions {}
+
+/**
+ * Options for `processes.start()`. Supplies the folder context plus any
+ * OData query options. At least one folder field (`folderId`, `folderPath`,
```
ninja-shreyash: who checks that one of the 3 must be present?

ninja-shreyash: in the actual prcoesses.start implemetation we fall back instead of throwing error if no folder context provided

src/services/orchestrator/processes/processes.ts
```diff
@@ -98,60 +101,108 @@ export class ProcessService extends BaseService implements ProcessServiceModel {
...
+        : (optionsOrFolderId ?? {});
+
+    const { folderId, folderPath, folderKey, ...queryOptions } = opts;
+
+    // Fall back to init-time folderKey (e.g. uipath:folder-key meta tag) only
+    // when the caller supplied no folder context at all.
+    const effectiveFolderKey =
+      folderKey ?? (folderPath || folderId !== undefined ? undefined : this.config.folderKey);
+
+    const headers = createHeaders({
```
ninja-shreyash: We are sending both as headers, but our doc says that folderPath wins over folderKey

Is it known that all folder scoped services will respect this heirarchy? even for new services that we are going to add in future

ninja-shreyash: Should we not handle this here itself by suppressing the folderKey in favour of folderPath

src/core/config/config.ts
swati354, line 1: why are changes required in this file, and even in `runtime.ts` if we don't want to pass the folderKey via constructor? 
`UiPathConfig` mirrors what the user would pass. We are deciding to keep `folderKey` only as runtime config.
Is there another way?
Can we consider this - passing the folderKey via `PrivateSDK`. 
In `uipath.ts` then we can populate the `SDKInternalsRegistry`, buy picking the `folderKey` directly from metatag and in `base.ts` we take the folderKey from the `SDKInternalsRegistry`

src/models/common/types.ts
swati354: But all of them will show at different places? 
We need it on top of methods and if somebody searches for `FolderScopedOptions`, the description should be there as well?

swati354:
To allow the users to locally run this, we would need to allow passing folderKey via `uipath.json`, which would require changes in docs(https://uipath.github.io/uipath-typescript/coded-apps/getting-started/?h=uipath+json#configure-uipathjson) + the coded-apps plugin.

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,26 @@ export interface AssetServiceModel {
...
+   * {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // Get asset by name with folder path
+   * const asset = await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });
+   *
+   * // Get asset by name with folder key
+   * const asset = await assets.getByName('ApiKey', { folderKey: 'folder-guid' });
+   *
+   * // Get asset by name (uses default folder context)
```
swati354: > uses default folder context

How do the users know what a default folder means? Can we be more explicit?

src/services/folder-scoped.ts
```diff
@@ -47,7 +56,74 @@ export class FolderScopedService extends BaseService {
...
+  ): Promise<T> {
+    const { folderPath, folderKey, ...queryOptions } = options;
+    const validated = validateGetByNameArgs(resourceType, name, folderPath, folderKey);
+
+    // Fall back to the SDK's init-time folderKey (e.g. populated from the
+    // `uipath:folder-key` meta tag in coded-app deployments) when the caller
+    // didn't supply any folder context.
+    const effectiveFolderKey =
+      validated.folderKey ?? (validated.folderPath ? undefined : this.config.folderKey);
+
```
swati354: why don't we throw a validation error here if `validated.folderPath` `effectiveFolderKey` both don't exist ?
This way all methods irrespective of how they deal with folders would throw consistent errors. 
Otherwise currently, assets would throw an error from the server since it uses the folder endpoint, but process would just return the first match (even if that was not the process which the user wanted)

src/core/config/runtime.ts
deepeshrai-tech: it was agrred not to pass via constructor

src/models/common/types.ts
deepeshrai-tech: yea its fine to have info present where its needed in JS docs

src/models/orchestrator/assets.models.ts
deepeshrai-tech: there's always gonna be one from meta tag

src/models/orchestrator/processes.models.ts
deepeshrai-tech: yes we will be modifying all the reosurces one by one with folderPath/key support

src/services/orchestrator/processes/processes.ts
deepeshrai-tech: just for Odata endpoints rest will be explored when i will modify those services, and since its handled already in orch level doenst make sense to have a logic at our side too. U can see python_sdk for reference

src/models/orchestrator/processes.types.ts
deepeshrai-tech: its present at orch level . we dont need to check anything as meta tag will always be there, we just pass in headers and orch will handle rest. If we get error from there side we just show that. Will add precedence in docs

src/models/orchestrator/assets.models.ts
deepeshrai-tech: actually examples with just asset name are not even meant to be here, i think its best to just remove these

src/core/config/config.ts
deepeshrai-tech: agree,missed to change this when i removed folderKey from constructor

swati354:
The PR description says `utils/validation/name-validator.ts throws ValidationError for non-string names`, but it doesn't? It also mentions `UiPath({ folderKey })`, which we don't support. Is the description stale?
Could you update it please.

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,28 @@ export interface AssetServiceModel {
...
+   * @param name - Asset name to search for
+   * @param options - Folder scoping (`folderId` / `folderPath`) + optional OData query parameters
+   * @returns Promise resolving to a single asset
+   * {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // By folder ID
+   * await assets.getByName('ApiKey', { folderId: 123 });
+   *
+   * // By folder key (GUID — auto-detected via the same `folderId` field)
```
ninja-shreyash:
```suggestion
   * // By folder key (passed within folderId field)
```

src/models/orchestrator/processes.models.ts
```diff
@@ -78,10 +78,34 @@ export interface ProcessServiceModel {
...
+   * @param name - Process name to search for
+   * @param options - Folder scoping (`folderId` / `folderPath`) + optional OData query parameters
+   * @returns Promise resolving to a single process
+   * {@link ProcessGetResponse}
+   * @example
+   * ```typescript
+   * // By folder ID
+   * await processes.getByName('MyProcess', { folderId: 123 });
+   *
+   * // By folder key (GUID — auto-detected via the same `folderId` field)
```
ninja-shreyash: same as above

src/utils/folder/folder-headers.ts
```diff
@@ -0,0 +1,96 @@
...
+    } else {
+      throw new ValidationError({
+        message: `${resourceType} folderId must be a number or a GUID-formatted string. Use the \`folderPath\` option to scope by folder path.`,
+      });
+    }
+  }
+
+  // No explicit folder context → meta-tag fallback or error.
+  if (Object.keys(headers).length === 0) {
+    if (!fallbackFolderKey) {
```
ninja-shreyash, line 73: Should we check fallbackFolderKey for GUID format as well?

src/core/internals/types.ts
```diff
@@ -18,4 +18,11 @@ export interface PrivateSDK {
   context: ExecutionContext;
   /** Token manager for authentication */
   tokenManager: TokenManager;
+  /**
+   * Default folder key (GUID), sourced only from `<meta name="uipath:folder-key">`
+   * (injected by jamjam in coded-app deployments). Used by folder-scoped
```
swati354: don't mention `jamjam`, just say injected during deployment.

src/models/common/types.ts
```diff
@@ -44,4 +44,18 @@ export interface BaseOptions {
...
   filter?: string;
   orderby?: string;
-}
\ No newline at end of file
+}
+
+/**
+ * Options that scope a name-based lookup (e.g. `getByName`) to a folder.
+ * Provide one of `folderId`, `folderKey`, or `folderPath`. When more than
+ * one is supplied, all are forwarded and the server resolves precedence.
```
swati354: can we mention that precedence as well.

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,28 @@ export interface AssetServiceModel {
...
    */
   getById(id: number, folderId: number, options?: AssetGetByIdOptions): Promise<AssetGetResponse>;
-} 
\ No newline at end of file
+
+  /**
+   * Retrieves a single asset by name.
+   *
+   * @param name - Asset name to search for
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) + optional OData query parameters
```
swati354: don't mention `OData`.
In the example as well. Users don't need to worry about `OData`.
Same for the other `getByName` method

src/models/orchestrator/assets.models.ts
```diff
@@ -66,4 +66,28 @@ export interface AssetServiceModel {
...
+   * await assets.getByName('ApiKey', { folderId: 123 });
+   *
+   * // By folder key (GUID)
+   * await assets.getByName('ApiKey', { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // By folder path
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });
+   *
+   * // With OData query parameters
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance', expand: 'KeyValueList' });
```
swati354: `expand : keyValueList`

src/models/orchestrator/processes.models.ts
```diff
@@ -78,10 +78,34 @@ export interface ProcessServiceModel {
...
+   * await processes.getByName('MyProcess', { folderId: 123 });
+   *
+   * // By folder key (GUID)
+   * await processes.getByName('MyProcess', { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // By folder path
+   * await processes.getByName('MyProcess', { folderPath: 'Shared/Finance' });
+   *
+   * // With OData query parameters
+   * await processes.getByName('MyProcess', { folderPath: 'Shared/Finance', expand: 'Robot' });
```
swati354: Is there a `Robot` field ?

src/services/orchestrator/assets/assets.ts
```diff
@@ -115,7 +115,44 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+   *
+   * // By folder path
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance' });
+   *
+   * // With OData query parameters
+   * await assets.getByName('ApiKey', { folderPath: 'Shared/Finance', expand: 'KeyValueList' });
+   * ```
+   */
+  @track('Assets.GetByName')
+  async getByName(name: string, options: AssetGetByNameOptions = {}): Promise<AssetGetResponse> {
```
swati354, line 149: can we throw validation error if name is not provided like we do for IDs

src/utils/folder/folder-headers.ts
deepeshrai-tech: minimal chance for faulty guid and even if there is we provide them folder not found message

swati354 (approved):

ninja-shreyash (approved):

PR #388, opened 2026-04-23
feat(document-understanding): add document-processing contracts [RE-11913]

anatolie-darii:
## Summary
Add the typescript contracts corresponding to `UiPath.DocumentProcessing.Contracts` C# contracts


ioanajoj (approved):

src/models/document-understanding/document-processing/dom.interface.ts
```diff
@@ -0,0 +1,148 @@
...
+    Dom?: Document;
+    Text: string;
+}
+
+export interface Polygon {
+    Points?: Array<{ X: number; Y: number }>;
+}
+
+export interface PolygonConverter extends JsonConverter {
+}
```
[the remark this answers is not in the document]
anatolie-darii: fixed

sergiubologa (approved):

maninder-uipath (approved):

src/models/document-understanding/document-processing/actions.interface.ts
```diff
@@ -0,0 +1,75 @@
...
+    AutomaticExtractionResultsPath: string;
+    ValidatedExtractionResultsPath: string;
+    ExtractorPayloadsPath: string;
+    ShowOnlyRelevantPageRange: boolean;
+    AdditionalDataPath: string;
+    OriginalDocumentFileName: string;
+    CustomizationInfoPath: string;
+}
+
+export interface DuDocumentActionData {
```
IoanaGligan: Can we limit the contract to only the cvd? The action is NOT for du to define

IoanaGligan (dismissed):

src/models/document-understanding/document-processing/classification.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+    DocumentTypeId: string;
+    Name: string;
+}
+
+export interface DuClassifierDocumentTypeCapability {
+    DocumentType: string;
+}
+
+export interface DuClassifierResult {
+    Classifications?: DuClassificationResult[];
```
IoanaGligan: This is unnecessary? We only deal with extraction for now

src/models/document-understanding/document-processing/data-extraction.interface.ts
```diff
@@ -0,0 +1,39 @@
...
+} from './results.interface';
+import type { DuField } from './taxonomy.interface';
+
+export interface DuExtractorDocumentType {
+    DocumentTypeId: string;
+    Fields?: DuField[];
+    Metadata?: DuExtractorMetadata[];
+}
+
+export interface DuExtractorDocumentTypeCapabilities {
```
IoanaGligan: What is this used for? Should not be used in vs

src/models/document-understanding/document-processing/data-extraction.interface.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export interface DuExtractorMetadata {
+    Key: string;
+    Value: string;
+}
+
+export interface DuExtractorResult {
+    DataPoints?: DuResultsDataPoint[];
+    Tables?: DuResultsTable[];
```
IoanaGligan: We want to deprecate the Tables, please remove

src/models/document-understanding/document-processing/results.interface.ts
```diff
@@ -0,0 +1,205 @@
...
+    DocumentTypeId: string;
+    DocumentId: string;
+    Confidence: number;
+    OcrConfidence: number;
+    Reference?: DuResultsContentReference;
+    DocumentBounds?: DuResultsDocumentBounds;
+    ClassifierName: string;
+}
+
+export interface DuDocumentSplittingResult {
```
IoanaGligan: Same comment as above - splitting should not be in

src/models/document-understanding/index.ts
```diff
@@ -0,0 +1 @@
+export * as DuFramework from './framework';
```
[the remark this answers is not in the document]
anatolie-darii: It creates naming collisions otherwise

src/models/document-understanding/index.ts
Raina451, line 1: below things are needed for consumers to actually consume these types:

1. Add `export * from './models/document-understanding'` to https://github.com/UiPath/uipath-typescript/blob/main/src/index.ts#L12-L19
2. dedicated subpath would be needed to consume DU related things (e.g., `@uipath/uipath-typescript/document-understanding`):
      - Add a service entry in [rollup.config.js](https://github.com/UiPath/uipath-typescript/blob/main/rollup.config.js#L136)
      - Add a `./document-understanding` export entry in [package.json exports](https://github.com/UiPath/uipath-typescript/blob/main/package.json#L29-L49)

Without these, the types ship in source but are invisible to consuming the package.  Please decide on the export path name and add accordingly

anatolie-darii: point 1 can't be done as it creates naming collisions between multiple folders. it's also a bad design to do so. we should either prefix all models or export them namespaced as I did for framework.

Will follow up on point 2.

anatolie-darii: addressed point 1 as well because it actually exports the namespace DuFramework as opposed to direct entities that would have created the naming collisions

src/models/document-understanding/index.ts
swati354, line 1: @anatolie-darii Is the namespace name based only on the folder structure here? If yes, could we give a different name. `Du` or `DocumentUnderstanding`? I think the shorthand works fine.
cc @Raina451  @vnaren23

Raina451 (approved):

src/models/document-understanding/index.ts
anatolie-darii: It's based on the API contracts it exposes, we could name it DU but if we add contracts for other DU APIs we might end up with naming collisions, this is why it's named specifically DuFramework

swati354 (approved):

PR #398, opened 2026-04-29
fix: handle HTML error responses for invalid org/tenant in auth-errors integration test

Sarath1018:
The isForbiddenError helper in auth-errors.integration.test.ts was checking for error instanceof SyntaxError to detect HTML error page responses from the API, but GitHub Actions runners receive a NetworkError (SDK's own class) instead of a native SyntaxError for the same scenario
Added error instanceof NetworkError check alongside the existing SyntaxError check so both local and CI environments are handled correctly

### DIFF
--- a/tests/integration/auth-errors.integration.test.ts
+++ b/tests/integration/auth-errors.integration.test.ts
@@ -59,6 +59,16 @@ async function expectApiToFail(
     await apiCall();
     expect.fail(`Expected API call to fail: ${testDescription}`);
   } catch (error: any) {
+    console.log(`[DEBUG] ${testDescription} — error shape:`, JSON.stringify({
+      constructorName: error?.constructor?.name,
+      type: error?.type,
+      message: error?.message,
+      statusCode: error?.statusCode,
+      status: error?.status,
+      requestId: error?.requestId,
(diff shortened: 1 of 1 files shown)


swati354:
`JSON.parse()` failed because the body was HTML. It is misleading to call it a network error. The error seems to come from here - https://github.com/UiPath/uipath-typescript/blob/45587cac220ca3e28dada2943b54365a5a79073a/src/core/http/api-client.ts#L121
`JSON.parse` shares a catch with the network-error wrapper. 
This needs fixing at the api-client leveI I believe.

Sarath1018:
> `JSON.parse()` failed because the body was HTML. It is misleading to call it a network error. The error seems to come from here -
> 
> https://github.com/UiPath/uipath-typescript/blob/45587cac220ca3e28dada2943b54365a5a79073a/src/core/http/api-client.ts#L121
> 
> 
> `JSON.parse` shares a catch with the network-error wrapper.
> This needs fixing at the api-client leveI I believe.

 created a ticket to do this https://uipath.atlassian.net/browse/PLT-102926 Lets merge this to unblock PR gates

swati354 (approved):

deepeshrai-tech (approved):

PR #399, opened 2026-04-29
feat(datafabric): Adding support for advanced field option[DS-8268]

AditiGoyalUipath:
Fixes a class of "Field type cannot be changed" errors caused by incomplete
  sqlType payloads on entity field creation.                                       
                                          
  Changes:                                                                         
  - buildSqlTypeConstraints now returns API-required defaults per field type:      
    - STRING → lengthLimit: 200                                              
    - MULTILINE_TEXT → lengthLimit: 200                                          
    - DECIMAL → lengthLimit: 1000, decimalPrecision: 2, maxValue/minValue: ±1e12
    - INTEGER, BIG_INTEGER, FLOAT, DOUBLE → maxValue/minValue: ±1e12               
    - BOOLEAN → lengthLimit: 100 (fixed)                            
    - DATE, DATETIME_WITH_TZ → lengthLimit: 1000 (fixed)                           
    - All defaults are user-overridable via EntityCreateFieldOptions

### DIFF
--- a/src/models/data-fabric/entities.constants.ts
+++ b/src/models/data-fabric/entities.constants.ts
@@ -1,5 +1,10 @@
 import { EntityFieldDataType, FieldDisplayType } from "./entities.types";
-import { EntitySchemaFieldMapping, SqlFieldType } from "./entities.internal-types";
+import {
+  EntitySchemaFieldMapping,
+  SqlFieldType,
+  EntityFieldConstraint,
+  FieldConstraintRange,
+} from "./entities.internal-types";
 export { SqlFieldType } from "./entities.internal-types";
 
 /**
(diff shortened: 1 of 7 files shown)


src/services/data-fabric/entities.ts
```diff
@@ -984,6 +997,47 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+        return { lengthLimit: 1000 };
+      case EntityFieldDataType.INTEGER:
+      case EntityFieldDataType.BIG_INTEGER:
+      case EntityFieldDataType.FLOAT:
+      case EntityFieldDataType.DOUBLE:
+        return {
+          maxValue: field.maxValue ?? 1000000000000,
+          minValue: field.minValue ?? -1000000000000,
+        };
+      default:
```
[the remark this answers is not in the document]
swati354: @AditiGoyalUipath could you check this ^

src/services/data-fabric/entities.ts
```diff
@@ -984,6 +997,47 @@ export class EntityService extends BaseService implements EntityServiceModel {
     };
   }
 
+  /**
+   * Returns the sqlType constraint fields for a given field type.
+   */
+  private buildSqlTypeConstraints(type: EntityFieldDataType, field: EntityCreateFieldOptions): Omit<SqlType, 'name'> {
+    switch (type) {
+      case EntityFieldDataType.STRING:
+        // NVARCHAR requires an explicit lengthLimit — without it the API stores the field
```
swati354: stores the field as?

src/services/data-fabric/entities.ts
```diff
@@ -984,6 +997,47 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+          decimalPrecision: field.decimalPrecision ?? 2,
+          maxValue: field.maxValue ?? 1000000000000,
+          minValue: field.minValue ?? -1000000000000,
+        };
+      case EntityFieldDataType.BOOLEAN:
+        return { lengthLimit: 100 };
+      case EntityFieldDataType.DATE:
+        return { lengthLimit: 1000 };
+      case EntityFieldDataType.DATETIME_WITH_TZ:
+        return { lengthLimit: 1000 };
```
swati354: Instead of allowing the users to pass the `lengthLimit` for these fields and then silently dropping them, could we throw a validation error if user passes this field for a type that doesn't accept it?

src/services/data-fabric/entities.ts
```diff
@@ -984,6 +997,47 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+        // NVARCHAR requires an explicit lengthLimit — without it the API stores the field
+        return { lengthLimit: field.lengthLimit ?? 200 };
+      case EntityFieldDataType.MULTILINE_TEXT:
+        return { lengthLimit: field.lengthLimit ?? 10000 };
+      case EntityFieldDataType.DECIMAL:
+        return {
+          lengthLimit: field.lengthLimit ?? 1000,
+          decimalPrecision: field.decimalPrecision ?? 2,
+          maxValue: field.maxValue ?? 1000000000000,
+          minValue: field.minValue ?? -1000000000000,
```
swati354: create constant for all the default length limits in `entities.constants.ts`

src/models/data-fabric/entities.types.ts
```diff
@@ -241,6 +241,14 @@ export interface EntityFieldBase {
...
   /** Default value for the field */
   defaultValue?: string;
+  /** Maximum character length for STRING fields (default: 200, max: 4000) and MULTILINE_TEXT fields (default: 10000). For DECIMAL fields, defaults to 1000 if not provided. */
+  lengthLimit?: number;
+  /** Maximum allowed value for numeric fields (INTEGER, BIG_INTEGER, FLOAT, DOUBLE, DECIMAL — default: 1,000,000,000,000) */
+  maxValue?: number;
+  /** Minimum allowed value for numeric fields (INTEGER, BIG_INTEGER, FLOAT, DOUBLE, DECIMAL — default: -1,000,000,000,000) */
+  minValue?: number;
+  /** Number of decimal places for DECIMAL fields (e.g., `2` for currency values) */
+  decimalPrecision?: number;
```
swati354: can we add one example of using these fields in the `@example` block of the methods?

src/services/data-fabric/entities.ts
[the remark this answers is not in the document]
swati354: Question - I think we have handled `CHOICE_SET_MULTIPLE`, what about the rest 3? does the api accept them with no sql type defaults?

src/services/data-fabric/entities.ts
```diff
@@ -984,6 +1027,110 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+    const sqlMapped = sqlName ? EntityFieldTypeMap[sqlName] : undefined;
+    return sqlMapped ?? EntityFieldDataType.STRING;
+  }
+
+  /**
+   * Validates that the user-supplied constraint properties on a field are
+   * supported by the field's data type. Throws a `ValidationError` listing
+   * any unsupported properties.
+   */
+  private validateFieldConstraints(
```
swati354, line 1062: nit : `minValue > maxValue` case is not caught?

src/services/data-fabric/entities.ts
```diff
@@ -1011,6 +1027,110 @@ export class EntityService extends BaseService implements EntityServiceModel {
     };
   }
 
+  /**
+   * Derives the user-facing {@link EntityFieldDataType} for a field on the raw
+   * API response. Prefers `fieldDisplayType` to disambiguate types that share
+   * a SQL type (FILE, CHOICE_SET_*, AUTO_NUMBER, RELATIONSHIP), and falls back
+   * to mapping the raw SQL type name. Returns `STRING` if neither is resolvable.
+   */
+  private resolveFieldDataType(f: FieldMetaData): EntityFieldDataType {
```
swati354, line 1047: could you verify if this duplicates the type-resolution logic already in `mapFieldTypes`. Both do `fieldDisplayType -> FieldDisplayTypeToDataType`, then fall back to `sqlType.name -> EntityFieldTypeMap`. 
they diverge in case of a no-match though. Would that cause issues later?

tests/unit/services/data-fabric/entities.test.ts
```diff
@@ -1919,6 +1888,376 @@ describe("EntityService Unit Tests", () => {
       expect(secretField?.isEncrypted).toBe(true);
     });
 
+    describe("sqlType constraint defaults", () => {
```
swati354, line 2147: all tests here are on `updateById`, could we have a few on `create` as well?

src/models/data-fabric/entities.constants.ts
```diff
@@ -48,6 +48,94 @@ export const FieldDisplayTypeToDataType: Partial<Record<FieldDisplayType, Entity
...
 
+/**
+ * Default and fixed sqlType constraint values applied when the user does not provide them.
+ * The API requires these to be present on field creation — without them the field
+ * is stored in an incomplete state, causing "Field type cannot be changed" errors
+ * when the UI later tries to edit advanced options.
+ */
+export const ENTITY_FIELD_CONSTRAINT_DEFAULTS = {
+  STRING_LENGTH_LIMIT: 200,
+  MULTILINE_TEXT_LENGTH_LIMIT: 200,
```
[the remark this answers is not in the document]
swati354: @AditiGoyalUipath ^
Please update the PR description with other changes as well if required.

src/services/data-fabric/entities.ts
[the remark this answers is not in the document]
AditiGoyalUipath: yes. Mentioned in comment as well

src/services/data-fabric/entities.ts
AditiGoyalUipath: API handles this check. But we can include in sdk as well

src/services/data-fabric/entities.ts
AditiGoyalUipath: 1 is for get and other for update. In update record we need to validate the request type. In get response We need not validate

swati354 (approved):

src/models/data-fabric/entities.models.ts
```diff
@@ -543,9 +522,11 @@ export interface EntityServiceModel {
    *
    * const entities = new Entities(sdk);
    *
+   * // Includes advanced sqlType constraints (lengthLimit, decimalPrecision, maxValue, minValue)
    * const id = await entities.create("product_catalog", [
-   *   { fieldName: "product_name", type: EntityFieldDataType.STRING, isRequired: true, isUnique: true },
-   *   { fieldName: "price", type: EntityFieldDataType.INTEGER, defaultValue: "0" },
+   *   { fieldName: "product_name", type: EntityFieldDataType.STRING, isRequired: true, isUnique: true, lengthLimit: 500 },
+   *   { fieldName: "price", type: EntityFieldDataType.DECIMAL, decimalPrecision: 4, maxValue: 999999, minValue: 0 },
+   *   { fieldName: "quantity", type: EntityFieldDataType.INTEGER, maxValue: 10000, minValue: 1 },
```
vnaren23: nit: `defaultValue` went away from examples, lets keep it. 
```
{ fieldName: "quantity", type: EntityFieldDataType.INTEGER, maxValue: 10000, minValue: 1, defaultValue: "0" }
```

src/services/data-fabric/entities.ts
```diff
@@ -737,9 +719,11 @@ export class EntityService extends BaseService implements EntityServiceModel {
    *
    * @example
    * ```typescript
+   * // Includes advanced sqlType constraints (lengthLimit, decimalPrecision, maxValue, minValue)
    * const entityId = await entities.create("product_catalog", [
-   *   { fieldName: "product_name", type: EntityFieldDataType.STRING, isRequired: true, isUnique: true },
-   *   { fieldName: "price", type: EntityFieldDataType.INTEGER, defaultValue: "0" },
+   *   { fieldName: "product_name", type: EntityFieldDataType.STRING, isRequired: true, isUnique: true, lengthLimit: 500 },
+   *   { fieldName: "price", type: EntityFieldDataType.DECIMAL, decimalPrecision: 4, maxValue: 999999, minValue: 0 },
+   *   { fieldName: "quantity", type: EntityFieldDataType.INTEGER, maxValue: 10000, minValue: 1 },
```
vnaren23: Same here, bring back `defaultValue`

vnaren23 (commented):
Minor doc updates, rest LGTM

vnaren23 (approved):

PR #402, opened 2026-04-30
feat(cas): add support for new tool confirmation

scottcmg:
## Description

Conversational agents revamped how tool confirmation works for unified runtime agents. The legacy SDK gained support for the new flow; this PR ports those changes into the new SDK so confirmation works end-to-end with the chat widget UI.

## Background

Previously, tool-call confirmation was modeled as an *interrupt* (`sendInterruptEnd` on the message). The runtime now models it as a first-class event on the tool call itself: a `requireConfirmation` flag on `startToolCall`, paired with a `confirmToolCall` response carrying the user's decision (and optionally edited input).

## Protocol changes (`protocol.types.ts`)

- `ToolCallStartEvent` gains:
  - `requireConfirmation?: boolean` — when true, the client should render a confirmation UI.
  - `inputSchema?: JSONValue` — JSON schema for the tool inputs so the client can render an editable form.
- New `ToolCallConfirmationEvent` — a discriminated union:
  ```ts
  | { approved: true; input: JSONValue }
  | { approved: false; input?: JSONValue }
  ```
  This enforces at compile time that `input` is required when approved and optional when rejected.
- `ToolCallEvent` gains an optional `confirmToolCall` field carrying that event.

## SDK surface

**`ToolCallStream`** (`tool-call.types.ts`)
- `onToolCallConfirm(cb)` — register a handler for confirmation responses on a specific tool call.
- `sendToolCallConfirm(event)` — send approve/reject from the client. Throws if the tool call has already ended. Replaces the legacy `sendInterruptEnd` confirmation flow.

**`MessageStream`** (`message.types.ts`)
- `onToolCallConfirm(cb)` — message-level handler that fires for *any* confirmed tool call on the message, even when no per-tool-call helper exists for that id.

### Why both message-level and tool-call-level dispatch?

A per-tool-call helper isn't always present when the confirmation arrives:
- On the agent side, the originating helper may already be cleaned up.
- On the client side, no per-tool-call helper is created if `onToolCallStart` isn't registered.

The message-level handler runs first (flat dispatch in `MessageEventHelperImpl.dispatch`), then the event is delegated to the per-tool-call helper if one exists. Tested explicitly: ordering, no-helper case, and per-tool-call dispatch.

## Tests

- `tool-call-event-helper.test.ts` — register/unregister, `sendToolCallConfirm` for both approve/reject, post-end throw behavior, dispatch routing, and isolation from `endToolCall`.
- `message-event-helper.test.ts` — register/unregister, message-level dispatch, dispatch when no per-tool-call helper exists, and message→tool-call ordering.

## Out of scope

- No widget UI changes — this PR is the protocol/SDK plumbing only. The widget will consume `onToolCallConfirm` / `inputSchema` in a follow-up [here](https://github.com/UiPath/uipath-ui-widgets/pull/67).
- The legacy interrupt-based confirmation flow remains; this is additive.

### DIFF
--- a/src/models/conversational-agent/conversations/types/events/message.types.ts
+++ b/src/models/conversational-agent/conversations/types/events/message.types.ts
@@ -17,6 +17,7 @@ import type {
   MessageEvent,
   MessageStartEvent,
   MetaEvent,
+  ToolCallConfirmationEvent,
   ToolCallStartEvent
 } from './protocol.types';
 import type { CompletedContentPart, ContentPartStream } from './content-part.types';
@@ -340,6 +341,26 @@ export interface MessageStream {
    * ```
    */
   sendInterruptEnd(interruptId: string, endInterrupt: InterruptEndEvent): void;
+
+  /**
+   * Registers a handler for tool-call confirmation events on this message
+   *
+   * Fired when a peer responds to a tool call that was emitted with
+   * `requireConfirmation: true`. The handler runs at the message level, so it
+   * fires even if no per-tool-call stream exists for the confirmed `toolCallId`.
+   *
+   * @param cb - Callback receiving the toolCallId and the confirmation event
+   * @returns Cleanup function to remove the handler
+   *
+   * @example Handling a tool-call confirmation response
+   * ```typescript
+   * message.onToolCallConfirm(({ toolCallId, confirmEvent }) => {
+   *   if (confirmEvent.approved) executeTool(toolCallId, confirmEvent.input);
+   *   else cancelToolCall(toolCallId);
+   * });
+   * ```
+   */
+  onToolCallConfirm(cb: (args: { toolCallId: string; confirmEvent: ToolCallConfirmationEvent }) => void): () => void;
 
   // ==================== Content Part Management ====================
 
--- a/src/models/conversational-agent/conversations/types/events/message.types.ts
+++ b/src/models/conversational-agent/conversations/types/events/message.types.ts
@@ -349,7 +349,7 @@ export interface MessageStream {
    * `requireConfirmation: true`. The handler runs at the message level, so it
    * fires even if no per-tool-call stream exists for the confirmed `toolCallId`.
    *
-   * @param cb - Callback receiving the toolCallId and the confirmation event
+   * @param callback - Callback receiving the toolCallId and the confirmation event
    * @returns Cleanup function to remove the handler
    *
    * @example Handling a tool-call confirmation response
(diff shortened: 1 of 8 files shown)


ninja-shreyash (commented):
Adding more details to the PR desc would help reviewers

deepeshrai-tech:
should we add @deprecated for the legacy types then now that we have moved from interrupt based logic(ToolCallConfirmationValue, ToolCallConfirmationEndValue, ToolCallConfirmationInterruptStartEvent)

src/services/conversational-agent/helpers/tool-call-event-helper.ts
```diff
@@ -106,6 +109,30 @@ export abstract class ToolCallEventHelper extends ConversationEventHelperBase<
...
+    };
+  }
+
+  /**
+   * Sends a tool call confirmation (approve/reject) for a tool call that was
+   * emitted with `requireConfirmation: true`. Replaces the legacy
+   * `sendInterruptEnd` flow for tool call confirmation.
+   * @throws Error if tool call has already ended.
+   */
+  public sendToolCallConfirm(confirmToolCall: ToolCallConfirmationEvent) {
```
deepeshrai-tech, line 131: what happens if no tool is pending confirmation? Are late confirmations tolerated by design?

scottcmg: The runtime will silently drop any event for a tool call it didn't request/doesn't need, the SDK doesn't need to do anything there.

Late confirmations are tolerated, the runtime sits in a wait state until a confirm event resumes the flow.

deepeshrai-tech (approved):

JoshParkSJ (approved):

src/models/conversational-agent/conversations/types/events/message.types.ts
```diff
@@ -341,6 +342,26 @@ export interface MessageStream {
...
   sendInterruptEnd(interruptId: string, endInterrupt: InterruptEndEvent): void;
 
+  /**
+   * Registers a handler for tool-call confirmation events on this message
+   *
+   * Fired when a peer responds to a tool call that was emitted with
+   * `requireConfirmation: true`. The handler runs at the message level, so it
+   * fires even if no per-tool-call stream exists for the confirmed `toolCallId`.
+   *
+   * @param cb - Callback receiving the toolCallId and the confirmation event
```
vnaren23: Please avoid short names. Lets call it `@param callback`

vnaren23 (commented):
Minor changes. 

Also can you add a code snippet example of how a user would need to handle this migration from old to new tool confirmation.

vnaren23 (approved):

PR #408, opened 2026-05-05
fix(auth): add setMultiLogin for login picker

dsuresh-ap:
## Summary
- replace the OAuth constructor opt-out with `sdk.setMultiLogin()`
- keep default authorize URL behavior unchanged with `acr_values` included
- omit `acr_values` only after multi-login is enabled before the OAuth flow
- remove the README/authentication docs that documented the constructor flag

## Tests
- npm run test:unit -- tests/unit/core/auth/service.test.ts tests/unit/core/uipath.test.ts
- npm run typecheck
- npm run lint
- git diff --check

No version bump included.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -40,15 +40,6 @@ const sdk = new UiPath();
 await sdk.initialize();
 ```
 
-To omit `acr_values` from the OAuth authorize URL, pass a partial constructor override while keeping the rest of the configuration injected by the platform:
-
-```typescript
-const sdk = new UiPath({ includeAcrValues: false });
-await sdk.initialize();
-```
-
-This option is constructor-only and is not read from `uipath.json` or injected meta tags.
-
 ## Secret-based Authentication
 ```typescript
 import { UiPath } from '@uipath/uipath-typescript/core';
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -40,6 +40,15 @@ const sdk = new UiPath();
 await sdk.initialize();
 ```
 
+To omit `acr_values` from the OAuth authorize URL, pass a partial constructor override while keeping the rest of the configuration injected by the platform:
+
+```typescript
+const sdk = new UiPath({ includeAcrValues: false });
+await sdk.initialize();
+```
+
+This option is constructor-only and is not read from `uipath.json` or injected meta tags.
+
 ## Secret-based Authentication
 ```typescript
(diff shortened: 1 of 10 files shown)


docs/authentication.md
```diff
@@ -40,6 +40,15 @@ const sdk = new UiPath();
 await sdk.initialize();
 ```
 
+To omit `acr_values` from the OAuth authorize URL, pass a partial constructor override while keeping the rest of the configuration injected by the platform:
+
+```typescript
+const sdk = new UiPath({ includeAcrValues: false });
```
Raina451: lets do this instead:

instead of introducing new field in constructor, lets have a separate method `sdk.useLoginPicker()`

so expectation from dev would be:

```
const sdk = new UiPath();
sdk.useLoginPicker();
sdk.initialize();
```

and useLoginPicker method can be used internally to disable acrvalues in query params for getAuthorizationUrl.
for eg, it can set `this.skipAcrValues = true` and getAuthorizationUrl method checks for this flag to decide on auth url

CC: @vnaren23

Raina451: also let's not document this.

vnaren23: Lets name it better. `sdk.setMultiLogin` maybe. MultiLogin being allowing user to either use SSO or OAuth.

src/core/uipath.ts
```diff
@@ -183,6 +187,14 @@ export class UiPath implements IUiPath {
     }
   }
 
+  /**
+   * Enables multi-login behavior for OAuth by showing the login picker.
+   */
+  public setMultiLogin(): void {
+    this.#multiLogin = true;
+    this.#authService?.setMultiLogin();
+  }
```
Raina451, line 198: lets also add the guard:
if already initialised, lets throw error something like setMultiLogin must be called before initialize.

src/core/auth/service.ts
```diff
@@ -117,6 +118,13 @@ export class AuthService {
     return this.tokenManager;
   }
 
+  /**
+   * Enables login picker behavior by omitting acr_values from the authorization URL.
+   */
```
Raina451, line 125: add @internal tag to hide it from docs.

Raina451 (approved):
minor comments, rest lgtm

src/core/uipath.ts
dsuresh-ap: As discussed will skip since this is undocumented

sysm (approved):

PR #409, opened 2026-05-06
fix(docs): remove clientid and clientsecret from login cmd flags

Raina451:
codedapp publish , deploy cmds cannot be used with clientid and client secret as 
1. it requires apps scope which are internal and not onboarded to external app.
2. current apis are not supported for service.internal token types

[PLT-103329](https://uipath.atlassian.net/browse/PLT-103329)

[PLT-103329]: https://uipath.atlassian.net/browse/PLT-103329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/coded-apps/cli-reference.md
+++ b/docs/coded-apps/cli-reference.md
@@ -48,6 +48,9 @@ $ uip login [options]
 | `-t, --tenant` | string | Tenant name (non-interactive mode) | — |
 | `--it, --interactive` | boolean | Interactively select tenant from list | — |
 
+!!! note
+    Publishing and deploying coded apps with `client-id` and `client-secret` based login is not currently supported. Use interactive browser-based user OAuth login only.
+
 **Examples**
 
 <!-- termynal -->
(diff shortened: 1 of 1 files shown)


docs/coded-apps/cli-reference.md
vnaren23, line 1: Lets also add a Note section mentioning the same. ie currently that support does not exist.

maninder-uipath, line 1: are these flags already removed in the CLI ?

Raina451: no, they wont be removed. 
uip cli does support login with clientid and clientsecret.

but codedapp tool cmds does not. 
adding [this](https://github.com/UiPath/uipath-typescript/pull/409#discussion_r3193668194)

maninder-uipath (approved):

swati354 (approved):

PR #410, opened 2026-05-06
feat: creating common telemetry package

Sandeepan-Ghosh-0312:
## Description
Creating a common package for telemetry

Check consumption in https://github.com/UiPath/uipath-typescript/pull/413 for coded-action-app sdk and ts-sdk

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
         "lint": "oxlint",
         "typecheck": "tsc --noEmit",
         "test": "vitest",
-        "test:unit": "vitest tests/unit",
+        "test:unit": "vitest run",
         "test:integration": "vitest --config vitest.integration.config.ts",
         "test:integration:v0": "vitest --config vitest.integration.config.ts --testNamePattern='\\[v0\\]'",
         "test:integration:v1": "vitest --config vitest.integration.config.ts --testNamePattern='\\[v1\\]'",
--- a/packages/telemetry/package.json
+++ b/packages/telemetry/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@uipath/core-telemetry",
+  "version": "1.0.0",
+  "description": "Shared typescript telemetry package. Sends Application Insights custom events through OpenTelemetry's batched log pipeline.",
+  "license": "MIT",
+  "type": "module",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "clean": "rimraf dist && rimraf node_modules",
+    "prebuild": "rimraf dist",
+    "build": "rollup -c",
+    "typecheck": "tsc --noEmit",
+    "lint": "oxlint",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "dependencies": {
+    "@opentelemetry/sdk-logs": "^0.204.0"
+  },
+  "devDependencies": {
+    "@rollup/plugin-commonjs": "^25.0.0",
+    "@rollup/plugin-node-resolve": "^15.2.0",
+    "@rollup/plugin-typescript": "^11.1.0",
+    "builtin-modules": "^3.3.0",
+    "oxlint": "^1.43.0",
+    "rimraf": "^6.0.1",
+    "rollup": "^4.60.2",
+    "rollup-plugin-dts": "^6.1.0",
+    "tslib": "^2.8.1",
+    "typescript": "^5.3.3",
+    "vitest": "^3.2.4"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/UiPath/uipath-typescript.git",
+    "directory": "packages/telemetry"
+  }
+}
--- a/packages/telemetry/src/client.ts
+++ b/packages/telemetry/src/client.ts
@@ -1,0 +1,296 @@
+import {
+    BatchLogRecordProcessor,
+    LoggerProvider,
+} from '@opentelemetry/sdk-logs';
+import type {
+    LogRecordExporter,
+    ReadableLogRecord,
+} from '@opentelemetry/sdk-logs';
+
+type Logger = ReturnType<LoggerProvider['getLogger']>;
+
+import {
+    APP_NAME,
+    CLOUD_CLIENT_ID,
+    CLOUD_ORGANIZATION_NAME,
+    CLOUD_REDIRECT_URI,
+    CLOUD_TENANT_NAME,
+    CLOUD_URL,
+    CONNECTION_STRING,
+    SERVICE,
+    UNKNOWN,
+    VERSION,
+} from './constants';
+import {
+    TelemetryAttributes,
+    TelemetryClientInitOptions,
+    TelemetryContext,
+} from './types';
+
+/** Result-callback parameter type derived from the OpenTelemetry interface. */
+type ExportResultCallback = Parameters<LogRecordExporter['export']>[1];
+
+const INSTRUMENTATION_KEY_RE = /InstrumentationKey=([^;]+)/;
+const INGESTION_ENDPOINT_RE = /IngestionEndpoint=([^;]+)/;
+
+/**
+ * Application Insights ingestion payload shape (subset).
+ *
+ * Documented at:
+ * https://learn.microsoft.com/azure/azure-monitor/app/data-model-event-telemetry
+ */
+interface ApplicationInsightsEnvelope {
+    name: 'Microsoft.ApplicationInsights.Event';
+    time: string;
+    iKey: string;
+    data: {
+        baseType: 'EventData';
+        baseData: {
+            ver: 2;
+            name: string;
+            properties: Record<string, string>;
+        };
+    };
+    tags: {
+        'ai.cloud.role': string;
+        'ai.cloud.roleInstance': string;
+    };
+}
+
+interface ApplicationInsightsTagsConfig {
+    cloudRoleName: string;
+    cloudRoleInstance: string;
+}
+
+/**
+ * Sends every emitted log record as an Application Insights custom event.
+ * Plugged into OpenTelemetry's `BatchLogRecordProcessor`, which buffers and
+ * flushes log records to this exporter on its own schedule.
+ */
+class ApplicationInsightsEventExporter implements LogRecordExporter {
+    private readonly connectionString: string;
+    private readonly tags: ApplicationInsightsTagsConfig;
+
+    constructor(connectionString: string, tags: ApplicationInsightsTagsConfig) {
+        this.connectionString = connectionString;
+        this.tags = tags;
+    }
+
+    public export(
+        logs: ReadableLogRecord[],
+        resultCallback: ExportResultCallback
+    ): void {
+        try {
+            for (const log of logs) {
+                this.sendAsCustomEvent(log);
+            }
+            resultCallback({ code: 0 });
+        } catch (error) {
+            const err = error instanceof Error ? error : new Error(String(error));
+            console.debug('Failed to export logs to Application Insights:', err);
+            resultCallback({ code: 1, error: err });
+        }
+    }
+
+    public shutdown(): Promise<void> {
+        return Promise.resolve();
+    }
+
+    private sendAsCustomEvent(logRecord: ReadableLogRecord): void {
+        const eventName = String(logRecord.body);
+
+        const payload: ApplicationInsightsEnvelope = {
+            name: 'Microsoft.ApplicationInsights.Event',
+            time: new Date().toISOString(),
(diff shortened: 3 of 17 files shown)


src/core/telemetry/index.ts
```diff
@@ -1,7 +1,46 @@
...
+    SDK_SERVICE_NAME,
+    SDK_VERSION,
+} from './constants';
+
+export { track, trackEvent } from '@uipath/telemetry';
+export type {
+    TelemetryAttributes,
+    TelemetryContext as TelemetryConfig,
+    TrackOptions,
+} from '@uipath/telemetry';
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: export type` is required here, not optional: tsconfig has
 `isolatedModules: true`, which makes plain `export { Foo }` an error
 when `Foo` is a type-only declaration (TS1205). The convention's
 "barrels must use `export *`" rule targets the wildcard form
 (`export type *`) which silently drops runtime values from a re-export
 — that risk doesn't apply to named type-only re-exports under
 isolatedModules, because TypeScript would fail to compile rather than
 silently drop a value-shaped export.

.github/workflows/publish.yml
```diff
@@ -2,16 +2,26 @@
...
+        default: 'ts-sdk'
+        type: choice
+        options:
+          - ts-sdk
+          - coded-action-apps-sdk
+          - telemetry
 
 jobs:
   publish-sdk:
+    if: github.event.inputs.package == 'ts-sdk'
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: fixed

packages/telemetry/tests/track.test.ts
```diff
@@ -0,0 +1,273 @@
...
+ */
+function makeClient(defaultEventName = 'Default.Run'): {
+    client: TelemetryClient;
+    fake: FakeClient;
+} {
+    const fake: FakeClient = {
+        track: vi.fn(),
+        getDefaultEventName: vi.fn(() => defaultEventName),
+    };
+    return { client: fake as unknown as TelemetryClient, fake };
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: the fix for this is too complicated, makes the main code messier and is not needed

packages/telemetry/tests/track.test.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    it('evaluates a condition callback against the call args', () => {
+        const track = createTrack(client);
+        const condition = vi.fn((value: number) => value > 10);
+
+        class Service {
+            // The cast to `(...args: unknown[]) => boolean` matches the
+            // public `TrackOptions.condition` signature; the callback
+            // sees the runtime args as expected.
+            @track('Tasks.Create', { condition: condition as unknown as (...args: unknown[]) => boolean })
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: the fix for this is too complicated, makes the main code messier and is not needed

ninja-shreyash:
### Code review

Found 2 issues:

1. `track` decorator fires telemetry BEFORE the wrapped method executes. For async methods (all SDK service methods), the event is emitted with no knowledge of the outcome — errors are recorded as successes, and duration cannot be measured. A method that always throws will generate "success" telemetry. The decorator should await the result (or catch rejection) before emitting, or at minimum attach success/failure attributes post-execution.

https://github.com/UiPath/uipath-typescript/blob/8d86ddd33681a2faefb80f1e8a710e8b8b7c1dbe/packages/telemetry/src/track.ts#L43-L55

2. `npm run test:unit` runs `vitest tests/unit` — an explicit path that bypasses the `vitest.config.ts` `projects` array. The telemetry tests under `packages/telemetry/tests/` only run via plain `npm test`. `npm run test:all` chains `test:unit && test:integration`, so it also skips telemetry tests. Any CI using the documented `test:unit` or `test:all` scripts silently has zero telemetry test coverage.

https://github.com/UiPath/uipath-typescript/blob/8d86ddd33681a2faefb80f1e8a710e8b8b7c1dbe/vitest.config.ts#L21-L35

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Sandeepan-Ghosh-0312:
Fixed the second issue.
For the first point, as an api-wrapper the SDK should not care about the success of the api call, those are tracked by respective FPS. For the SDK, the telemetry should just be fire and forget when the method is invoked.

packages/telemetry/src/client.ts
```diff
@@ -0,0 +1,294 @@
...
+    ): void {
+        try {
+            for (const log of logs) {
+                this.sendAsCustomEvent(log);
+            }
+            resultCallback({ code: 0 });
+        } catch (error) {
+            const err = error instanceof Error ? error : new Error(String(error));
+            console.debug('Failed to export logs to Application Insights:', err);
+            resultCallback({ code: 1, error: err });
```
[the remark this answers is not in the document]
Sandeepan-Ghosh-0312: no need to add a package for a single enum

packages/telemetry/src/client.ts
```diff
@@ -0,0 +1,294 @@
...
+    private logProvider?: LoggerProvider;
+    private logger?: Logger;
+    private telemetryContext?: TelemetryContext;
+
+    public initialize(options: TelemetryClientInitOptions): void {
+        if (this.isInitialized) {
+            return;
+        }
+
+        this.isInitialized = true;
```
ninja-shreyash, line 196: we should set this after the connection string is validated?

packages/telemetry/src/constants.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Common telemetry attribute keys + Application Insights connection string.
+ *
+ * Producer-specific identity (SDK version, role name, etc.) is supplied at
+ * runtime via `TelemetryClient.initialize(...)` so the same package can be
+ * reused by multiple packages. The Application Insights connection string is
+ * shared across all consumers and is patched into this file at publish time.
+ */
+
+// Connection string placeholder that will be replaced during build
```
ninja-shreyash, line 10: So all consumers will use the same connection string - since it gets baked in at build time?

packages/telemetry/src/client.ts
```diff
@@ -0,0 +1,294 @@
...
+        this.tags = tags;
+    }
+
+    public export(
+        logs: ReadableLogRecord[],
+        resultCallback: ExportResultCallback
+    ): void {
+        try {
+            for (const log of logs) {
+                this.sendAsCustomEvent(log);
```
ninja-shreyash, line 85: should we wait for promises to resolve before returning resultCallback?

packages/telemetry/src/client.ts
```diff
@@ -0,0 +1,294 @@
...
+export class TelemetryClient {
+    private isInitialized = false;
+    private options?: TelemetryClientInitOptions;
+    private logProvider?: LoggerProvider;
+    private logger?: Logger;
+    private telemetryContext?: TelemetryContext;
+
+    public initialize(options: TelemetryClientInitOptions): void {
+        if (this.isInitialized) {
+            return;
```
ninja-shreyash, line 193: we should add a debug log here that the sdk is already initialized

ninja-shreyash: because creating a client and initialize is a two-step process

ninja-shreyash (commented):
Are we going to cleanup existing pkg/core/telemetry code?

packages/telemetry/package.json
```diff
@@ -0,0 +1,57 @@
...
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "files": [
+    "dist",
+    "README.md"
```
ninja-shreyash: there's no readme.md

packages/telemetry/src/client.ts
Sandeepan-Ghosh-0312: it doesn't matter if the connection string is invalid (which it never should) it will never initialize in any case

packages/telemetry/src/constants.ts
Sandeepan-Ghosh-0312: yes

packages/telemetry/src/client.ts
Sandeepan-Ghosh-0312: not needed, current telemetry implementation is also fire and forget and works well.

packages/telemetry/src/client.ts
Sandeepan-Ghosh-0312: why is it needed ? Consumer packages should only call this once, end users are not going to use this directly

Sandeepan-Ghosh-0312: and if it is already initialized there is no point in adding a debug log, telemetry will still work

ninja-shreyash: but if it is already initialized.. and someone tries initializing again. then the new config is silently dropped

ninja-shreyash (approved):

package.json
```diff
@@ -183,7 +183,7 @@
         "lint": "oxlint",
         "typecheck": "tsc --noEmit",
         "test": "vitest",
-        "test:unit": "vitest tests/unit",
+        "test:unit": "vitest run",
```
Raina451, line 186: why this change?

packages/telemetry/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "sourceMap": true,
+    "strict": true,
+    "noImplicitAny": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
Raina451: not sure if `@track` decorator works with [reflect-metadata](https://www.npmjs.com/package/reflect-metadata). 
ref: https://www.typescriptlang.org/tsconfig/#emitDecoratorMetadata

can you check and remove if not needed?

packages/telemetry/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "prebuild": "rimraf dist",
+    "build": "rollup -c",
+    "typecheck": "tsc --noEmit",
+    "lint": "oxlint",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "dependencies": {
+    "@opentelemetry/api-logs": "^0.204.0",
+    "@opentelemetry/sdk-logs": "^0.204.0"
```
Raina451, line 35: why we added api-logs depndency, till now we only had:
"@opentelemetry/sdk-logs" in non common telemetry package

packages/telemetry/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@uipath/telemetry",
```
Raina451: this name already exist, telemtry package is being published by uip cli team, its a common package.
https://www.npmjs.com/package/@uipath/telemetry

1. have different name
2. and should we check if we can use that package itself? 😅

packages/telemetry/package.json
Sandeepan-Ghosh-0312: removed

packages/telemetry/package.json
Sandeepan-Ghosh-0312: decided to use this package with name core-telemetry, the uip telemetry package cannot be reused effectively here

package.json
Sandeepan-Ghosh-0312: test:unit will now run all the unit tests in sdk and telemetry package as defined in projects in vite.config.ts

packages/telemetry/tsconfig.json
Sandeepan-Ghosh-0312: removed

packages/telemetry/src/client.ts
```diff
@@ -0,0 +1,296 @@
...
+            resultCallback({ code: 1, error: err });
+        }
+    }
+
+    public shutdown(): Promise<void> {
+        return Promise.resolve();
+    }
+
+    private sendAsCustomEvent(logRecord: ReadableLogRecord): void {
+        const eventName = String(logRecord.body ?? '');
```
Raina451: can we default to "sdk.run" or something rather than ''?

packages/telemetry/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export type { Track } from './track';
+export {
+    APP_NAME,
+    CLOUD_CLIENT_ID,
+    CLOUD_ORGANIZATION_NAME,
+    CLOUD_REDIRECT_URI,
+    CLOUD_TENANT_NAME,
+    CLOUD_URL,
+    SERVICE,
+    UNKNOWN,
```
Raina451: nit: Are the attribute key constants and `UNKNOWN` intentionally public? 
If consumers only use `TelemetryClient`/`createTrack` etc, these could stay internal. ?

Raina451 (approved):
minor nits

packages/telemetry/src/client.ts
Sandeepan-Ghosh-0312: this scenario cannot be reached, just there for OTEL typecheck, removing the operator

packages/telemetry/src/index.ts
Sandeepan-Ghosh-0312: removed

PR #425, opened 2026-05-09
feat(conversations): add release and search filters to getAll

scottcmg:
## Summary
- Adds optional `agentReleaseKey`, `agentReleaseId`, and `search` filter options to `ConversationService.getAll()`. They forward as plain query params (not OData `$`-prefixed) so callers can scope conversation history to a single agent release and substring-match labels.
- Exposes an optional `releaseKey` field on `RawAgentGetResponse` so consumers can read the GUID off an agent and feed it into the new filter.
- Restructures the `getAll` JSDoc into named examples (basic / pagination / sorted / filter+search) and clarifies the `processKey` field as a dotted-path string.

## Test plan
- [x] `npm run typecheck` clean
- [x] `npm run lint` clean
- [x] `npm run test:unit` — 1263 tests pass; new unit test asserts the three options forward to `PaginationHelpers` and land in `excludeFromPrefix`
- [x] `npm run build` clean
- [x] Manually verified against `agents/frontend` consuming the new `uipath-ui-widgets`: request URL includes the new filter params when widget refs are populated
- [x] Backwards-compat verified by running the older worktree widget against the new SDK — request shape unchanged, no breakage

### DIFF
--- a/src/models/conversational-agent/conversations/conversations.types.ts
+++ b/src/models/conversational-agent/conversations/conversations.types.ts
@@ -73,14 +73,14 @@ export interface ConversationUpdateOptions {
 export type ConversationGetAllOptions = PaginationOptions & {
   /** Sort order for conversations */
   sort?: SortOrder;
-  /** GUID key of the agent release to filter conversations by. */
-  agentReleaseKey?: string;
-  /** Numeric ID of the agent release to filter conversations by. */
-  agentReleaseId?: number;
+  /** GUID key of the agent to filter conversations by. */
+  agentKey?: string;
+  /** Numeric ID of the agent to filter conversations by. */
+  agentId?: number;
   /**
    * Case-insensitive substring filter applied to conversation labels (1–100 chars).
    */
-  search?: string;
+  label?: string;
 }
 
 // ==================== Attachment Types ====================
--- a/src/models/conversational-agent/conversations/conversations.types.ts
+++ b/src/models/conversational-agent/conversations/conversations.types.ts
@@ -73,6 +73,14 @@ export interface ConversationUpdateOptions {
 export type ConversationGetAllOptions = PaginationOptions & {
   /** Sort order for conversations */
   sort?: SortOrder;
+  /** GUID key of the agent release to filter conversations by. */
+  agentReleaseKey?: string;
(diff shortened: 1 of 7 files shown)


src/models/conversational-agent/conversations/conversations.types.ts
```diff
@@ -71,13 +68,19 @@ export interface ConversationUpdateOptions {
...
   /** Sort order for conversations */
   sort?: SortOrder;
+  /** GUID key of the agent release to filter conversations by. */
+  agentReleaseKey?: string;
+  /** Numeric ID of the agent release to filter conversations by. */
+  agentReleaseId?: number;
+  /**
+   * Case-insensitive substring filter applied to conversation labels (1–100 chars).
+   */
+  search?: string;
```
vnaren23: Can we rename these to 

```
agentKey
agentId
label
```

Internally we can map it each parameter to match with backend naming. From a dev/SDK user perspective current naming does not make sense.

vnaren23 (approved):

src/models/conversational-agent/conversations/conversations.types.ts
```diff
@@ -71,13 +68,19 @@ export interface ConversationUpdateOptions {
...
   /** Sort order for conversations */
   sort?: SortOrder;
+  /** GUID key of the agent to filter conversations by. */
+  agentKey?: string;
+  /** Numeric ID of the agent to filter conversations by. */
+  agentId?: number;
+  /**
+   * Case-insensitive substring filter applied to conversation labels (1–100 chars).
+   */
+  label?: string;
```
rahulmukherjee97, line 83: As mentioned in the comment `labels (1–100 chars).` are we validating it at the time of search before requesting it to conversational agent API?

deepeshrai-tech (approved):

src/models/conversational-agent/conversations/conversations.types.ts
scottcmg: This is only validated in the backend. If we duplicate the check SDK-side the two rules can drift out of sync so I think for now it's safe to keep the validation strictly in the backend and just surface the restriction in the JSDoc. I'm open to discussion about this though, and if needed let me know and I'll open a follow-up PR to add.

PR #446, opened 2026-05-19
ci: publish docs only on SDK release or manual dispatch [PLT-104071]

Raina451:
## Summary

- Docs no longer publish on every PR merge to main
- Now publish via two paths:
  - **SDK release** — publish workflow automatically triggers docs deployment after successful publish
  - **Manual dispatch** — trigger from Actions UI for doc-only changes

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,6 @@
 name: Deploy Documentation
 on:
-  push:
-    branches:
-      - main
+  workflow_call:
   workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
maninder-uipath, line 1: > Docs-only changes — auto-publishes when only docs/** or mkdocs.yml change on main

This is wrong. If a new feature has been added and docs have been updated, but new version has not been released, then docs will reflect unpublished changes.

Raina451: oh you mean if already unreleased code merged in main and then another pr for md file changes only then it will publish all docs?

Raina451: makes sense , i will remove this

Raina451: updated

maninder-uipath (approved):

ninja-shreyash (approved):

vnaren23 (approved):

PR #447, opened 2026-05-19
fix(core): throw ServerError instead of NetworkError for non-JSON responses

Sarath1018:
## Summary

- `JSON.parse()` failures on successful HTTP responses (e.g. server returns HTML for an invalid org/tenant redirect) were caught by the outer `fetch()` catch block and incorrectly wrapped as `NetworkError`
- Extracted `JSON.parse` into its own inner try/catch that throws `ServerError` with the actual HTTP status code, keeping the outer catch exclusively for genuine network/fetch failures
- Updated the auth-errors integration test to check for `ServerError` instead of the previous `NetworkError`/`SyntaxError` workaround

## Test plan

- [ ] `npm run typecheck` — clean
- [ ] `npm run lint` — 0 errors
- [ ] `npm run test:unit` — all 1527 tests pass
- [ ] `npm run build` — dist/ output produced
- [ ] Integration: invalid org/tenant now surfaces as `ServerError` (not `NetworkError`) with `statusCode` and descriptive message

## Files

| Area | Files |
|------|-------|
| Core HTTP | `src/core/http/api-client.ts` |
| Integration test | `tests/integration/auth-errors.integration.test.ts` |

### DIFF
--- a/src/core/http/api-client.ts
+++ b/src/core/http/api-client.ts
@@ -4,6 +4,7 @@ import { RequestSpec } from '../../models/common/request-spec';
 import { TokenManager } from '../auth/token-manager';
 import { errorResponseParser } from '../errors/parser';
 import { ErrorFactory } from '../errors/error-factory';
+import { ServerError } from '../errors/server';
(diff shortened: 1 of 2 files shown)


Sarath1018:
Closing to run sdk-review before re-raising.

PR #453, opened 2026-05-20
feat(assets): add updateValueById for PUT /odata/Assets({key})

maninder-uipath:
https://uipath.atlassian.net/browse/PLT-103905

Adds `Assets.updateValueById(id, folderId, newValue)` that fetches the asset to determine its `valueType`, validates the newValue runtime type matches (Text/Integer/Bool only), and PUTs the update while preserving the asset's name, scope, and description. Unsupported valueTypes (Credential, Secret) throw ValidationError.

[Swagger](https://alpha.uipath.com/avinash/addy_test_22/orchestrator_/swagger/index.html#/Assets/Assets_Put) for reference

### DIFF
--- a/src/models/orchestrator/assets.models.ts
+++ b/src/models/orchestrator/assets.models.ts
@@ -1,4 +1,4 @@
-import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions, AssetGetByNameOptions } from './assets.types';
+import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions, AssetGetByNameOptions, AssetNewValue, AssetUpdateValueByIdOptions } from './assets.types';
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
 
 /**
@@ -90,4 +90,37 @@ export interface AssetServiceModel {
    * ```
    */
   getByName(name: string, options?: AssetGetByNameOptions): Promise<AssetGetResponse>;
+
+  /**
+   * Updates the value of an existing asset by ID.
+   *
+   * Fetches the asset internally to determine its type, then updates only the value while
+   * preserving the asset's name, scope, and description.
+   *
+   * **Supported value types:** `Text`, `Integer`, and `Bool` only. Other types
+   * (`Credential`, `Secret`) throw a `ValidationError`.
+   *
+   * The `newValue` runtime type must match the asset's `valueType`:
+   * - `Text` → `string`
+   * - `Integer` → `number` (integer)
+   * - `Bool` → `boolean`
+   *
+   * @param id - Asset ID
+   * @param newValue - New value to apply (string for `Text`, number for `Integer`, boolean for `Bool`)
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`)
+   * @returns Promise resolving when the asset has been updated
+   *
+   * @example
+   * ```typescript
+   * // Update a Text asset by folder ID
+   * await assets.updateValueById(<assetId>, 'new-value', { folderId: <folderId> });
+   *
+   * // Update an Integer asset by folder key (GUID)
+   * await assets.updateValueById(<assetId>, 42, { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // Update a Bool asset by folder path
+   * await assets.updateValueById(<assetId>, true, { folderPath: 'Shared/Finance' });
+   * ```
+   */
+  updateValueById(id: number, newValue: AssetNewValue, options?: AssetUpdateValueByIdOptions): Promise<void>;
 }
--- a/src/models/orchestrator/assets.types.ts
+++ b/src/models/orchestrator/assets.types.ts
@@ -13,14 +13,10 @@ export enum AssetValueScope {
  * Enum for Asset Value Type
  */
 export enum AssetValueType {
-  DBConnectionString = 'DBConnectionString',
-  HttpConnectionString = 'HttpConnectionString',
   Text = 'Text',
   Bool = 'Bool',
   Integer = 'Integer',
   Credential = 'Credential',
-  WindowsCredential = 'WindowsCredential',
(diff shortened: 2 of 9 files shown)


ninja-shreyash:
pls address claude comments

src/models/orchestrator/assets.models.ts
```diff
@@ -90,4 +90,47 @@ export interface AssetServiceModel {
...
   getByName(name: string, options?: AssetGetByNameOptions): Promise<AssetGetResponse>;
+
+  /**
+   * Updates the value of an existing asset by ID.
+   *
+   * Fetches the asset internally to determine its type, then updates only the value while
+   * preserving the asset's name, scope, and description.
+   *
+   * **Supported value types:** `Text`, `Integer`, and `Bool` only. Other types
+   * (`Credential`, `Secret`) throw a `ValidationError`.
```
Raina451, line 101: better add "for other types * (`Credential`, `Secret`) values cant be updated" if this is the case.

src/models/orchestrator/assets.models.ts
```diff
@@ -90,4 +90,47 @@ export interface AssetServiceModel {
...
+   *
+   * The `newValue` runtime type must match the asset's `valueType`:
+   * - `Text` → `string`
+   * - `Integer` → `number` (integer)
+   * - `Bool` → `boolean`
+   *
+   * A `ValidationError` is thrown when the `newValue` type does not match.
```
Raina451: i dont think we need to add this

src/models/orchestrator/assets.models.ts
```diff
@@ -90,4 +90,47 @@ export interface AssetServiceModel {
...
+   * await assets.updateValueById(<assetId>, <folderId>, 'new-value');
+   *
+   * // Update an Integer asset
+   * await assets.updateValueById(<assetId>, <folderId>, 42);
+   *
+   * // Update a Bool asset
+   * await assets.updateValueById(<assetId>, <folderId>, true);
+   * ```
+   */
+  updateValueById(id: number, folderId: number, newValue: AssetNewValue): Promise<void>;
```
Raina451: can we add folder options in this pr too?

src/models/orchestrator/assets.types.ts
```diff
@@ -73,3 +73,13 @@ export interface AssetGetByIdOptions extends BaseOptions {}
...
+
+/**
+ * New value accepted by {@link AssetServiceModel.updateValueById}.
+ *
+ * The runtime type must match the asset's `valueType`:
+ * - `Text` → `string`
+ * - `Integer` → `number`
+ * - `Bool` → `boolean`
+ */
+export type AssetNewValue = string | number | boolean;
```
Raina451, line 81: can we have enum?

src/models/orchestrator/assets.models.ts
maninder-uipath: why

src/models/orchestrator/assets.types.ts
maninder-uipath: why enum? These have to be primitive types

Raina451 (approved):

src/models/orchestrator/assets.types.ts
```diff
@@ -13,14 +13,10 @@ export enum AssetValueScope {
  * Enum for Asset Value Type
  */
 export enum AssetValueType {
```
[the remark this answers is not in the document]
maninder-uipath: This breaking change is fine because it's the same as in assets FPS

src/models/orchestrator/assets.models.ts
maninder-uipath: There was a misunderstanding. You meant remove the comment, I understood remove the ValidationError.

src/models/orchestrator/assets.models.ts
maninder-uipath: It can be updated, but it'll require more code changes which I'll take care in another PR

src/models/orchestrator/assets.types.ts
maninder-uipath: Discussed offline. Nothing needs to be done here. There was some confusion which is resolved.

src/models/orchestrator/assets.models.ts
```diff
@@ -90,4 +90,41 @@ export interface AssetServiceModel {
...
+   * (`Credential`, `Secret`) throw a `ValidationError`.
+   *
+   * The `newValue` runtime type must match the asset's `valueType`:
+   * - `Text` → `string`
+   * - `Integer` → `number` (integer)
+   * - `Bool` → `boolean`
+   *
+   * Returns void; the API does not return the updated asset, so call
+   * {@link AssetServiceModel.getById} or {@link AssetServiceModel.getByName}
+   * if you need the refreshed value.
```
vnaren23: Is this needed?

src/models/orchestrator/assets.types.ts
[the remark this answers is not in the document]
vnaren23: You mean the backend never had such enums?

src/models/orchestrator/assets.types.ts
vnaren23: Add whatever was the confusion to the thread so others are also aware.

vnaren23 (commented):
1. Have we whitelisted this endpoint on gateway?
2. Please check the PR gate failure.

src/models/orchestrator/assets.models.ts
maninder-uipath: I just added this so that user knows beforehand that this API returns nothing (undefined.. not even empty object). Because usually APIs return something. Why do you think this is not needed?

vnaren23: Adds no value in terms of documentation. a void return is self explanatory.

src/models/orchestrator/assets.types.ts
Raina451: An enum is not appropriate here because this is a value type, not a label type. Enums are for fixed, discrete sets of named constants.

vnaren23 (approved):

PR #473, opened 2026-05-28
feat: add Data Fabric roles and directory services

dsuresh-ap:
## Summary
- Add `DataFabricRoles` and `DataFabricDirectory` services to the SDK surface.
- Introduce typed models for role listing, directory listing, and role assignment payloads/results.
- Wire the new services into `UiPath` legacy accessors and export them from the Data Fabric module.
- Add unit coverage for role listing, directory paging, and role assignment behavior.

## Testing
- Unit tests added for Data Fabric roles and directory services.
- Legacy `UiPath` service wiring test updated to verify the new accessors are exposed.
- Tested against alp


docs/pagination.md
```diff
@@ -111,6 +111,19 @@ console.log(`Retrieved ${allAssets.items.length} assets`);
 console.log(`Total count: ${allAssets.totalCount}`);
 ```
 
+### Service-specific Offset Pagination
+
+Some endpoints expose product-native offset parameters instead of the shared cursor/jump wrapper. Data Fabric Directory uses `list({ skip, top })` for a single page, and `getAll({ pageSize })` to fetch every page.
```
vnaren23: We dont want such product specific pagination. AFAIK we've done similar masking for DF APIs.

docs/oauth-scopes.md
```diff
@@ -67,6 +67,20 @@ This page lists the specific OAuth scopes required in external app for each SDK
...
+|--------|-------------|
+| `getAll()` | `DataFabric.Data.Read` |
+
+## Data Fabric Directory
+
+| Method | OAuth Scope |
+|--------|-------------|
+| `list()` | `DataFabric.Data.Read` |
+| `getAll()` | `DataFabric.Data.Read` |
+| `assignRoles()` | `DataFabric.Data.Write` |
```
vnaren23: @dsuresh-ap Why do we want to add these methods? They do not work with the mentioned scopes.

vnaren23 (changes requested):
Please share use case/need for these methods. 
1. We need them first accessible via external app (which is not the case today)

docs/oauth-scopes.md
dsuresh-ap: @vnaren23 Seems like this was added by Codex when I asked it to review if docs updates are required.

The assumption is that these commands are run by an admin with access to DF. Do I need to document that anywhere? In the meantime going to remove this

vnaren23: Even these methods, is this meant to be for CLI or TS SDK?

src/models/data-fabric/directory.internal-types.ts
```diff
@@ -0,0 +1,28 @@
...
+  isUIEnabled: boolean;
+}
+
+export interface DataFabricDirectoryRevokePayload {
+  externalIds: string[];
+}
+
+export interface RawDataFabricDirectoryListResponse {
+  totalCount?: number;
+  TotalCount?: number;
```
AditiGoyalUipath: Why the response has duplicate object totalCount and TotalCount?. Also 
results and Results

dsuresh-ap: Good catch. That duplication was only defensive raw-envelope fallback code, not an intended response contract. The MRS setup source-of-truth for this Data Fabric Directory endpoint uses lowercase `totalCount` / `results`, so I cleaned this up in `6863ba0d` by removing the PascalCase directory fields and pagination fallbacks. The directory response now keeps the known lowercase fields, plus only the non-casing envelope fallbacks (`value` / `data`).

src/services/data-fabric/roles.ts
```diff
@@ -0,0 +1,60 @@
...
+import { RawDataFabricRoleListResponse } from '../../models/data-fabric/roles.internal-types';
+import { DATA_FABRIC_ENDPOINTS } from '../../utils/constants/endpoints/data-fabric';
+import { createParams } from '../../utils/http/params';
+import { BaseService } from '../base';
+
+function extractRoles(data: RawDataFabricRoleListResponse | DataFabricRole[]): DataFabricRole[] {
+  if (Array.isArray(data)) {
+    return data;
+  }
+  return data.results ?? data.Results ?? data.value ?? data.Value ?? data.data ?? [];
```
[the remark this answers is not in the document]
dsuresh-ap: Applied the same cleanup in `863ebd5c`. I removed the PascalCase role envelope fields from `RawDataFabricRoleListResponse` and dropped `data.Results` / `data.Value` from `extractRoles`, keeping only the lowercase known shape plus the non-casing `value` / `data` fallbacks and array response handling.

src/models/data-fabric/directory.internal-types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface DataFabricDirectoryRevokePayload {
+  externalIds: string[];
+}
+
+export interface RawDataFabricDirectoryListResponse {
+  totalCount?: number;
+  results?: DataFabricDirectoryEntry[];
+  value?: DataFabricDirectoryEntry[];
+  data?: DataFabricDirectoryEntry[];
```
AditiGoyalUipath: directory endpoint doesn't return "data" or even "value" in response. I believe it should only be   export interface RawDataFabricDirectoryListResponse {
    totalCount?: number;
    results?: DataFabricDirectoryEntry[];
  }

src/models/data-fabric/directory.types.ts
```diff
@@ -0,0 +1,108 @@
...
+  /** External application principal. */
+  Application = 2,
+}
+
+/**
+ * @internal
+ */
+export type DataFabricDirectoryEntityTypeInput =
+  | DataFabricDirectoryEntityType
+  | 'User'
```
AditiGoyalUipath: Is this intentional to support lenient casing in directorytype?

src/models/data-fabric/directory.types.ts
```diff
@@ -0,0 +1,108 @@
...
+  | 'DirectoryRobotUser'
+  | 'directoryRobotUser';
+
+/**
+ * @internal
+ */
+export interface DataFabricDirectoryRole {
+  id: string;
+  name?: string;
+  [key: string]: unknown;
```
AditiGoyalUipath: The backend has id: Guid, name: string . 
Is there any specifix reason we kept it open to include any other keys ?
[key: string]: unknown; 
Same in DataFabricDirectoryEntry

src/models/data-fabric/directory.types.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * @internal
+ */
+export interface DataFabricDirectoryEntry {
+  externalId: string;
+  name?: string;
+  email?: string;
+  type?: DataFabricDirectoryEntityType | string;
```
AditiGoyalUipath: Do we need DataFabricDirectoryEntityType and string here? DataFabricDirectoryEntityType is numeric enum right ?

src/models/data-fabric/directory.types.ts
```diff
@@ -0,0 +1,108 @@
...
+  name?: string;
+  [key: string]: unknown;
+}
+
+/**
+ * @internal
+ */
+export interface DataFabricDirectoryEntry {
+  externalId: string;
+  name?: string;
```
AditiGoyalUipath: I believe name, type are not optional . Please check again

src/models/data-fabric/roles.internal-types.ts
```diff
@@ -0,0 +1,7 @@
+import { DataFabricRole } from './roles.types';
+
+export interface RawDataFabricRoleListResponse {
+  results?: DataFabricRole[];
```
AditiGoyalUipath: Please check the input output request body again . Aligned with whatever API response is in Datafabric commonentityplatform code repo

src/services/data-fabric/roles.ts
```diff
@@ -0,0 +1,60 @@
...
+   * ```
+   *
+   * @internal
+   */
+  @track('DataFabricRoles.GetAll')
+  async getAll(options: DataFabricRoleGetAllOptions = {}): Promise<DataFabricRole[]> {
+    const params = createParams({
+      stats: options.stats ?? true,
+    });
+    const response = await this.get<RawDataFabricRoleListResponse | DataFabricRole[]>(
```
AditiGoyalUipath: const response = await this.get<DataFabricRole[]>(

src/services/data-fabric/roles.ts
```diff
@@ -0,0 +1,60 @@
...
+      DATA_FABRIC_ENDPOINTS.ROLES.GET_ALL,
+      { params }
+    );
```
AditiGoyalUipath, line 88: Add validation for response format

src/services/base.ts
```diff
@@ -21,6 +21,16 @@ export interface ApiResponse<T> {
   data: T;
 }
 
+function resolveFirstField(data: Record<string, unknown>, fields: string[]): unknown {
+  for (const field of fields) {
+    const value = resolveNestedField(data, field);
+    if (value !== undefined) {
```
AditiGoyalUipath: Null check ?

AditiGoyalUipath:
Are these new methods tested ?

AditiGoyalUipath (approved):
Tests all methods and verify Folder scope.

README.md
```diff
@@ -381,6 +386,21 @@ await entities.updateRecordsById('entity-uuid', [
...
+if (dataWriter) {
+  await dataFabricDirectory.assignRoles(
+    '<identity-group-id>',
+    'Group',
+    [dataWriter.id]
+  );
+}
+
+// Revoke direct Data Fabric access
+await dataFabricDirectory.revokeRoles('<identity-user-id>');
```
Sarath1018: Since this is internal method do not mention it in Readme

src/services/data-fabric/index.ts
```diff
@@ -24,10 +28,16 @@
...
 // Re-export service-specific types
 export * from '../../models/data-fabric/entities.types';
 export { LogicalOperator, QueryFilterOperator } from '../../models/data-fabric/entities.types';
 export * from '../../models/data-fabric/entities.models';
 export * from '../../models/data-fabric/choicesets.types';
 export * from '../../models/data-fabric/choicesets.models';
+export * from '../../models/data-fabric/roles.types';
+export * from '../../models/data-fabric/roles.models';
+export * from '../../models/data-fabric/directory.types';
+export * from '../../models/data-fabric/directory.models';
```
Sarath1018: internal types shouldn't be re-exported from barrel files. Please remove these four lines

src/models/data-fabric/index.ts
```diff
@@ -1,4 +1,8 @@
 export * from './entities.types';
 export * from './entities.models';
 export * from './choicesets.types';
-export * from './choicesets.models'; 
\ No newline at end of file
+export * from './choicesets.models';
+export * from './roles.types';
+export * from './roles.models';
+export * from './directory.types';
+export * from './directory.models';
```
Sarath1018: Same, internal types shouldn't be re-exported from barrel files. Please remove these four lines

src/services/data-fabric/index.ts
```diff
@@ -24,10 +28,16 @@
 // Export service with cleaner name and keep EntityService for legacy UiPath class
 export { EntityService as Entities, EntityService } from './entities';
 export { ChoiceSetService as ChoiceSets, ChoiceSetService } from './choicesets';
+export { DataFabricRoleService as DataFabricRoles, DataFabricRoleService } from './roles';
+export { DataFabricDirectoryService as DataFabricDirectory, DataFabricDirectoryService } from './directory';
```
Sarath1018: Drop the `as DataFabricRoles` / `as DataFabricDirectory` aliases

src/services/base.ts
```diff
@@ -233,7 +245,9 @@ export class BaseService {
         limitedPageSize = getLimitedPageSize(params.pageSize);
         const pageSizeParam = paginationParams?.pageSizeParam || ODATA_OFFSET_PARAMS.PAGE_SIZE_PARAM;
         const offsetParam = paginationParams?.offsetParam || ODATA_OFFSET_PARAMS.OFFSET_PARAM;
-        const countParam = paginationParams?.countParam || ODATA_OFFSET_PARAMS.COUNT_PARAM;
+        const countParam = paginationParams && Object.prototype.hasOwnProperty.call(paginationParams, 'countParam')
+          ? paginationParams.countParam
+          : ODATA_OFFSET_PARAMS.COUNT_PARAM;
```
Sarath1018: This looks like a correct fix but this impacts other methods so this require testing. We already have a [ticket](https://uipath.atlassian.net/browse/PLT-102334?atlOrigin=eyJpIjoiZTdkYTIwZmRkNTdjNDYzZTg3N2ExNDAzMjE3YmE1MzQiLCJwIjoiaiJ9) to track this bug. Can you drop this change

src/services/base.ts
```diff
@@ -283,18 +297,27 @@ export class BaseService {
...
+    const items = Array.isArray(rawItems) ? rawItems as T[] : [];
+    const rawTotalCount = Array.isArray(response.data)
+      ? undefined
+      : resolveFirstField(responseData, totalCountFields);
     const totalCount = typeof rawTotalCount === 'number' ? rawTotalCount : undefined;
-    const continuationToken = response.data[continuationTokenField];
+    const rawContinuationToken = Array.isArray(response.data)
+      ? undefined
+      : resolveNestedField(responseData, continuationTokenField);
+    const continuationToken = typeof rawContinuationToken === 'string' ? rawContinuationToken : undefined;
```
Sarath1018: These changes in base.ts don't seem to be used by the services in this PR, can you clarify why they were added?

Field fallbacks `(itemsFieldFallbacks, totalCountFieldFallbacks, resolveFirstField)`, I don't see any consumer.

continuationToken typeof string hardening this is in the TOKEN pagination path, but the directory service is PaginationType.OFFSET and never hits it.

src/services/data-fabric/directory.ts
```diff
@@ -0,0 +1,368 @@
...
+   * const firstPrincipal = page.results[0];
+   * ```
+   *
+   * @internal
+   */
+  @track('DataFabricDirectory.List')
+  async list(options: DataFabricDirectoryListOptions = {}): Promise<DataFabricDirectoryListResponse> {
+    const params = createParams({
+      skip: options.skip ?? 0,
+      top: options.top ?? DEFAULT_DIRECTORY_PAGE_SIZE,
```
Sarath1018: Should we also clamp top to <100? similar to `MAX_DIRECTORY_PAGE_SIZE = 100`

src/models/data-fabric/directory.models.ts
```diff
@@ -0,0 +1,137 @@
...
+    principalType: DataFabricDirectoryEntityTypeInput,
+    roleIds: string[],
+    options?: DataFabricDirectoryAssignOptions
+  ): Promise<DataFabricDirectoryAssignmentResult[]>;
+
+  /**
+   * Revokes all direct Data Fabric roles from one or more principals.
+   *
+   * The Data Fabric API removes all role assignments for each supplied external
+   * ID. Use this when a principal should no longer have direct Data Fabric
```
Sarath1018, line 111: question: API allows only to revoke all roles? cant we revoke a particular role?

tests/integration/shared/agents/feedback.integration.test.ts
```diff
@@ -6,12 +6,22 @@ import { registerResource } from '../../utils/cleanup';
...
   let feedback: Feedback;
 
+  beforeAll(() => {
+    if (!hasAgentFeedbackIntegration) {
+      throw new Error(
+        'AGENT_FEEDBACK_INTEGRATION must be set to run Agent Feedback integration tests ' +
+        '(requires feedback API access on the configured external app/PAT)'
+      );
+    }
+  });
```
Sarath1018: Not related to this PR, please drop

tests/.env.integration.example
```diff
@@ -33,6 +33,19 @@ INTEGRATION_TEST_SKIP_CLEANUP=false
...
+DATA_FABRIC_ACCESS_INTEGRATION=false
+
+# Enable Agent Feedback integration tests.
+# Requires feedback API access on the configured token/app.
+AGENT_FEEDBACK_INTEGRATION=false
+
+# Enable Observability Traces integration tests.
+# Requires traces API access and a pre-existing trace ID.
+TRACES_INTEGRATION=false
+TRACES_TEST_TRACE_ID=
```
Sarath1018: Not related to this PR

tests/integration/shared/observability/traces.integration.test.ts
```diff
@@ -16,6 +17,13 @@ describe.each(modes)('Traces - Integration Tests [%s]', (mode) => {
   let existingSpanId!: string;
 
   beforeAll(async () => {
+    if (!hasTracesIntegration) {
+      throw new Error(
+        'TRACES_INTEGRATION must be set to run Traces integration tests ' +
+        '(requires traces API access and a pre-existing trace ID)'
+      );
+    }
+
```
Sarath1018: Please drop all the changes that are not related to roles and directory

src/models/data-fabric/directory.models.ts
dsuresh-ap: @Sarath1018 The UI revokes roles using the `POST /datafabric_/api/Directory/Role`. You pass the roles you want the end state to be. 

`revokeRoles` is to match the UI behavior `/datafabric_/api/Directory/RevokeRole` for removing all roles since the above endpoint doesn't allow empty array.

tests/unit/services/data-fabric/index.ts
```diff
@@ -3,4 +3,6 @@
  */
 
 export * from './entities.test';
-
+export * from './choicesets.test';
+export * from './roles.test';
+export * from './directory.test';
```
Sarath1018: nit: Can we drop the changes to the test-barrel index.ts files as they dont serve any purpose

tests/.env.integration.example
```diff
@@ -33,6 +33,10 @@ INTEGRATION_TEST_SKIP_CLEANUP=false
 # Leave empty to use default folder
 INTEGRATION_TEST_FOLDER_ID=
 
+# Enable Data Fabric role and directory access integration tests.
+# Requires DataFabric.Data.Read and DataFabric.Data.Write on the configured token/app.
+DATA_FABRIC_ACCESS_INTEGRATION=false
```
Sarath1018: Lets not use this approach to not run integration tests.

Use `describe.skip` by commenting 403 reason by mentioning the scopes issue.

Sarath1018 (approved):

vnaren23 (approved):

PR #474, opened 2026-05-29
fix(data-fabric): support RELATIONSHIP/FILE field creation and tighte…

AditiGoyalUipath:
…n name validation

The entity-create payload previously serialized relationship fields as referenceEntityName/referenceFieldName strings the API does not resolve, leaving RELATIONSHIP fields effectively unreachable from the SDK. FILE fields had no way to bind a target attachment entity at all.

Wire-side changes:
- RELATIONSHIP: emit nested referenceEntity/referenceField { id }, plus isForeignKey: true and referenceType: ManyToOne
- FILE: accept referenceEntityId/referenceFieldId, emit the same nested shape plus isForeignKey: true
- Add MULTILINE_MAX_TEXT field type (128 KB byte budget)
- Add isHiddenField on EntityFieldBase + payload

Name validation tightened to match StorageManagementService.ValidateCommonInput:
- Block C# and VB reserved keywords (case-insensitive)
- Reject field name == entity name
- Cap displayName at 128 chars and description at 512 chars
- Extend reserved system field set with RecordOwner and KmsKeyId


src/models/data-fabric/entities.types.ts
```diff
@@ -294,10 +296,10 @@ export interface EntityCreateFieldOptions extends EntityFieldBase {
   type?: EntityFieldDataType;
   /** Choice set ID for choice-set fields */
   choiceSetId?: string;
-  /** Name of the referenced entity for relationship fields */
-  referenceEntityName?: string;
```
vnaren23, line 298: Will this not work anymore? Or has this never worked?

vnaren23 (commented):
Please add integration tests explicitly for this use case.

src/models/data-fabric/entities.types.ts
AditiGoyalUipath: This was never used in any method and not exposed to coded apps

AditiGoyalUipath:
> Please add integration tests explicitly for this use case.

This is Schema CRUD operation. Since scope is not there integration test are not running. Verified it locally and can add it as skipped integration test.

vnaren23 (approved):

src/models/data-fabric/entities.types.ts
```diff
@@ -269,6 +269,8 @@ export interface EntityFieldBase {
   isRbacEnabled?: boolean;
   /** Whether the field value is encrypted at rest (default: false) */
   isEncrypted?: boolean;
+  /** Whether the field is hidden from the UI (default: false) */
+  isHiddenField?: boolean;
```
swati354, line 273: `isHiddenField` is accepted on `updateFields` (via `EntityFieldBase`) but never applied in `applySchemaUpdate's` update spread. Can you check.

swati354 (approved):

PR #476, opened 2026-05-29
feat(data-fabric): support explicit expansions in queryRecordsById

AditiGoyalUipath:
## Summary

- Adds Mode B `expansions` to `queryRecordsById` — per-branch expansion with `expandedField`, `alias`, `selectedFields`, and recursive nested `expansions`. Lets callers pick which FKs to follow and which of their columns to bring back, instead of the all-or-nothing `expansionLevel` knob.
- Rejects the silent-data-loss combo of `expansionLevel > 0` + a non-empty `expansions` array (the backend drops the array otherwise — easy to miss).
- JSDoc: `queryRecordsById` gets two new `@example` blocks (auto-expand vs explicit); `EntityQueryFilterGroup` gets a mixed-AND/OR example using `continueLogicalOperator`.

No behavior change for callers who don't pass the new option.

## Test plan
- [x] `npm run lint` — 0 warnings
- [x] `npm run test:unit` — 171/171 entities unit tests pass
- [ ] Live API: query an entity with `expansions: [{ expandedField, alias, selectedFields, expansions: [...] }]` and confirm the response has the aliased keys + projected fields only.
- [ ] Live API: pass both `expansionLevel: 2` and `expansions: [...]` and confirm the SDK throws `ValidationError` before the request leaves.

### DIFF
--- a/src/models/data-fabric/entities.models.ts
+++ b/src/models/data-fabric/entities.models.ts
@@ -354,10 +354,11 @@ export interface EntityServiceModel {
   deleteRecordById(entityId: string, recordId: string): Promise<void>;
 
   /**
-   * Queries entity records with filters, sorting, aggregates, and SDK-managed pagination
+   * Queries entity records with filters, sorting, aggregates, expansion, and SDK-managed pagination
    *
    * @param id - UUID of the entity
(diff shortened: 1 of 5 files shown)


vnaren23:
Closing stale PR

PR #477, opened 2026-05-29
[PLT-104923] feat(coded-apps): inject all uipath.json keys as meta tags in local dev

Raina451:
- Makes the coded-apps plugin **generic**: any key in `uipath.json` is now injected as a `<meta name="uipath:<kebab-case>">` tag into the local dev `index.html`
- No plugin changes needed when new config keys are added (e.g. `folderKey` → `uipath:folder-key`)
- Well-known SDK keys (`clientId`, `scope`, `orgName`, `tenantName`, `baseUrl`, `redirectUri`) still use their predefined suffix mapping
- Unknown keys are converted from camelCase to kebab-case automatically via a new `camelToKebab` utility

deployment though will only inject `clientId`, `scope`, `orgName`, `tenantName`, `baseUrl`, `redirectUri`, `folderkey` , behaviour documented for plugin code.

### DIFF
--- a/packages/plugins/coded-apps/src/constants.ts
+++ b/packages/plugins/coded-apps/src/constants.ts
@@ -26,10 +26,35 @@ export const CONFIG_FILE_NAME = 'uipath.json'
 export const PACKAGE_JSON_FILE_NAME = 'package.json'
 
 /**
- * Mapping from config keys to meta tag names
+ * Prefix for all injected meta tags: `<meta name="uipath:..." ...>`
  */
 export const META_TAG_PREFIX = 'uipath'
 
+/**
+ * Well-known SDK configuration keys and their meta tag suffixes.
+ *
+ * These are the keys the UiPath SDK reads at runtime. The plugin uses this
+ * mapping for exact suffix names; any key NOT listed here is still injected
+ * — its suffix is derived by converting camelCase to kebab-case automatically
+ * (e.g. `folderKey` → `uipath:folder-key`).
+ *
+ * ----- Deployment-time meta tags (server-side injection) -----
+ * At deployment, the Apps Service (process-coded-app-package.ts) injects
+ * the following meta tags into the published index.html:
+ *
+ *   uipath:client-id      — OAuth client ID (from uipath.json in the .nupkg)
+ *   uipath:scope           — OAuth scopes (from uipath.json in the .nupkg)
+ *   uipath:org-name        — Organization ID
+ *   uipath:tenant-name     — Tenant ID
+ *   uipath:base-url        — Portal base URL
+ *   uipath:redirect-uri    — OAuth redirect URI (computed from routing name)
+ *   uipath:cdn-base        — CDN base path (computed: /{systemName}/{folderKey}/{deployVersion}/)
+ *   uipath:app-base        — App base path (computed from routing name or legacy path)
+ *   uipath:folder-key      — Folder identifier where the app is deployed
+ *
+ * The plugin mirrors this for local dev: anything in uipath.json is injected
+ * so the SDK can read the same meta tags locally as it would in production.
+ */
 export const UIPATH_META_TAGS: Record<ValidConfigKey, string> = {
   clientId: 'client-id',
   scope: 'scope',
@@ -64,7 +89,6 @@ export const MESSAGES = {
   // Validation errors
   INVALID_KEY_FORMAT: (key: string, suggestion: string) =>
     `❌ Invalid key "${key}" - use camelCase: "${suggestion}"`,
-  UNKNOWN_KEY: (key: string) => `⚠️  Unknown key "${key}" will be ignored by the SDK`,
 
   // Validation warnings
   MISSING_DEV_KEYS: (keys: string[]) =>
--- a/packages/plugins/coded-apps/src/constants.ts
+++ b/packages/plugins/coded-apps/src/constants.ts
@@ -44,8 +44,8 @@ export const META_TAG_PREFIX = 'uipath'
  *
  *   uipath:client-id      — OAuth client ID (from uipath.json in the .nupkg)
  *   uipath:scope           — OAuth scopes (from uipath.json in the .nupkg)
- *   uipath:org-name        — Organization ID
- *   uipath:tenant-name     — Tenant ID
+ *   uipath:org-name        — Organization name
+ *   uipath:tenant-name     — Tenant name
  *   uipath:base-url        — Portal base URL
  *   uipath:redirect-uri    — OAuth redirect URI (computed from routing name)
  *   uipath:cdn-base        — CDN base path (computed: /{systemName}/{folderKey}/{deployVersion}/)
--- a/packages/plugins/coded-apps/src/core/config-reader.ts
+++ b/packages/plugins/coded-apps/src/core/config-reader.ts
@@ -135,7 +135,6 @@ function validateKeyFormats(
       // Known SDK key in wrong format (e.g. client_id → clientId)
       collector.addError(MESSAGES.INVALID_KEY_FORMAT(key, matchingKey))
     }
-    // Unknown keys are allowed — they are injected as uipath:<kebab-case> meta tags
   }
 }
 
--- a/packages/plugins/coded-apps/src/core/config-reader.ts
+++ b/packages/plugins/coded-apps/src/core/config-reader.ts
@@ -132,10 +132,10 @@ function validateKeyFormats(
     const matchingKey = findMatchingKey(key)
 
     if (matchingKey && matchingKey !== key) {
+      // Known SDK key in wrong format (e.g. client_id → clientId)
       collector.addError(MESSAGES.INVALID_KEY_FORMAT(key, matchingKey))
-    } else if (!matchingKey) {
-      collector.addWarning(MESSAGES.UNKNOWN_KEY(key))
     }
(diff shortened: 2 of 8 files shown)


deepeshrai-tech (approved):

packages/plugins/coded-apps/tests/meta-tag-generator.test.ts
```diff
@@ -47,22 +47,47 @@ describe('generateMetaTagsForVite', () => {
       'uipath:tenant-name',
     ])
   })
+
+  it('injects arbitrary keys from uipath.json as kebab-case meta tags', () => {
+    const config: UiPathConfig = {
+      ...fullConfig,
+      folderKey: 'my-folder-key-123',
+      cdnBase: 'https://cdn.example.com',
+    }
```
ruudandriessen, line 56: mmm I'm not sure if I'd just randomly go do this. Was there a specific reason why we'd do this? Generally this would only be a thing in dev and when we deploy it won't have these keys right? So it seems a bit shady and may lead to dev/deploy diffs

Raina451: > only be a thing in dev and when we deploy it won't have these keys right?

yes that's right. 

> So it seems a bit shady and may lead to dev/deploy diffs

I completely agree, same thing i pointed out to @vnaren23 ,but we decided to make it generic to be forward compatible, we dont want to keep updating the plugin.

vnaren23: > we dont want to keep updating the plugin. 

This was my reasoning primarily because it'll become hard to coordinate the change across rings without a FF (since this is client side library)

ruudandriessen: I don't really think "keep updating the plugin" is a good argument - I think we should make sure dev+prod is more closely aligned. I'd much rather push a new update to the vite plugin to mirror what we have on prod then doing stuff like secretly passing through vars that won't exist on deploy.

ruudandriessen: Because the plugin only does this in dev right? Everything in production is done by whatever service you have running?

vnaren23: Nothing is secret per se. We define the uipath.json contract which is public and this will keep mirroring/be in sync with what server passes to the app.

packages/plugins/coded-apps/src/core/config-reader.ts
```diff
@@ -132,10 +132,10 @@ function validateKeyFormats(
     const matchingKey = findMatchingKey(key)
 
     if (matchingKey && matchingKey !== key) {
+      // Known SDK key in wrong format (e.g. client_id → clientId)
       collector.addError(MESSAGES.INVALID_KEY_FORMAT(key, matchingKey))
-    } else if (!matchingKey) {
-      collector.addWarning(MESSAGES.UNKNOWN_KEY(key))
     }
+    // Unknown keys are allowed — they are injected as uipath:<kebab-case> meta tags
```
vnaren23: Do we need this comment?

packages/plugins/coded-apps/src/core/config-reader.ts
```diff
@@ -132,10 +132,10 @@ function validateKeyFormats(
     const matchingKey = findMatchingKey(key)
 
     if (matchingKey && matchingKey !== key) {
+      // Known SDK key in wrong format (e.g. client_id → clientId)
       collector.addError(MESSAGES.INVALID_KEY_FORMAT(key, matchingKey))
-    } else if (!matchingKey) {
-      collector.addWarning(MESSAGES.UNKNOWN_KEY(key))
```
vnaren23, line 137: Where was this warning coming?

packages/plugins/coded-apps/src/core/meta-tag-generator.ts
```diff
@@ -1,45 +1,55 @@
 import { META_TAG_PREFIX, UIPATH_META_TAGS } from '../constants'
-import { isValidKey } from './utils'
+import { isValidKey, camelToKebab } from './utils'
 import type { UiPathConfig, ViteMetaTag } from '../types'
 
 /**
  * Generate meta tags in Vite format (for Vite's transformIndexHtml hook).
+ *
+ * Every key in the config is injected as a `<meta name="uipath:..." content="...">` tag.
```
vnaren23: Lets mention on comments that this is meant to be used only for SDK keys.

packages/plugins/coded-apps/src/core/meta-tag-generator.ts
```diff
@@ -1,45 +1,55 @@
...
+ * Unknown keys are converted from camelCase to kebab-case (e.g. folderKey → uipath:folder-key).
+ */
 function getMetaTagName(configKey: string): string {
-  if (!isValidKey(configKey)) return ''
-  const suffix = UIPATH_META_TAGS[configKey]
-  return suffix ? `${META_TAG_PREFIX}:${suffix}` : ''
+  if (isValidKey(configKey)) {
+    return `${META_TAG_PREFIX}:${UIPATH_META_TAGS[configKey]}`
+  }
+  return `${META_TAG_PREFIX}:${camelToKebab(configKey)}`
```
vnaren23: if else or ternary operator please. This is neither.

packages/plugins/coded-apps/src/core/meta-tag-generator.ts
```diff
@@ -1,45 +1,55 @@
...
 export function generateMetaTagsHtml(config: UiPathConfig): string {
   const lines: string[] = []
 
   for (const [key, value] of Object.entries(config)) {
-    const metaName = getMetaTagName(key)
-    if (metaName && value) {
-      lines.push(`<meta name="${metaName}" content="${value}">`)
-    }
+    if (!value) continue
+    lines.push(`<meta name="${getMetaTagName(key)}" content="${value}">`)
```
vnaren23: avoid using `continue`. Below will still work right?

````
if (value)
    lines.push(`<meta name="${getMetaTagName(key)}" content="${value}">`)
````

File: packages/plugins/coded-apps/src/constants.ts [hunk 7]
```diff
@@ -26,10 +26,35 @@ export const CONFIG_FILE_NAME = 'uipath.json'
...
+ * (e.g. `folderKey` → `uipath:folder-key`).
+ *
+ * ----- Deployment-time meta tags (server-side injection) -----
+ * At deployment, the Apps Service (process-coded-app-package.ts) injects
+ * the following meta tags into the published index.html:
+ *
+ *   uipath:client-id      — OAuth client ID (from uipath.json in the .nupkg)
+ *   uipath:scope           — OAuth scopes (from uipath.json in the .nupkg)
+ *   uipath:org-name        — Organization ID
+ *   uipath:tenant-name     — Tenant ID
```
vnaren23: Since we know we are making this change, should we document these are org names and tenant names. And maybe also add explicit org id and tenant id line items as well

Raina451: that change still still in pr review phase. but I'll document here now

packages/plugins/coded-apps/src/core/config-reader.ts
Raina451: It was a console warning during config validation , if one put a key in  uipath.json that wasn't in the SDK config list (like folderKey), it'd print warning `Unknown key "folderKey" will be ignored by the SDK`. 
Removed it since unknown keys are now intentionally passed through.

vnaren23: unknown keys are only passed during development time but not at deploy time right?

packages/plugins/coded-apps/src/core/config-reader.ts
Raina451: removed

packages/plugins/coded-apps/src/core/config-reader.ts
Raina451: right

packages/plugins/coded-apps/src/constants.ts
Raina451: updated this:

> should we document these are org names and tenant name

orgid and tenantid comment will better we update / add when changes reaches prod.

packages/plugins/coded-apps/src/core/config-reader.ts
vnaren23: Then we can still have the warning and tweak the message to mention the same. ie key will not be injected when app is hosted on UiPath.

Raina451: Then again i have to maintain a list of known keys which are supported  to:
- identify non supported keys to throw warning.

vnaren23: We already have the list [here ](https://github.com/UiPath/uipath-typescript/blob/8153f517da43f3a1a38c01e052f56537dbacff84/packages/plugins/coded-apps/src/types.ts#L16)isnt it?

Raina451: thats only sdk config keys, i havent extend it beyond that, for eg: folder key.

vnaren23 (approved):

PR #488, opened 2026-06-03
feat(tasks): add createQuickForm for QuickForm HITL tasks[ACTN-11300]

chetanyauipath:
Adds QuickForm HITL task support to `tasks.create`. Existing callers keep working — `tasks.create({ title, data, priority }, folderId)` still posts a `TaskType.External` task.

`TaskCreateOptions` is a **single flat interface** (extends `TaskCreateBaseOptions`: title, data, priority, labels → converted to tags, common to every type). QuickForm fields — `type`, `taskSchemaKey`, `schema`, `isActionableMessageEnabled`, `actionableMessageMetaData`, `creatorJobKey` — are optional at the type level; **`taskSchemaKey` is required only for QuickForm**, enforced by a runtime `ValidationError` guard in `create()` (mirrors the backend's 400). For a QuickForm task pass `type: TaskType.QuickForm` with `taskSchemaKey` + inline `schema`. Both paths POST to `/orchestrator_/tasks/GenericTasks/CreateTask` and return a Task.

**Schema semantics:** the backend keys off the top-level `taskSchemaKey` (the `id` inside the schema body is ignored). On first use the `schema` is registered under `taskSchemaKey`; on later calls with the same key it is **reused, not updated** (keys are unique per tenant). The SDK sends `taskSchemaKey` + inline `schema` in a single call (the backend's create-on-first-use path).

**Errors** from the backend (e.g. invalid/oversized schema, assignment/folder failures) propagate unchanged as SDK errors.

Modeled as a flat interface rather than a discriminated union so callers can construct the options object before calling `create()` and consumers can extend the type; the QuickForm `taskSchemaKey` requirement is enforced at runtime.

### DIFF
--- a/src/models/action-center/tasks.types.ts
+++ b/src/models/action-center/tasks.types.ts
@@ -43,7 +43,9 @@ export enum TaskType {
   /** A document classification task for categorizing documents */
   DocumentClassification = 'DocumentClassificationTask',
   /** A data labeling task for annotating training data */
-  DataLabeling = 'DataLabelingTask'
+  DataLabeling = 'DataLabelingTask',
+  /** A schema-first HITL task rendered by FormLib in Action Center via a registered TaskSchema */
+  QuickForm = 'QuickFormTask'
 }
 
 export enum TaskPriority {
@@ -162,11 +164,49 @@ export interface TaskBaseResponse {
   lastModifiedTime: string | null;
 }
 
-export interface TaskCreateOptions {
+/**
+ * Options for creating an External task. `type` is the discriminator for the
+ * `TaskCreateOptions` union; omit it (or set it to `TaskType.External`) for
+ * the default external-task shape.
+ */
+export interface TaskCreateExternalOptions {
+  type?: TaskType.External;
   title: string;
   data?: Record<string, unknown>;
   priority?: TaskPriority;
 }
+
+/**
+ * Options for creating a QuickForm task — a schema-first HITL task rendered by
+ * FormLib in Action Center.
+ *
+ * Both `taskSchemaKey` and `schema` are sent on every call: Orchestrator upserts
+ * the schema by `taskSchemaKey`, then creates the task in the same call.
+ */
+export interface TaskCreateQuickFormOptions {
+  type: TaskType.QuickForm;
+  title: string;
+  /** UUID key under which Orchestrator registers/looks up the schema in TaskSchemas */
+  taskSchemaKey: string;
+  /** Inline schema body. Sent on every call so Orchestrator can upsert it */
+  schema: Record<string, unknown>;
+  data?: Record<string, unknown>;
+  priority?: TaskPriority;
+  /** Converted to wire-shape `tags` on submit */
+  labels?: string[];
+  isActionableMessageEnabled?: boolean;
+  /** When null on QuickForm, Orchestrator derives it from the referenced TaskSchema */
+  actionableMessageMetaData?: Record<string, unknown>;
+  /** Identifies the job that triggered the schema creation (paired with `schema`) */
+  creatorJobKey?: string;
+}
+
+/**
+ * Discriminated union over `type`. `tasks.create` accepts either shape — the
+ * `type` field on `TaskCreateQuickFormOptions` narrows it; omitting `type`
+ * defaults to the external-task shape.
+ */
+export type TaskCreateOptions = TaskCreateExternalOptions | TaskCreateQuickFormOptions;
 
 export interface RawTaskCreateResponse extends TaskBaseResponse {
   waitJobState: JobState | null;
@@ -243,6 +283,7 @@ export type TaskCompleteOptions =
   | { type: TaskType.DataLabeling; data?: any; action?: string }
   | { type: TaskType.Form; data: any; action: string }
   | { type: TaskType.App; data: any; action: string }
+  | { type: TaskType.QuickForm; data: any; action: string }
 
 /**
  * Options for completing a task when called from the service
--- a/src/models/action-center/tasks.types.ts
+++ b/src/models/action-center/tasks.types.ts
@@ -43,7 +43,9 @@ export enum TaskType {
   /** A document classification task for categorizing documents */
   DocumentClassification = 'DocumentClassificationTask',
   /** A data labeling task for annotating training data */
-  DataLabeling = 'DataLabelingTask'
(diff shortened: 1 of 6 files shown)


src/services/action-center/tasks.ts
```diff
@@ -74,6 +75,69 @@ export class TaskService extends BaseService implements TaskServiceModel {
...
+   *       { id: "reject",  name: "Reject",  type: "string", isPrimary: false },
+   *     ],
+   *   },
+   *   data: { invoice: "INV-1234" },
+   *   priority: TaskPriority.Medium,
+   * }, 123);
+   * ```
+   */
+  @track('Tasks.CreateQuickForm')
+  async createQuickForm(options: TaskCreateQuickFormOptions, folderId: number): Promise<TaskCreateResponse> {
```
Sandeepan-Ghosh-0312: lets add this to above create method itself, take in type as a parameter

src/services/action-center/tasks.ts
```diff
@@ -37,35 +37,65 @@ import { BaseService } from '../base';
...
-    const externalTask = {
-      ...task,
-      type: TaskType.External //currently only external task is supported
-    };
-    
+
+    let payload: Record<string, unknown>;
+    if (options.type === TaskType.QuickForm) {
+      const { labels, ...rest } = options;
+      payload = labels === undefined ? rest : {
```
Sandeepan-Ghosh-0312: why are labels added only for quick forms?

src/services/action-center/tasks.ts
```diff
@@ -37,35 +37,65 @@ import { BaseService } from '../base';
...
    *   data: { key: "value" }
-   * }, 123); // folderId is required
+   * }, 123);
+   *
+   * // QuickForm task — both taskSchemaKey AND schema are sent on every call:
+   * // Orchestrator upserts the schema by taskSchemaKey, then creates the task.
+   * await tasks.create({
+   *   type: TaskType.QuickForm,
+   *   title: "Approve invoice INV-1234",
+   *   taskSchemaKey: "8e4f2a91-3c7e-4d2b-9b5c-1a6f8d3e2c91",
```
Sandeepan-Ghosh-0312: Please make taskSchemaKey as optional here, from sdk only the schema should be passed. It is confusing from a developer perspective to add both schema and schemaKey here

src/models/action-center/tasks.types.ts
```diff
@@ -162,12 +164,50 @@ export interface TaskBaseResponse {
...
+}
+
+/**
+ * Options for creating a QuickForm task — a schema-first HITL task rendered by
+ * FormLib in Action Center.
+ *
+ * Both `taskSchemaKey` and `schema` are sent on every call: Orchestrator upserts
+ * the schema by `taskSchemaKey`, then creates the task in the same call.
+ */
+export interface TaskCreateQuickFormOptions {
```
Sandeepan-Ghosh-0312: instead of creating separate options for each please create a base one which accepts the common fields, refer to models in tasks for the same

dushyant-uipath:
To be updated with Types support in a followup, when ready, allowing people to pass in 'UI compliant' schemas. Today, it allows all JSONs

dushyant-uipath (approved):

src/models/action-center/tasks.types.ts
chetanyauipath: Done — added TaskCreateBaseOptions with the common fields (title, data, priority, labels)

src/services/action-center/tasks.ts
chetanyauipath: TaskCreateRequest : BaseTaskCreateRequest exposes Tags for every task type. Moved labels into the base and apply labels→tags for both

src/services/action-center/tasks.ts
chetanyauipath: Per product decision, the SDK caller always creates and passes taskSchemaKey, mirroring StudioWeb where the client owns the schemaKey. Keeping it required is deliberate

src/services/action-center/tasks.ts
```diff
@@ -44,17 +44,18 @@ import { FolderScopedService } from '../folder-scoped';
  */
 export class TaskService extends FolderScopedService implements TaskServiceModel {
   @track('Tasks.Create')
-  async create(task: TaskCreateOptions, folderId: number): Promise<TaskCreateResponse> {
+  async create(options: TaskCreateOptions, folderId: number): Promise<TaskCreateResponse> {
```
dushyant-uipath: We might not be able to break a public contract - should continue calling it tasks

src/models/action-center/tasks.types.ts
```diff
@@ -162,12 +164,57 @@ export interface TaskBaseResponse {
...
+  /** Identifies the job that triggered the schema creation (paired with `schema`) */
+  creatorJobKey?: string;
+}
+
+/**
+ * Discriminated union over `type`. `tasks.create` accepts either shape — the
+ * `type` field on `TaskCreateQuickFormOptions` narrows it; omitting `type`
+ * defaults to the external-task shape.
+ */
+export type TaskCreateOptions = TaskCreateExternalOptions | TaskCreateQuickFormOptions;
```
ninja-shreyash, line 216: this breaks for callers who construct the options object before calling the tasks.create() method. A single method signature cannot decide the type specific fields if user constructs options object beforehand

ninja-shreyash, line 216: Also union types cant be extended by consumers

ninja-shreyash, line 216: this also breaks an existing integration test in this repo: tests/integration/shared/action-center/tasks.integration.test.ts:85, which uses exactly the hoisted-object pattern above. CI doesn't surface it because tsconfig.json excludes tests/.

ninja-shreyash (changes requested):
Pls address claude comments as well

src/models/action-center/tasks.types.ts
chetanyauipath: Flattened TaskCreateOptions into a single interface, dropped the External | QuickForm union. type/taskSchemaKey/schema are optional at the type level, so options objects built before the create() call type-check and the interface is extensible.

chetanyauipath: Fixed.

ninja-shreyash (approved):

PR #489, opened 2026-06-03
fix(data-fabric): align entity field type emission with UI's narrower FieldType enum

AditiGoyalUipath:
## Summary

The Manage Entity UI accepts a narrower set of SQL types than the backend (FE doc §3.1, §3.4). When the SDK emitted `INT`/`BIGINT`/`FLOAT`/`REAL`/`DATETIME2` — or omitted the field-level `type` discriminator entirely — the UI silently rewrote those fields to `NVARCHAR` on load. Saving again then failed with **"Field type cannot be changed"** because `SqlType.Name` is immutable on update (backend doc §6.4), even when the user made no changes.

## Changes

**1. Emit `type` (UI logical type) on every outbound field**
- New `apiTypeName` column in `EntitySchemaFieldTypeMap` (`"text"`, `"number"`, `"dateTime"`, `"relationship"`, etc.)
- `buildSchemaFieldPayload()` now sets `type: mapping.apiTypeName` on the wire payload
- `FieldMetaData.type` carries it back on read so re-upserts preserve it

**2. Collapse types the UI rewrites**
- `INTEGER`, `BIG_INTEGER`, `FLOAT`, `DOUBLE` → `DECIMAL` (was `INT`/`BIGINT`/`FLOAT`/`REAL`)
- `DATETIME` → `DATETIMEOFFSET` (was `DATETIME2`)
- `INTEGER`/`BIG_INTEGER` set `decimalPrecision: 0` (new `INTEGER_DECIMAL_PRECISION` constant) to preserve integer semantics
- `FLOAT`/`DOUBLE` retain the user-overridable precision (default 2)

**3. Relax FILE-field create**
- FILE no longer requires `referenceEntityId`/`referenceFieldId` — server auto-provisions the EntityAttachment FK (backend doc §5.10)
- FILE no longer emits `isForeignKey`/`referenceType`/`referenceEntity`/`referenceField` (those are RELATIONSHIP-only)
- Validation guard tightened from `(isRelationship || isFile)` to `isRelationship` only

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run test:unit` — passes (unit assertions updated for new SQL types, precision, and `type` field)
- [ ] Integration tests cover:
  - FILE field create without `referenceEntityId`/`referenceFieldId`
  - Full create → record-insert → uploadAttachment round-trip
- [ ] Verify in Manage Entity UI: create an entity with INTEGER/FLOAT/DATETIME fields via SDK, open in UI, save without changes — should succeed (previously failed with "Field type cannot be changed")

### DIFF
--- a/src/models/data-fabric/entities.constants.ts
+++ b/src/models/data-fabric/entities.constants.ts
@@ -1,4 +1,4 @@
-import { EntityFieldDataType, FieldDisplayType } from "./entities.types";
+import { EntityApiFieldType, EntityFieldDataType, FieldDisplayType } from "./entities.types";
 import {
   EntitySchemaFieldMapping,
   SqlFieldType,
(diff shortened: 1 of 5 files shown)


vnaren23 (changes requested):
> The Manage Entity UI accepts a narrower set of SQL types than the backend (FE doc §3.1, §3.4). When the SDK emitted INT/BIGINT/FLOAT/REAL/DATETIME2 — or omitted the field-level type discriminator entirely 

Is the backend meant to accept such wider set of SQL types? Forcing SDK changes just to confer to lack of support from DF UI seems incorrect. If anything we should add these support into DF UI rather than making SDK changes.

vnaren23:
@AditiGoyalUipath are these still needed? If yes please resolve the conflict, else close the PR

PR #492, opened 2026-06-03
feat(agents): Agents service foundation + getAll (1/5) [PLT-103787]

ashishupadhyay88:
## Part 1 of 5 — splitting #438

#438 added the full Agents service (14 methods) in one PR, too large to review. This stack breaks it into reviewable pieces. **This PR is the foundation** — it establishes the service module and the foundational `getAll` method plus the shared filter scaffolding, so PRs 2–5 only *append* per-method code.

> **Note:** `getNames()` is **not** part of this stack — it stays on `feat/sdk-plt-103787` and won't be merged separately. `getAll` is the foundation method instead.

| PR | Theme | Methods |
|----|-------|---------|
| **1 (this)** | **Foundation + first method** | `getAll` |
| 2 | Timelines | `getErrorsTimeline`, `getConsumptionTimeline`, `getLatencyTimeline` |
| 3 | Summaries & distributions | `getSummary`, `getUnitConsumptionSummary`, `getTopConsumingAgents`, `getTopErroredAgents`, `getIncidentDistribution` |
| 4 | Paginated incidents | `getIncidents` |
| 5 | Trace controller | `getTraceErrorsTimeline`, `getTraceLatencyTimeline`, `getTraceUnitConsumption` |

PRs 2–5 branch off this one and rebase on `main` before each merge.

## What's in this PR (19 files, +656 / −5)

- **Service module + subpath**: `@uipath/uipath-typescript/agents` (`package.json` exports, `rollup` entry, barrels)
- **Endpoint**: `AGENTS_ENDPOINTS.GET_AGENTS` (`insightsrtm_/Agents/agents`, POST)
- **`getAll()`**: paginated/non-paginated agent list with consumption + health metadata; conditional `PaginatedResponse<AgentListItem>` / `NonPaginatedResponse<AgentListItem>`. Routes through the shared `PaginationHelpers.getAll()` using the `zeroBased` flag (`convertToSkip: false`) — the API is a 0-based page-number endpoint, and the helper handles the SDK 1-based ↔ API 0-based conversion + cursor navigation.
- **Shared pagination (nested envelopes)**: item extraction in `base.ts` (`createPaginatedResponseFromResponse`) and `helpers.ts` (`getAllNonPaginated`) now uses `resolveNestedField`, so a dotted `itemsField` like `data.agents` (and `totalCountField` `pagination.totalCount`) resolves. Backward-compatible — flat field names (`value`, `items`) short-circuit to identical behavior.
[...]

PR #503, opened 2026-06-10
feat(agents): getErrors — paginated agent errors (2/5) [PLT-103787]

ashishupadhyay88:
## Agents split — `getErrors`

Second PR of the Agents service split. The foundation (`getAll` + shared scaffolding) merged in **#492**; this adds **`getErrors`** on top of `main`.

`getErrors` returns a paginated/non-paginated list of agent **errors** (error-classes observed for agents) over a time window. It mirrors the merged `getAll` pattern exactly — `Date` params, delegates to `PaginationHelpers.getAll()` with 0-based `pageNumber` (`convertToSkip: false`, `zeroBased: true`).

> Method and public types use the **errors** vocabulary. The backend endpoint (`/Agents/incidents`) and internal constants (`GET_INCIDENTS`, `AGENTS_INCIDENTS_PAGINATION`) keep their original names — the API path is owned by the service, not renamed here.

## What's in this PR

- **Endpoint**: `AGENTS_ENDPOINTS.GET_INCIDENTS` (`insightsrtm_/Agents/incidents`)
- **Pagination constants**: `AGENTS_INCIDENTS_PAGINATION` — items are the `data` array directly, count under `pagination.totalCount` (reuses `AGENTS_OFFSET_PARAMS`)
- **Types**: `AgentError`, `AgentJobInfo`, `AgentErrorSortColumn`, `AgentErrorOrderBy`, `AgentErrorsOptions`
- **`getErrors()`** service method + `AgentServiceModel` entry (synced JSDoc); telemetry label `Agents.GetErrors`
- **Tests**: unit (8) mirroring `getAll` — incl. the filter-passthrough/OData-prefix-exclusion test guarding `excludeFromPrefix`; integration under the existing `describe.skip` suite (insightsrtm_ rejects PAT → needs OAuth)
- **OAuth scope** + pagination docs rows

## Design note

The API returns a `totalErrorCount` aggregate alongside the errors. It's **intentionally not surfaced** — consistent with the merged `getAll`, which drops its aggregate totals from the paginated list (`PaginationHelpers.getAll()` returns a plain `PaginatedResponse<T>`; aggregate totals belong with the summary methods in a later PR). Happy to revisit if reviewers want it.

## Verification

- ✅ `typecheck` clean · `lint` 0/0 · `test:unit` pass · `build` OK
- ✅ **E2E passed** — validated live via the OAuth browser flow against `alpha.uipath.com`. The endpoint returned real data (`totalCount=62`), pagination shape correct (`hasNextPage`, `currentPage`), all fields camelCase, nested `firstSeenJob.jobKey` correctly cased, and no PascalCase leakage.

OAuth scope: `Insights.RealTimeData Insights OR.Folders.Read`. Refs PLT-103787.


src/services/agents/agents.ts
```diff
@@ -105,4 +108,86 @@ export class AgentService extends BaseService implements AgentServiceModel {
...
+   * if (page.hasNextPage && page.nextCursor) {
+   *   const next = await agents.getIncidents(
+   *     new Date('2025-05-01T00:00:00Z'),
+   *     new Date('2026-05-14T00:00:00Z'),
+   *     { cursor: page.nextCursor },
+   *   );
+   * }
+   * ```
+   */
+  @track('Agents.GetIncidents')
```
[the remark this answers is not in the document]
ashishupadhyay88: taken care

docs/oauth-scopes.md
```diff
@@ -244,3 +244,4 @@ The `ConversationalAgents` scope is required for real-time WebSocket sessions (`
 | Method | OAuth Scope |
 |--------|-------------|
 | `getAll()` | `Insights.RealTimeData Insights OR.Folders.Read` |
+| `getIncidents()` | `Insights.RealTimeData Insights OR.Folders.Read` |
```
vnaren23: Why name it incidents? Seems very engineering driven ie leaking from our API naming. For users and customers, why not just name it `getErrors`?

ashishupadhyay88: Hey @vnaren23 , I understand your point but the fact that this API returns `Type=Error` today but the BE schema is built on extending this to may be `warning`, `Anomaly` etc. so keeping `getErrors()` might be irrelevant in future.

vnaren23: They are just different types of errors right?

ashishupadhyay88: yes @vnaren23 it is, but it may extend to Policy / Escalation in future. which I don't think an error type,`Incident` is intentionally a superset of {errors, policies, escalations}. `Error` is only today's populated subset.

vnaren23: Can you share me an example of how errors, policies, escalations are shown to the user now on UI?

ashishupadhyay88: currently it only pulls with query type=error,  so others fields doesn't exist today. every row in the response today is type=error but as said the BE has defined the schema in such a way that it may extend beyond error.

I can change the name to getErrors(), will check with insights if they extend this api for other things like policies, escalations in future. naming getErrors() today is absolutely correct and makes sense.

vnaren23: Lets change to getErrors for now. Parallelly lets start with a thread with both agents (tag Scott Florentino) and insights team.

ashishupadhyay88: taken care in the PR and starting a thread with insights.

vnaren23 (approved):

deepeshrai-tech (approved):

PR #508, opened 2026-06-11
feat: add client-side tool support [JAR-9962]

norman-le:
### Summary
- Adds support for client-side tools — tools that execute on the client rather than server-side (CAS PR here: https://github.com/UiPath/AgentInterfaces/pull/949)
- Clients can declare supported tools on exchange start, detect client-side tool calls via `isClientSideTool`, and respond to the `executingToolCall` signal to begin local execution

### What's new

- **`ExchangeStartEvent.clientSideTools`** — declare which client-side tools the client supports so the server can filter accordingly. When omitted, all agent-defined client-side tools are assumed available.
- **`ToolCallStartEvent.isClientSideTool`** — server-set flag indicating the tool should be executed client-side
- **`ToolCallStartEvent.outputSchema`** — JSON schema describing the expected result shape, used by the client to render a result form if needed
- **`ExecutingToolCallEvent`** — new event signaling the client to begin execution (fires after optional confirmation)
- **`ToolCallStream.onExecutingToolCall()`** — register a handler for the executing signal

### Usage

```typescript
// Declare client-side tools when starting an exchange
const exchange = session.startExchange({
  clientSideTools: [{ name: 'launch-calculator' }]
});

// Handle client-side tool execution
message.onToolCallStart((toolCall) => {
  const { isClientSideTool, toolName } = toolCall.startEvent;
  if (!isClientSideTool) return;

  toolCall.onExecutingToolCall(async (event) => {
    const result = await runLocalProcess(toolName, event.input);
    toolCall.sendToolCallEnd({ output: result });
  });
});
```

### Tests
- Dispatch, handler registration/unregistration, event buffering when paused, ID filtering

### DIFF
--- a/src/models/conversational-agent/conversations/types/events/protocol.types.ts
+++ b/src/models/conversational-agent/conversations/types/events/protocol.types.ts
@@ -109,6 +109,16 @@ export interface SessionEndingEvent {
 }
 
 /**
+ * A client-side tool that the client supports, declared during exchange start
+ * so the server knows which tools to route client-side.
+ */
+export interface ClientSideTool {
+  name: string;
+  inputSchema?: JSONValue;
+  outputSchema?: JSONValue;
+}
+
+/**
  * Signals the start of an exchange of messages within a conversation.
  */
 export interface ExchangeStartEvent {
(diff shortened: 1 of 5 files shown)


vnaren23 (approved):

src/models/conversational-agent/conversations/types/events/protocol.types.ts
```diff
@@ -414,6 +460,21 @@ export interface ToolCallEndEvent {
   metaData?: MetaData;
 }
 
+/**
+ * Signals to the client that the tool is about to be executed. Emitted in all scenarios
+ * (server-side and client-side tools). For client-side tools, the client should begin
+ * executing its registered handler upon receiving this event.
+ * @internal
```
maxduu, line 467: Nice, thanks for the detailed comment!

maxduu (approved):

PR #510, opened 2026-06-11
feat: allow to emit end exchange event from client for stopping response

norman-le:
Adds tests demonstrating client-side exchange termination via `sendExchangeEnd()` on `ExchangeStream`. This validates that the existing WebSocket event path correctly supports the stop-response flow added in [UiPath/AgentInterfaces#1018](https://github.com/UiPath/AgentInterfaces/pull/1018), where CAS now handles client `endExchange` events by ending the exchange in DB and killing the running Orchestrator job.

**What changed:**
- Added 4 unit tests covering client-initiated stop scenarios:
  - Stopping mid-stream while assistant message is being chunked
  - Blocking further message operations after stop
  - `onExchangeEnd` handler firing on client-initiated stop (echo mode)
  - Stopping during an in-progress tool call

**Why no REST API `end()` method:**
The WebSocket `sendExchangeEnd()` is the consistent pattern — all other exchange lifecycle operations (start, message, tool calls) use the WebSocket `ConversationEvent` protocol, not REST. The CAS PR wired up the server-side handler for client `endExchange` events, making the existing SDK method fully functional for stop-response without needing a new HTTP endpoint.


src/models/conversational-agent/conversations/exchanges.models.ts
```diff
@@ -124,6 +125,24 @@ export interface ExchangeServiceModel {
     exchangeId: string,
     options: CreateFeedbackOptions
   ): Promise<FeedbackCreateResponse>;
+
+  /**
+   * Ends an exchange, setting the endedAt timestamp and stopping any running agent job
+   *
+   * @internal
```
vnaren23: Since we are adding the scopes on OAuth lets change this to `@experimental`

vnaren23 (approved):

src/models/conversational-agent/conversations/exchanges.models.ts
```diff
@@ -124,6 +125,24 @@ export interface ExchangeServiceModel {
     exchangeId: string,
     options: CreateFeedbackOptions
   ): Promise<FeedbackCreateResponse>;
+
+  /**
+   * Ends an exchange, setting the endedTime timestamp and stopping any running agent job
+   *
+   * @experimental
```
maxduu: I thought we do `@internal` rather than `@experimental`?
EDIT: Oh I see, both are okay according to Naren's message - are we doing `@experimental` here since it would be deployed to prod soon but still in a "preview" phase

norman-le: Naren suggested experimental since it touches OAuth scope https://github.com/UiPath/uipath-typescript/pull/510#discussion_r3425918871

maxduu:
Hm, I thought we were going with the event-through-websocket approach? It seems like this is using the API rather than the end-event?

norman-le:
> Hm, I thought we were going with the event-through-websocket approach? It seems like this is using the API rather than the end-event?

@maxduu 
Hm that's a good point. I was following the file since it had REST API for all the operations, but those aren't really session operations (e.g. getById, getAll, feedback, etc.). They both should technically work , but for consistency I'll change it to only use the webSocket approach (like start/end exchanges, send messages, chunks, tool calls)                    

Edit: Actually it seems like endExchange was already set up through the websocket - it was just ignored by CAS, so we shouldnt need to add anything here (it's just handled in CAS now instead of a no-op)

maxduu (approved):

maxduu (approved):

PR #500, opened 2026-06-09
feat(notifications): add Notifications and Subscriptions services

sarthak688:
Onboards the user-facing Notification platform as a new modular subpath `@uipath/uipath-typescript/notifications` with two services: `Notifications` (inbox) and `Subscriptions` (preferences). 15 methods total, no folder context, no method binding. Spec source: [Design Document — Notification Service CLI Tool & SDK Integration for Coding Agents](https://uipath.atlassian.net/wiki/spaces/BPA/pages/90542604630/).

### `Notifications` (6 methods)

| Layer | Method | Signature |
|-------|--------|-----------|
| Service | `notifications.getAll()` | `getAll<T extends NotificationGetAllOptions>(tenantId: string, options?: T): Promise<T extends HasPaginationOptions<T> ? PaginatedResponse<NotificationGetResponse> : NonPaginatedResponse<NotificationGetResponse>>` |
| Service | `notifications.markRead()` | `markRead(tenantId: string, notificationIds: string[]): Promise<NotificationUpdateReadResponse>` |
| Service | `notifications.markUnread()` | `markUnread(tenantId: string, notificationIds: string[]): Promise<NotificationUpdateReadResponse>` |
| Service | `notifications.markAllRead()` | `markAllRead(tenantId: string): Promise<NotificationMarkAllReadResponse>` |
| Service | `notifications.deleteNotifications()` | `deleteNotifications(tenantId: string, notificationIds: string[]): Promise<NotificationDeleteResponse>` |
| Service | `notifications.deleteAll()` | `deleteAll(tenantId: string): Promise<NotificationDeleteAllResponse>` |

### `Subscriptions` (9 methods)

| Layer | Method | Signature |
|-------|--------|-----------|
| Service | `subscriptions.getAll()` | `getAll(tenantId: string, options?: SubscriptionGetAllOptions): Promise<SubscriptionGetResponse>` |
| Service | `subscriptions.getPublishers()` | `getPublishers(tenantId: string, options?: SubscriptionGetPublishersOptions): Promise<SubscriptionGetResponse>` |
| Service | `subscriptions.getSupportedChannels()` | `getSupportedChannels(tenantId: string): Promise<SubscriptionGetSupportedChannelsResponse>` |
| Service | `subscriptions.updateTopic()` | `updateTopic(tenantId: string, subscriptions: TopicSubscriptionUpdate[]): Promise<SubscriptionUpdateTopicResponse>` |
[...]

PR #512, opened 2026-06-12
feat(notifications): foundation + getAll (PR 1/7)

sarthak688:
First in a stack of 7 PRs onboarding the user-facing Notification platform as a new modular subpath `@uipath/uipath-typescript/notifications`. This PR lays the foundation and ships **one method** (`Notifications.getAll`). Each subsequent PR (stacked on top of this branch) adds a small group of related methods.

## Methods Added

| Service | `notifications.getAll()` | `getAll<T extends NotificationGetAllOptions>(tenantId: string, options?: T): Promise<T extends HasPaginationOptions<T> ? PaginatedResponse<NotificationGetResponse> : NonPaginatedResponse<NotificationGetResponse>>` |

## Endpoint Called

| `notifications.getAll()` | GET | `notificationservice_/notificationserviceapi/odata/v1/NotificationEntry` | `NotificationService` |

- `NotificationService` extends `BaseService` — UserContext auth, **no folder header**.
- `getAll()` supports OData pagination via `PaginationHelpers.getAll` (`$top`, `$skip`, `$count`) and OData query options (`filter`, `orderby`).
- `$select` is **not** exposed (the API returns HTTP 500 when `$select` is supplied — verified live).
- The method carries an `@track('Notifications.GetAll')` telemetry decorator.

## Critical design point #1: `tenantId` as a per-call argument

The Notification API does not include the tenant in the URL path. Instead, it identifies the acting tenant via the `X-UIPATH-Internal-TenantId` header, which expects a tenant **GUID** (not the configured `tenantName`). The SDK takes `tenantId` as the first positional argument on every NS method and forwards it into the header via `createHeaders()` on each call. This mirrors how folder-scoped Orchestrator services already require `folderId` from the caller.

## Critical design point #2: URL traversal for tenant-less routing

The notification service routes at the **organization** level — its URLs do NOT include a tenant segment, unlike most UiPath services. `ApiClient` always inserts `${orgName}/${tenantName}/${path}`, so the endpoint constants use a `../` prefix to collapse the tenant segment:

```ts
// src/utils/constants/endpoints/base.ts
export const NOTIFICATION_BASE = '../notificationservice_';
```

`${orgName}/${tenantName}/../notificationservice_/...` resolves via `new URL()` normalization to `${orgName}/notificationservice_/...`. The `NOTIFICATION_BASE` JSDoc documents this in detail. **Verified live:**

- ✅ `alpha.uipath.com/testmanagerdev/notificationservice_/...` → HTTP 200
- ❌ `alpha.uipath.com/testmanagerdev/TestManagerDevTenant/notificationservice_/...` → 302 `/portal_/unregistered`

## Example Usage

```typescript
import { Notifications } from '@uipath/uipath-typescript/notifications';

await sdk.initialize();

const notifications = new Notifications(sdk);
const tenantId = '<tenant-guid>';

// List unread, most recent first
const unread = await notifications.getAll(tenantId, {
  filter: 'isRead eq false',
  orderby: 'publishedOn desc',
  pageSize: 20,
});

// First page with pagination
const page1 = await notifications.getAll(tenantId, { pageSize: 20 });
if (page1.hasNextPage) {
  const page2 = await notifications.getAll(tenantId, { cursor: page1.nextCursor });
}
```

### Transform pipeline

`PaginationHelpers.getAll` → `transformFn: stripInternalNotificationFields` → returned to consumer.

The API already returns camelCase, so `pascalToCamelCaseKeys()` is **not** applied. There are no semantic field renames. The only transform is dropping internal/transport-layer fields.

### Field-strip (dropped from `NotificationGetResponse`)

| Dropped field | Reason |
|---|---|
| `entityOrgName`, `entityTenantName` | Internal routing metadata |
| `serviceRegistryName` | Internal infrastructure |
| `messageTemplateKey`, `messageVersion` | Internal templating |
| `publicationId` | Internal publish-event id |
| `correlationId` | Internal tracing |
| `partitionKey` | Storage detail (was `readOnly: true` in spec) |

## Sample SDK Response

<details>
<summary><code>notifications.getAll(tenantId, { pageSize: 1 })</code></summary>

```json
{
  "items": [
    {
      "id": "9a3b0db5-9b8b-44a7-9b6a-08de0a17b4e2",
      "message": "A new model is now available for your product.",
      "isRead": false,
      "publisherName": "LlmGateway",
      "publisherId": "adc041d6-3098-4766-ae45-08dd3ac994aa",
      "topicName": "New model available for product",
      "topicKeyName": "New.Model.Available",
      "topicId": "66e76b72-5ef1-4567-6a00-08dd3ac994ba",
      "userId": "9ABFE2B8-F30F-4917-9F7F-A91D28F7B23C",
      "userEmail": null,
      "tenantId": null,
      "priority": "Low",
      "category": "Info",
      "messageParam": null,
      "redirectionUrl": "https://alpha.uipath.com/.../portal_/admin/ai-trust-layer/llm-configurations",
      "publishedOn": 1780413050
    }
  ],
  "totalCount": 500,
  "hasNextPage": true,
  "supportsPageJump": true,
  "currentPage": 1
}
```

Note: `publishedOn` is **Unix epoch seconds** (not milliseconds). Internal fields (`entityOrgName`, `partitionKey`, `correlationId`, etc.) are stripped before the SDK consumer sees the response.

</details>

## Verification

| Check | Status |
|---|---|
| `npm run typecheck` | ✅ clean |
| `npm run lint` | ✅ 0 warnings, 0 errors |
| `npm run test:unit` | ✅ 64 files / 1801 tests (4 in notification suite) |
| `npm run build` | ✅ `dist/notifications/index.{mjs,cjs,d.ts}` produced |

Live API curl confirmed the endpoint works against `alpha/testmanagerdev`.

## Files

| Area | Files |
|---|---|
| Endpoint constants | `src/utils/constants/endpoints/notification.ts` (new), `src/utils/constants/endpoints/base.ts` (`NOTIFICATION_BASE`), `src/utils/constants/endpoints/index.ts` |
| Types | `src/models/notification/notifications.types.ts` |
| Internal types | `src/models/notification/notifications.internal-types.ts` |
| Models | `src/models/notification/notifications.models.ts`, `index.ts` |
| Services | `src/services/notification/notifications.ts`, `index.ts` |
| Build | `package.json` (subpath `./notifications`), `rollup.config.js` (`notifications` entry) |
| Unit tests | `tests/unit/services/notification/notifications.test.ts` (4 tests) |
| Integration tests | `tests/integration/shared/notification/notifications.integration.test.ts` (4 tests, `v1` mode only) |
| Integration wiring | `tests/integration/config/test-config.ts` (`notificationTenantId`), `tests/integration/config/unified-setup.ts`, `tests/.env.integration.example` (`NOTIFICATION_TEST_TENANT_ID`) |
| Test utils | `tests/utils/constants/notification.ts`, `tests/utils/mocks/notification.ts` (+ barrel updates) |
| Docs | `docs/oauth-scopes.md`, `docs/pagination.md`, `mkdocs.yml` |

## PR Stack

This PR is part of a 7-PR stack that incrementally builds out the Notifications + Subscriptions services. Each subsequent PR targets the previous PR's branch.

| # | Branch | Methods |
|---|--------|---------|
| 1 | `feat/notifications-sdk` *(this PR)* | `getAll` |
| 2 | `feat/notifications-mark-read` | `markRead`, `markUnread`, `markAllRead` |
| 3 | `feat/notifications-delete` | `deleteNotifications`, `deleteAll` |
| 4 | `feat/subscriptions-sdk` | Subscriptions foundation + `getAll`, `getPublishers`, `getSupportedChannels` |
| 5 | `feat/subscriptions-topic-updates` | `updateTopic`, `updateCategory` |
| 6 | `feat/subscriptions-publisher-updates` | `updatePublisher`, `updateTopicGroup` |
| 7 | `feat/subscriptions-mode-reset` | `updateMode`, `reset` |

Replaces #500.

### DIFF
--- a/src/models/notification/notifications.constants.ts
+++ b/src/models/notification/notifications.constants.ts
@@ -1,0 +1,9 @@
+/**
+ * Notification field mappings (API field name → SDK field name).
+ *
+ * Semantic renames only — case conversion is handled by `pascalToCamelCaseKeys()`
+ * (not needed here, the API already returns camelCase).
+ */
+export const NotificationMap: { [key: string]: string } = {
+  isRead: 'hasRead',
+};
--- a/src/models/notification/notifications.internal-types.ts
+++ b/src/models/notification/notifications.internal-types.ts
@@ -1,0 +1,58 @@
+/**
+ * Internal-only types for the Notification service.
+ *
+ * NOT exported from the public barrel (`src/models/notification/index.ts`).
+ */
+
+import type { NotificationPriority, NotificationCategory } from './notifications.types';
+
+/**
+ * Raw notification entry shape as returned by `/odata/v1/NotificationEntry`.
+ *
+ * Mirrors the API contract exactly: it uses the API's `isRead` field (renamed to
+ * `hasRead` in the SDK response) and includes the internal/transport-layer fields
+ * the SDK drops before returning to consumers.
+ */
+export interface RawNotificationEntry {
+  id: string;
+  message: string | null;
+  /** API read flag — renamed to `hasRead` in the SDK response. */
+  isRead: boolean;
+  publisherName: string;
+  publisherId: string;
+  topicName: string;
+  topicKeyName: string;
+  topicId: string;
+  userId: string;
+  userEmail: string | null;
+  tenantId: string | null;
+  priority: NotificationPriority;
+  category: NotificationCategory;
+  messageParam: string | null;
+  redirectionUrl: string | null;
+  publishedOn: number;
+  // Internal/transport-layer fields the SDK strips before returning to consumers:
+  entityOrgName?: string | null;
+  entityTenantName?: string | null;
+  serviceRegistryName?: string | null;
+  messageTemplateKey?: string | null;
+  messageVersion?: number;
+  publicationId?: string;
+  correlationId?: string | null;
+  partitionKey?: string;
+}
+
+/**
+ * Fields stripped from each {@link RawNotificationEntry} before it is returned to the
+ * SDK consumer as a {@link NotificationGetResponse}.
+ */
+export const INTERNAL_NOTIFICATION_FIELDS = [
+  'entityOrgName',
+  'entityTenantName',
+  'serviceRegistryName',
+  'messageTemplateKey',
+  'messageVersion',
+  'publicationId',
+  'correlationId',
+  'partitionKey',
+] as const satisfies ReadonlyArray<keyof RawNotificationEntry>;
--- a/src/models/notification/notifications.models.ts
+++ b/src/models/notification/notifications.models.ts
@@ -1,0 +1,70 @@
+/**
+ * Notification service model — public response shapes and the ServiceModel interface
+ * that drives generated API documentation.
+ */
+
+import type {
+  HasPaginationOptions,
+  NonPaginatedResponse,
+  PaginatedResponse,
+} from '../../utils/pagination/types';
+
+import type {
+  NotificationGetAllOptions,
+  NotificationGetResponse,
+} from './notifications.types';
+
+/**
+ * Public surface of the Notifications service. JSDoc on this interface drives
+ * the generated API reference documentation.
+ *
+ * Every method takes the tenant GUID as the first argument — the notification
+ * API identifies the acting tenant via the `X-UIPATH-Internal-TenantId` header
+ * and the SDK forwards `tenantId` into that header on each call.
+ */
+export interface NotificationServiceModel {
+  /**
+   * Lists notifications from the current user's inbox.
+   *
+   * Returns the full list when no pagination params are provided, or a paginated cursor result
+   * when any of `pageSize`/`cursor`/`jumpToPage` is supplied. Supports OData `filter` and
+   * `orderby` query options.
+   *
+   * @param tenantId - Tenant GUID
+   * @param options - Optional OData query and pagination options
+   * @returns Promise resolving to either a {@link NonPaginatedResponse}<{@link NotificationGetResponse}> or a {@link PaginatedResponse}<{@link NotificationGetResponse}> when pagination options are used.
+   *
+   * @example Basic usage
+   * ```typescript
+   * import { Notifications } from '@uipath/uipath-typescript/notifications';
+   *
+   * const notifications = new Notifications(sdk);
+   * const all = await notifications.getAll('<tenantId>');
+   * ```
+   *
+   * @example Filter unread, most recent first
+   * ```typescript
+   * const unread = await notifications.getAll('<tenantId>', {
+   *   filter: 'hasRead eq false',
+   *   orderby: 'publishedOn desc',
+   * });
+   * ```
+   *
+   * @example First page with pagination
+   * ```typescript
+   * const page1 = await notifications.getAll('<tenantId>', { pageSize: 20 });
+   * if (page1.hasNextPage) {
+   *   const page2 = await notifications.getAll('<tenantId>', { cursor: page1.nextCursor });
+   * }
+   * ```
+   * @internal
+   */
+  getAll<T extends NotificationGetAllOptions = NotificationGetAllOptions>(
+    tenantId: string,
+    options?: T
+  ): Promise<
+    T extends HasPaginationOptions<T>
+      ? PaginatedResponse<NotificationGetResponse>
+      : NonPaginatedResponse<NotificationGetResponse>
+  >;
+}
--- a/src/models/notification/notifications.types.ts
+++ b/src/models/notification/notifications.types.ts
@@ -1,0 +1,88 @@
+/**
+ * Notification inbox types — raw API shapes and request/response options.
+ */
+
+import type { PaginationOptions } from '../../utils/pagination/types';
+
+/**
+ * Priority level assigned to a notification by the publisher.
+ */
+export enum NotificationPriority {
+  Low = 'Low',
+  Medium = 'Medium',
+  High = 'High',
+  Critical = 'Critical',
+}
+
+/**
+ * Severity classification of a notification topic.
+ */
+export enum NotificationCategory {
+  /** Informational — no action required. */
+  Info = 'Info',
+  /** Successful operation completed. */
+  Success = 'Success',
+  /** Warning — degraded behaviour or non-fatal issue. */
+  Warn = 'Warn',
+  /** Error — operation failed but the system continues. */
+  Error = 'Error',
+  /** Fatal — unrecoverable failure. */
+  Fatal = 'Fatal',
+}
+
+/**
+ * Notification entry as returned by `GET /odata/v1/NotificationEntry`.
+ *
+ * Field selection: many internal/transport-layer fields (`partitionKey`, `correlationId`,
+ * `publicationId`, `messageVersion`, `messageTemplateKey`, `serviceRegistryName`,
+ * `entityOrgName`, `entityTenantName`) are returned by the API but dropped from the SDK
+ * because they have no use for an application developer.
+ */
+export interface NotificationGetResponse {
+  /** Notification GUID. */
+  id: string;
+  /** Resolved notification message text. */
+  message: string | null;
+  /** Whether the user has read this notification. */
+  hasRead: boolean;
+  /** Name of the publisher (e.g. `Orchestrator`, `Actions`). */
+  publisherName: string;
+  /** Publisher GUID. */
+  publisherId: string;
+  /** Human-readable topic name. */
+  topicName: string;
+  /** Stable topic identifier (e.g. `Process.JobExecution.Faulted`). */
+  topicKeyName: string;
+  /** Topic GUID. */
+  topicId: string;
+  /** GUID of the user this notification belongs to (returned uppercase by the API). */
+  userId: string;
+  /** Email of the user. Often `null`. */
+  userEmail: string | null;
+  /** Tenant GUID this notification belongs to. Often `null` for org-scoped notifications. */
+  tenantId: string | null;
+  /** Notification priority. */
+  priority: NotificationPriority;
+  /** Notification severity category. */
+  category: NotificationCategory;
+  /** JSON string of template parameters — parse with `JSON.parse()`. May be `null`. */
+  messageParam: string | null;
+  /** URL to navigate to when the notification is clicked. */
+  redirectionUrl: string | null;
+  /** Unix epoch **seconds** when the notification was published. */
+  publishedOn: number;
+}
+
+/**
+ * Options for `Notifications.getAll()`.
+ *
+ * Supports OData query options (`filter`, `orderby`) and SDK cursor pagination.
+ *
+ * Notes:
+ * - `$select` and `$expand` are not exposed because the API returns 500 on `$select`
+ *   and there are no expandable relationships on this endpoint.
+ */
+export type NotificationGetAllOptions = PaginationOptions & {
+  filter?: string;
+  orderby?: string;
+};
--- a/src/services/notification/notifications.ts
+++ b/src/services/notification/notifications.ts
@@ -1,0 +1,128 @@
+/**
+ * NotificationService — manages the current user's notification inbox.
+ */
+
+import { track } from '../../core/telemetry';
+import { BaseService } from '../base';
+
+import type {
+  NotificationGetAllOptions,
+  NotificationGetResponse,
+} from '../../models/notification/notifications.types';
+import type {
+  NotificationServiceModel,
+} from '../../models/notification/notifications.models';
+import {
+  INTERNAL_NOTIFICATION_FIELDS,
+  type RawNotificationEntry,
+} from '../../models/notification/notifications.internal-types';
+
+import { NotificationMap } from '../../models/notification/notifications.constants';
+import { ODATA_OFFSET_PARAMS, ODATA_PAGINATION } from '../../utils/constants/common';
+import { NOTIFICATION_ENDPOINTS } from '../../utils/constants/endpoints';
+import { TENANT_ID } from '../../utils/constants/headers';
+import { createHeaders } from '../../utils/http/headers';
+import { transformData, transformOptions } from '../../utils/transform';
+import {
+  HasPaginationOptions,
+  NonPaginatedResponse,
+  PaginatedResponse,
+} from '../../utils/pagination/types';
+import { PaginationHelpers } from '../../utils/pagination/helpers';
+import { PaginationType } from '../../utils/pagination/internal-types';
+
+/**
+ * Service for interacting with the UiPath Notification inbox.
+ *
+ * Provides list operations against the current user's notifications (the
+ * `/odata/v1/NotificationEntry` API). Further inbox operations (mark-read,
+ * delete) land in follow-up PRs.
+ *
+ * Every public method takes the acting tenant GUID as the first argument — the
+ * notification API identifies the tenant via the `X-UIPATH-Internal-TenantId`
+ * header and the SDK forwards `tenantId` into that header on each call.
+ */
+export class NotificationService extends BaseService implements NotificationServiceModel {
+  /**
+   * Lists notifications from the current user's inbox.
+   *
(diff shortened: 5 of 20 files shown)


src/models/notification/notifications.types.ts
```diff
@@ -0,0 +1,107 @@
...
+export enum NotificationMode {
+  /** Real-time in-app push (SignalR WebSocket). Always available. */
+  InApp = 'InApp',
+  /** Email delivery. */
+  Email = 'Email',
+  /** Slack delivery via Integration Service. */
+  Slack = 'Slack',
+  /** Microsoft Teams delivery via Integration Service. */
+  Teams = 'Teams',
+}
```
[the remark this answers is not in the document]
sarthak688: Addressed in 337c26cf — NotificationMode enum + JSDoc removed from notifications.types.ts. PR 4 (#516) will (re-)define it where it's first consumed, restoring the JSDoc cross-reference once getSupportedChannels exists.

docs/oauth-scopes.md
```diff
@@ -209,6 +209,12 @@ The `ConversationalAgents` scope is required for real-time WebSocket sessions (`
 | `getPolicyTraces()` | `Insights.RealTimeData Insights OR.Folders.Read` |
 | `getOperationSummary()` | `Insights.RealTimeData Insights OR.Folders.Read` |
 
+## Notifications
+
+| Method | OAuth Scope |
+|--------|-------------|
+| `getAll()` | `NotificationService` |
```
vnaren23: We shouldnt document here. `NotificationService` is an internal scope and not external scope. 
https://github.com/UiPath/first-party-service-metadata/blob/master/UiPath.Identity.ClientDefinitions/UiPath.Identity.ClientDefinitions/Definitions/NotificationService/ScopeProperties.json#L4

We only document external scoped method in this file.

src/models/notification/notifications.models.ts
vnaren23, line 1: All methods should be tagged `@internal` since we do not want them documented in TS SDK docs. See some DF method implementations for examples.

docs/oauth-scopes.md
sarthak688: Addressed in bd839141 — removed the Notifications section from docs/oauth-scopes.md. Also removed the docs/pagination.md row and the mkdocs.yml nav entry (since the same internal-scope reasoning applies to those). Same treatment will land in PRs #513–#519 for their respective methods.

src/models/notification/notifications.models.ts
sarthak688: Addressed in bd839141 — tagged `getAll` `@internal` in both the ServiceModel interface and the service class JSDoc (per-method, matching the DF pattern in `entities.models.ts` / `choicesets.models.ts`). Same treatment will land in PRs #513–#519 for their respective methods.

src/services/notification/notifications.ts
```diff
@@ -0,0 +1,125 @@
...
+   * ```
+   *
+   * @example First page with pagination
+   * ```typescript
+   * const page1 = await notifications.getAll('<tenantId>', { pageSize: 20 });
+   * if (page1.hasNextPage) {
+   *   const page2 = await notifications.getAll('<tenantId>', { cursor: page1.nextCursor });
+   * }
+   * ```
+   * @internal
```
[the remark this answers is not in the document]
sarthak688: @vnaren23 do we want to make this method internal?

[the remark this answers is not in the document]
vnaren23: Yes. 
All methods which are not accessible via [external app](https://docs.uipath.com/automation-cloud/automation-cloud/latest/admin-guide/authorizing-external-applications) should be marked internal

src/services/notification/notifications.ts
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Lists notifications from the current user's inbox.
+   *
+   * Returns the full list when no pagination params are provided, or a paginated cursor result
+   * when any of `pageSize`/`cursor`/`jumpToPage` is supplied. Supports OData `filter` and
+   * `orderby` query options.
+   *
+   * @param tenantId - Tenant GUID (sent via `X-UIPATH-Internal-TenantId`)
+   * @param options - Optional OData query and pagination options
+   * @returns Array of notifications, or a paginated result when pagination params are supplied
```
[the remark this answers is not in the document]
sarthak688: Addressed in 7634b6c3 — service class `@returns` now matches the ServiceModel wording (Promise resolving to NonPaginatedResponse/PaginatedResponse with TypeDoc `@link` refs).

tests/utils/constants/notification.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const NOTIFICATION_TEST_CONSTANTS = {
+  // Tenant GUID — passed as the first arg to every Notifications method
+  // (sent to the API via the X-UIPATH-Internal-TenantId header)
+  TENANT_ID: '99999999-9999-4999-8999-999999999999',
+
+  // Notification entry identifiers
+  NOTIFICATION_ID: '11111111-1111-4111-8111-111111111111',
+  NOTIFICATION_ID_2: '22222222-2222-4222-8222-222222222222',
+  NOTIFICATION_ID_3: '33333333-3333-4333-8333-333333333333',
```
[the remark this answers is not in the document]
sarthak688: Addressed in 7634b6c3 — removed `NOTIFICATION_ID_2`, `NOTIFICATION_ID_3`, and `ERROR_NOTIFICATION_NOT_FOUND` from this PR. They'll be reintroduced in the follow-up PRs that actually use them (mark-read / delete flows).

src/utils/constants/endpoints/base.ts
```diff
@@ -9,3 +9,14 @@ export const IDENTITY_BASE = 'identity_';
...
+/**
+ * Notification service base. The notification service is routed at the **organization**
+ * level — its URLs do not include a tenant segment (unlike most UiPath services).
+ *
+ * The `../` prefix relies on `URL` path normalization to collapse the tenant segment
+ * that {@link ApiClient} unconditionally inserts (`{orgName}/{tenantName}/{path}`). Concretely,
+ * `{orgName}/{tenantName}/../notificationservice_/...` resolves to `{orgName}/notificationservice_/...`.
+ *
+ * Do NOT remove the leading `../`. See real-API curl confirmation in the PR description.
+ */
```
vnaren23, line 21: @Raina451 is there a better way of handling this?

src/models/notification/notifications.types.ts
```diff
@@ -0,0 +1,86 @@
...
+ * `entityOrgName`, `entityTenantName`) are returned by the API but dropped from the SDK
+ * because they have no use for an application developer.
+ */
+export interface NotificationGetResponse {
+  /** Notification GUID. */
+  id: string;
+  /** Resolved notification message text. */
+  message: string | null;
+  /** Whether the user has read this notification. */
+  isRead: boolean;
```
vnaren23: `hasRead` is a better name.

sarthak688: Going with `isRead` as we have same contract in API

vnaren23: We do not want to leak whatever is present from API/Swagger. Please revisit any/all namings which are user facing and rename them if needed. SDK internally can map it to the exact property which the API has. Across the SDK we've made multiple renames to make it more developer/user friendly.

sarthak688: Ok renaming but i don't think it has to do anything with the user experience. Both wording are understandable from technical POV

vnaren23: It is from a user experience perspective. When I read the property `isRead` it wasn't intuitive, I had to read the description to figure out. Even the description mentions `has read` which is why I suggested `hasRead`. Naming is always subjective :)

sarthak688: @vnaren23 There's a gap I see in request filters that's worth calling out — and it isn't specific to notifications, it already exists across the Orchestrator methods.

Field renames are applied **only to the response, not the request**. So a user filters with the API field name but reads the response with the renamed field — two different names for the same thing.

The existing user experience is not good IMO. We first need to close this gap generically by some common logic and then do renaming.

  **Example — `jobs.getAll` (already shipped):** `JobMap` renames `releaseName` → `processName`:

  ```ts
  const page = await jobs.getAll({ filter: "releaseName eq 'InvoiceBot'" });
  //                                        ^^^^^^^^^^^ filter uses the API name
  page.items[0].processName;   // ✅ response uses the renamed field
  page.items[0].releaseName;   // ❌ undefined

  filter: "processName eq ..." (the name the user actually sees in the response) gets forwarded verbatim and rejected by the API. 
Same for orderby and every other renamed field. isRead → hasRead here would inherit the same gap.

vnaren23: @swati354 @Raina451 can also add more details. We are aware of this and have a backlog item to fix it. Are there any NS methods we are about to introduce have such filtering capability?

sarthak688: Yes isRead has filtering capability

vnaren23: Okay, we'll priotitize this at our end. Even then this would be a SDK level change, lets not use this as a reason to not rename.

sarthak688: It's not a reason.
It is a pre-requisite before we do rename at our end

tests/unit/services/notification/notifications.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      // public fields preserved with exact values
+      expect(stripped.id).toBe(NOTIFICATION_TEST_CONSTANTS.NOTIFICATION_ID);
+      expect(stripped.priority).toBe(NotificationPriority.High);
+      expect(stripped.category).toBe(NotificationCategory.Error);
+      expect(stripped.publishedOn).toBe(NOTIFICATION_TEST_CONSTANTS.PUBLISHED_ON);
+      expect(stripped.message).toBe(NOTIFICATION_TEST_CONSTANTS.MESSAGE);
+      expect(stripped.isRead).toBe(false);
+    });
+  });
+});
```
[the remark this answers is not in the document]
sarthak688: Discussed here https://uipath-product.slack.com/archives/C0B967CCFM3/p1781604744756559

src/models/notification/notifications.types.ts
sarthak688: renaming done

Sarath1018:
Please update PR description

Sarath1018:
Can you add integration tests, test them locally using bearer token and once local testing is done skip them and push. So that in future if PAT token supports the required scopes we can just un-skip those tesrs

src/models/notification/notifications.constants.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Notification field mappings (API field name → SDK field name).
+ *
+ * Semantic renames only — case conversion is handled by `pascalToCamelCaseKeys()`
+ * (not needed here, the API already returns camelCase).
+ */
+export const NotificationMap: { [key: string]: string } = {
+  isRead: 'hasRead',
```
Sarath1018, line 8: question: why this transformation? hasRead looks good but why have you selected to change only this field, Is this decided by your team?

src/models/notification/notifications.internal-types.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import type { NotificationGetResponse } from './notifications.types';
+
+/**
+ * Raw notification entry shape as returned by `/odata/v1/NotificationEntry` — uses the
+ * API's `isRead` field (renamed to `hasRead` in the SDK response) and includes the
+ * internal/transport-layer fields the SDK drops before returning to consumers.
+ */
+export interface RawNotificationEntry extends Omit<NotificationGetResponse, 'hasRead'> {
```
Sarath1018: For Raw API response its fine to duplicate the fields as we do in other places, Deriving the raw type by taking the SDK response and modifying it is not a good practice

src/utils/constants/endpoints/base.ts
Sarath1018: You can construct the URL correctly in ApiClient rather than relying on ../
(or)
can you add a test asserting the org-level URL drops the tenant segment? there's no guard against a future ApiClient change silently breaking the routing

src/models/notification/notifications.constants.ts
sarthak688: No it was a ask from @vnaren23 to rename the field.

src/utils/constants/endpoints/base.ts
sarthak688: added unit test

src/models/notification/notifications.internal-types.ts
sarthak688: done

src/models/notification/notifications.types.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Options for `Notifications.getAll()`.
+ *
+ * Supports OData query options (`filter`, `orderby`) and SDK cursor pagination.
+ *
+ * Notes:
+ * - `$select` and `$expand` are not exposed because the API returns 500 on `$select`
+ *   and there are no expandable relationships on this endpoint.
+ */
+export type NotificationGetAllOptions = Omit<RequestOptions, 'expand' | 'select'> & PaginationOptions;
```
Sarath1018: nit: Declare like this instead of using Omit

```
export type NotificationGetAllOptions = PaginationOptions & {
  filter?: string;
  orderby?: string;
};
```

src/services/notification/notifications.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @internal
+ */
+export function stripInternalNotificationFields(item: RawNotificationEntry): NotificationGetResponse {
+  const stripped: RawNotificationEntry = { ...item };
+  for (const field of INTERNAL_NOTIFICATION_FIELDS) {
+    delete stripped[field];
+  }
+  return transformData(stripped, NotificationMap) as unknown as NotificationGetResponse;
+}
```
Sarath1018: Inline this as a local `const transformNotificationResponse` inside `getAll`.
Also rename stripInternalNotificationFields → transformNotificationResponse to match the convention.
Test the transformation through `getAll()` rather than a direct import.

Sarath1018 (approved):

src/services/notification/notifications.ts
sarthak688: done

src/models/notification/notifications.types.ts
sarthak688: done

src/models/notification/notifications.models.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export interface NotificationServiceModel {
+  /**
+   * Lists notifications from the current user's inbox.
+   *
+   * Returns the full list when no pagination params are provided, or a paginated cursor result
+   * when any of `pageSize`/`cursor`/`jumpToPage` is supplied. Supports OData `filter` and
+   * `orderby` query options.
+   *
+   * @param tenantId - Tenant GUID (sent via `X-UIPATH-Internal-TenantId`)
```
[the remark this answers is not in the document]
Sarath1018: @sarthak688 pls fix this

src/models/notification/notifications.models.ts
```diff
@@ -0,0 +1,70 @@
...
```
vnaren23: We dont need to mention sent via header. Lets remove that.

src/services/notification/notifications.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export class NotificationService extends BaseService implements NotificationServiceModel {
+  /**
+   * Lists notifications from the current user's inbox.
+   *
+   * Returns the full list when no pagination params are provided, or a paginated cursor result
+   * when any of `pageSize`/`cursor`/`jumpToPage` is supplied. Supports OData `filter` and
+   * `orderby` query options.
+   *
+   * @param tenantId - Tenant GUID (sent via `X-UIPATH-Internal-TenantId`)
```
vnaren23: here as well

vnaren23 (approved):

PR #532, opened 2026-06-16
fix(auth): bootstrap token for host-embedded coded apps via UIP protocol

NishankSiddharth:
## Summary

Generic **host-embedded** coded apps (using the `UIP.refreshToken`/`UIP.tokenRefreshed` postMessage protocol) never bootstrap their first token, so embedded auth is effectively dead. This fixes it by mirroring exactly what Action Center already does.

### Root cause
`TokenManager.getValidToken()` throws `"No authentication token available"` when `tokenInfo` is missing — **before** it would delegate to the embedded/AC manager (`src/core/auth/token-manager.ts:70`). Action Center avoids this because `UiPath.#initializeWithConfig()` seeds an empty token for it (`src/core/uipath.ts`):

```ts
if (hasSecretAuth || isInActionCenter) {
  this.#authService.authenticateWithSecret(config.secret ?? '');
  this.#initialized = true;
}
```

That empty seed does two things AC relies on: (a) `getValidToken()` no longer throws → it delegates to the manager, which requests a token via postMessage; (b) `isAuthenticated()` becomes `true` → `initialize()` returns early instead of starting a real OAuth redirect. The generic embedded path (`isHostEmbedded`) was simply never added to this gate.

### Fix (surgical — `uipath.ts` only)
Add the generic-embedded case to the **same** seed gate, using the identical predicate `TokenManager` already uses to construct the `EmbeddedTokenManager`:

```ts
const isGenericHostEmbedded =
  isHostEmbedded && !!embeddingOrigin && isValidHostOrigin(embeddingOrigin);
if (hasSecretAuth || isInActionCenter || isGenericHostEmbedded) {
  this.#authService.authenticateWithSecret(config.secret ?? '');
  this.#initialized = true;
}
```

No change to the Action Center path, no `getValidToken` reorder, no unrelated changes.

## Test plan
- [x] Added `Host-Embedded Seeding` tests in `tests/unit/core/uipath.test.ts` (mirrors the AC seeding pattern): seeds + `isInitialized()` for a trusted `embeddingOrigin`; `initialize()` returns early (no OAuth redirect); does NOT seed for untrusted/null origin; regression guard that a normal non-embedded OAuth app is NOT seeded.
- [x] Build/typecheck passes (`npm run build`, exit 0).
- [x] Telemetry-free auth suites pass locally (`host-token-request`, `embedded-token-manager`, `action-center-token-manager` — 35 tests).
- [ ] **CI must validate** `uipath.test.ts` + `service.test.ts` (Action Center): these import `@uipath/core-telemetry`, a private package that couldn't be installed in the local dev environment (401), so the new tests + AC no-regression check run in CI.

## Context
Enables the UiPath Experiences portal (governance coded dashboards) to host coded apps over the UIP embedded OAuth protocol. The host side is implemented separately; this is the SDK-side bootstrap that lets an embedded app emit its first `UIP.refreshToken`.

Generated with Claude Code

### DIFF
--- a/src/core/uipath.ts
+++ b/src/core/uipath.ts
@@ -9,6 +9,7 @@ import { SDKInternalsRegistry } from './internals';
 import { loadFromMetaTags } from './config/runtime';
 import type { IUiPath } from './types';
 import { isInActionCenter } from '../utils/platform';
+import { trustedEmbeddingOrigin } from './auth/host-token-request';
 
 /**
  * UiPath - Core SDK class for authentication and configuration management.
@@ -136,10 +137,12 @@ export class UiPath implements IUiPath {
     // Track SDK initialization
     trackEvent('Sdk.Auth');
 
-    /** Auto-initialize for secret-based auth
-     * When viewed in Action Center, initialize tokenInfo with empty token. When an sdk call is made Action Center passes the token to sdk.
+    /** Auto-initialize for secret-based auth, Action Center, and generic host-embedded apps.
+     * When viewed in Action Center or embedded in a UiPath host frame via the UIP protocol,
+     * initialize tokenInfo with an empty token so getValidToken() can bootstrap via postMessage.
+     * When an sdk call is made, the host passes the token to the sdk.
      */
-    if (hasSecretAuth || isInActionCenter) {
+    if (hasSecretAuth || isInActionCenter || trustedEmbeddingOrigin) {
       this.#authService.authenticateWithSecret(config.secret ?? '');
       this.#initialized = true;
     }
--- a/tests/unit/core/uipath.test.ts
+++ b/tests/unit/core/uipath.test.ts
@@ -47,6 +47,34 @@ vi.mock('../../../src/core/config/runtime', () => ({
   loadFromMetaTags: mockLoadFromMetaTags,
 }));
 
+// Mock platform utilities so we can simulate host-embedded scenarios.
+// Defaults to non-embedded (matching normal browser / server-side use).
+const { mockPlatform } = vi.hoisted(() => ({
+  mockPlatform: {
+    isBrowser: false,
+    isInActionCenter: false,
+    isHostEmbedded: false,
+    embeddingOrigin: null as string | null,
+  },
+}));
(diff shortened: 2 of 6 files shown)


src/core/uipath.ts
```diff
@@ -136,10 +137,14 @@ export class UiPath implements IUiPath {
...
 
-    /** Auto-initialize for secret-based auth
-     * When viewed in Action Center, initialize tokenInfo with empty token. When an sdk call is made Action Center passes the token to sdk.
+    /** Auto-initialize for secret-based auth, Action Center, and generic host-embedded apps.
+     * When viewed in Action Center or embedded in a UiPath host frame via the UIP protocol,
+     * initialize tokenInfo with an empty token so getValidToken() can bootstrap via postMessage.
+     * When an sdk call is made, the host passes the token to the sdk.
      */
-    if (hasSecretAuth || isInActionCenter) {
+    const isGenericHostEmbedded =
```
deepeshrai-tech: same logic is there in auth/token-manager, can we extract it ?

deepeshrai-tech:
rest LGTM

deepeshrai-tech (approved):

src/core/uipath.ts
NishankSiddharth: extracted to a common const and reused in both places

tests/unit/core/uipath.test.ts
```diff
@@ -430,6 +458,82 @@ describe('UiPath Core', () => {
...
+      redirectUri: TEST_CONSTANTS.REDIRECT_URI,
+      scope: 'offline_access',
+    };
+
+    afterEach(() => {
+      // Reset platform to non-embedded state after each test so this block's
+      // state never leaks into the next describe (per agent_docs/rules.md).
+      mockPlatform.isInActionCenter = false;
+      mockPlatform.isHostEmbedded = false;
+      mockPlatform.embeddingOrigin = null;
```
Raina451, line 484: also reset `isBrowser`

Raina451 (approved):

PR #536, opened 2026-06-17
fix: rewrite SDK field names in OData filter/orderby/select/expand [PLT-102084]

ninja-shreyash:
## Summary

Closes the gap where SDK-side field renames were applied only on responses. Filtering / sorting / projecting by the SDK name silently shipped that name to the API and got rejected. Also removes a latent chained-rename quirk in `transformData` that affected maps with overlapping source/target keys (`ProcessMap` specifically).

Scope is **Jobs only** so reviewers can evaluate the approach in isolation. The helper (`transformOptions`) and the no-chain `transformData` refactor are generic and benefit every map; the other affected services are picked up in follow-up PRs from the rollout tracker.

Tracker: https://uipath.atlassian.net/wiki/spaces/~712020f280189e56f64c78bc3f288430f26209/pages/90827849894/

### 1. `transformOptions` — symmetric counterpart of `transformRequest`

Adds **`transformOptions(options, EntityMap)`** in `src/utils/transform.ts`. Where `transformRequest` rewrites SDK→API names on object keys in a request body, `transformOptions` rewrites SDK→API names on identifier tokens inside OData `filter` / `orderby` / `select` / `expand` strings.

The tokenizer is a single regex with two alternatives — quoted literals (atomic, with `''` escape) and identifiers (greedy, full-token). `String.prototype.replace` walks the string; quoted-literal matches pass through unchanged, identifier matches are looked up in `reverseMap(EntityMap)`.

\`\`\`ts
const ODATA_TOKEN_RE = /'(?:[^']|'')*'|[A-Za-z_][A-Za-z0-9_]*/g;

export function rewriteODataIdentifiers(expression, requestMap) {
  if (!expression) return expression;
  return expression.replace(ODATA_TOKEN_RE, (match) =>
    match.startsWith(\"'\") ? match : (requestMap[match] ?? match),
  );
}
\`\`\`

### 2. `transformData` — no more chained renames

The previous implementation iterated map entries and mutated the running result in place, which caused chained renames to compose: a map with both `a → b` and `b → c` would turn input `{a: 1}` into `{c: 1}` (passing through `b` along the way). `ProcessMap`'s `releaseKey → processKey → packageKey` is the canonical example.

The new implementation walks the **input data's keys** and looks each up in the mapping exactly once. Renames are independent — chains can't form. Strictly cheaper (`O(D)` vs `O(D + M)`, no shallow-copy, no `delete` calls / hidden-class transitions), and output key order now mirrors input order.

All pre-existing tests pass unchanged — no test mock contained both ends of a chain, so the chained behaviour was a latent quirk, not an active behaviour.

### 3. Jobs wired end-to-end

The service layer calls `transformOptions(options, JobMap)` directly before delegating, mirroring how `processes.start` calls `transformRequest` on its body. `PaginationHelpers.getAll` stays map-agnostic — services own request-side translation, helpers plumb.

### 4. Symmetry with `transformRequest`

Both request-side helpers now follow the same convention: **services own request-side translation. Helpers plumb.**

| Helper | Where the field name lives | Used by |
|---|---|---|
| `transformRequest(body, EntityMap)` | Object key in a request body | `processes.start`, conversations create |
| `transformOptions(opts, EntityMap)` | Identifier token inside an OData string | Jobs (this PR) — follow-up services next |

Both reverse `EntityMap` internally via `reverseMap`, so each service declares its map exactly once (API → SDK direction for responses); the request direction is derived at call time.

## Before / after

\`\`\`ts
// Before — user needed two names per field:
await jobs.getAll({ filter: \"releaseName eq 'InvoiceBot'\" });   // API name (worked)
page.items[0].processName;                                       // SDK name (read)

// After — one name end-to-end:
await jobs.getAll({
  filter: \"processName eq 'InvoiceBot'\",
  orderby: 'createdTime desc',
});
page.items[0].processName;
\`\`\`

API names continue to work unchanged. The rewriter is a forward-only lookup (`requestMap[match] ?? match`) — anything not in the reversed map (API names, operators like `eq`, non-renamed fields like `state`) passes through untouched.

## Test plan

- [x] 17 unit tests for the regex tokenizer covering quoted-literal preservation, `''` escape, multi-clause filters with `and`/`or`, function args (`contains`, `startswith`), orderby `asc`/`desc`, comma-separated `select`/`expand`, identifier substring-collision safety, empty input, empty map.
- [x] 7 unit tests for the no-chain `transformData` contract: basic rename, passthrough, original-key dropped, no chaining, both ends of a chain present without compose, array recursion, empty input.
- [x] Per-service wiring tests: `getAll` asserts the translated string arrives at `PaginationHelpers.getAll` (e.g. `releaseName eq 'InvoiceBot' and organizationUnitId eq 7`); `getById` asserts rewritten `select` reaches the HTTP layer.
- [x] Integration test exercising SDK-renamed fields end-to-end on a live tenant: `orderby: 'createdTime desc'` and `select: 'key,processName,state,createdTime'`. The API would 400 on either of those without the rewrite.
- [x] \`npm run typecheck\` — clean
- [x] \`npm run lint\` — 0 errors
- [x] \`npm run test:unit\` — 1905 / 1905 pass
- [x] \`npm run build\` — succeeds
- [x] **Manual E2E** on a tenant: \`processName\` (SDK) filter 400s on 1.4.1, succeeds on this PR; \`orderby createdTime desc\` and \`getById\` with \`select: 'processName,createdTime,state'\` likewise.

Refs: [PLT-102084](https://uipath.atlassian.net/browse/PLT-102084), surfaced in [PR #512 thread](https://github.com/UiPath/uipath-typescript/pull/512#discussion_r3426534034)

[PLT-102084]: https://uipath.atlassian.net/browse/PLT-102084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/models/orchestrator/jobs.models.ts
+++ b/src/models/orchestrator/jobs.models.ts
@@ -42,11 +42,6 @@ export interface JobServiceModel {
    * const folderJobs = await jobs.getAll({ folderId: <folderId> });
    *
    * // With filtering
-   * const runningJobs = await jobs.getAll({
-   *   filter: "state eq 'Running'"
-   * });
-   *
-   * // Filter and sort by SDK field names — same names you read from the response
    * const recentInvoiceJobs = await jobs.getAll({
    *   filter: "processName eq 'InvoiceBot'",
    *   orderby: 'createdTime desc',
--- a/src/models/orchestrator/jobs.models.ts
+++ b/src/models/orchestrator/jobs.models.ts
@@ -46,6 +46,12 @@ export interface JobServiceModel {
    *   filter: "state eq 'Running'"
    * });
    *
+   * // Filter and sort by SDK field names — same names you read from the response
+   * const recentInvoiceJobs = await jobs.getAll({
+   *   filter: "processName eq 'InvoiceBot'",
+   *   orderby: 'createdTime desc',
+   * });
+   *
    * // First page with pagination
    * const page1 = await jobs.getAll({ pageSize: 10 });
    *
--- a/src/services/orchestrator/jobs/jobs.ts
+++ b/src/services/orchestrator/jobs/jobs.ts
@@ -1,7 +1,7 @@
 import { FolderScopedService } from '../../folder-scoped';
 import { RawJobGetResponse, JobGetAllOptions, JobGetByIdOptions, JobStopOptions, JobResumeOptions } from '../../../models/orchestrator/jobs.types';
 import { JobServiceModel, JobGetResponse, createJobWithMethods } from '../../../models/orchestrator/jobs.models';
-import { addPrefixToKeys, pascalToCamelCaseKeys, rewriteODataRequestFields, transformData } from '../../../utils/transform';
+import { addPrefixToKeys, pascalToCamelCaseKeys, transformOptions, transformData } from '../../../utils/transform';
 import { JOB_ENDPOINTS } from '../../../utils/constants/endpoints';
 import { ODATA_PAGINATION, ODATA_OFFSET_PARAMS, ODATA_PREFIX } from '../../../utils/constants/common';
 import { JobMap, JOB_KEY_RESOLUTION_CHUNK_SIZE } from '../../../models/orchestrator/jobs.constants';
@@ -149,7 +149,7 @@ export class JobService extends FolderScopedService implements JobServiceModel {
     }
 
     const headers = createHeaders({ [FOLDER_ID]: folderId });
-    const rewrittenOptions = options ? rewriteODataRequestFields(options, JobMap) : undefined;
+    const rewrittenOptions = options ? transformOptions(options, JobMap) : undefined;
     const keysToPrefix = Object.keys(rewrittenOptions ?? {});
     const apiOptions = rewrittenOptions ? addPrefixToKeys(rewrittenOptions, ODATA_PREFIX, keysToPrefix) : {};
 
--- a/src/services/orchestrator/jobs/jobs.ts
+++ b/src/services/orchestrator/jobs/jobs.ts
@@ -55,11 +55,6 @@ export class JobService extends FolderScopedService implements JobServiceModel {
    * const folderJobs = await jobs.getAll({ folderId: <folderId> });
    *
    * // With filtering
-   * const runningJobs = await jobs.getAll({
-   *   filter: "state eq 'Running'"
-   * });
-   *
-   * // Filter and sort by SDK field names — same names you read from the response
    * const recentInvoiceJobs = await jobs.getAll({
    *   filter: "processName eq 'InvoiceBot'",
    *   orderby: 'createdTime desc',
@@ -153,9 +148,8 @@ export class JobService extends FolderScopedService implements JobServiceModel {
     }
 
     const headers = createHeaders({ [FOLDER_ID]: folderId });
-    const rewrittenOptions = options ? transformOptions(options, JobMap) : undefined;
-    const keysToPrefix = Object.keys(rewrittenOptions ?? {});
-    const apiOptions = rewrittenOptions ? addPrefixToKeys(rewrittenOptions, ODATA_PREFIX, keysToPrefix) : {};
+    const apiFieldOptions = options ? transformOptions(options, JobMap) : {};
+    const apiOptions = addPrefixToKeys(apiFieldOptions, ODATA_PREFIX, Object.keys(apiFieldOptions));
 
     const response = await this.get<Record<string, unknown>>(
       JOB_ENDPOINTS.GET_BY_KEY(id),
--- a/src/services/orchestrator/jobs/jobs.ts
+++ b/src/services/orchestrator/jobs/jobs.ts
@@ -96,7 +96,7 @@ export class JobService extends FolderScopedService implements JobServiceModel {
     // Rewrite renamed SDK field names → API names inside OData strings
     // before delegating, mirroring the transformRequest pattern used for
     // request bodies.
-    const apiOptions = options ? transformOptions(options, JobMap) as T : options;
+    const apiOptions = options ? transformOptions(options, JobMap) : options;
 
     return PaginationHelpers.getAll({
       serviceAccess: this.createPaginationServiceAccess(),
--- a/src/services/orchestrator/jobs/jobs.ts
+++ b/src/services/orchestrator/jobs/jobs.ts
@@ -93,12 +93,16 @@ export class JobService extends FolderScopedService implements JobServiceModel {
       return createJobWithMethods(rawJob, this);
     };
 
+    // Rewrite renamed SDK field names → API names inside OData strings
+    // before delegating, mirroring the transformRequest pattern used for
+    // request bodies.
+    const apiOptions = options ? transformOptions(options, JobMap) as T : options;
+
     return PaginationHelpers.getAll({
       serviceAccess: this.createPaginationServiceAccess(),
       getEndpoint: () => JOB_ENDPOINTS.GET_ALL,
       getByFolderEndpoint: JOB_ENDPOINTS.GET_ALL,
       transformFn: transformJobResponse,
-      fieldMap: JobMap,
       pagination: {
         paginationType: PaginationType.OFFSET,
         itemsField: ODATA_PAGINATION.ITEMS_FIELD,
@@ -109,7 +113,7 @@ export class JobService extends FolderScopedService implements JobServiceModel {
           countParam: ODATA_OFFSET_PARAMS.COUNT_PARAM,
         },
       },
-    }, options) as any;
+    }, apiOptions) as any;
   }
 
   /**
--- a/src/services/orchestrator/jobs/jobs.ts
+++ b/src/services/orchestrator/jobs/jobs.ts
@@ -1,7 +1,7 @@
 import { FolderScopedService } from '../../folder-scoped';
 import { RawJobGetResponse, JobGetAllOptions, JobGetByIdOptions, JobStopOptions, JobResumeOptions } from '../../../models/orchestrator/jobs.types';
 import { JobServiceModel, JobGetResponse, createJobWithMethods } from '../../../models/orchestrator/jobs.models';
-import { addPrefixToKeys, pascalToCamelCaseKeys, transformData } from '../../../utils/transform';
+import { addPrefixToKeys, pascalToCamelCaseKeys, rewriteODataRequestFields, transformData } from '../../../utils/transform';
 import { JOB_ENDPOINTS } from '../../../utils/constants/endpoints';
 import { ODATA_PAGINATION, ODATA_OFFSET_PARAMS, ODATA_PREFIX } from '../../../utils/constants/common';
 import { JobMap, JOB_KEY_RESOLUTION_CHUNK_SIZE } from '../../../models/orchestrator/jobs.constants';
@@ -59,6 +59,12 @@ export class JobService extends FolderScopedService implements JobServiceModel {
    *   filter: "state eq 'Running'"
    * });
    *
+   * // Filter and sort by SDK field names — same names you read from the response
+   * const recentInvoiceJobs = await jobs.getAll({
+   *   filter: "processName eq 'InvoiceBot'",
+   *   orderby: 'createdTime desc',
+   * });
+   *
    * // First page with pagination
    * const page1 = await jobs.getAll({ pageSize: 10 });
    *
@@ -92,6 +98,7 @@ export class JobService extends FolderScopedService implements JobServiceModel {
       getEndpoint: () => JOB_ENDPOINTS.GET_ALL,
       getByFolderEndpoint: JOB_ENDPOINTS.GET_ALL,
       transformFn: transformJobResponse,
+      fieldMap: JobMap,
       pagination: {
         paginationType: PaginationType.OFFSET,
         itemsField: ODATA_PAGINATION.ITEMS_FIELD,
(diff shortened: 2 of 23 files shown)


tests/unit/services/orchestrator/buckets.test.ts
```diff
@@ -416,6 +417,20 @@ describe('BucketService Unit Tests', () => {
...
+      vi.mocked(PaginationHelpers.getAll).mockResolvedValue({
+        items: createMockBuckets(0),
+        totalCount: 0,
+      });
+
+      await bucketService.getAll({ filter: "path eq '/foo'" });
+
+      expect(PaginationHelpers.getAll).toHaveBeenCalledWith(
+        expect.objectContaining({ fieldMap: BucketMap }),
+        expect.objectContaining({ filter: "path eq '/foo'" }),
```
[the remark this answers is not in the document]
ninja-shreyash: Out of scope for this PR — the wiring for this service was reverted when scope narrowed to Jobs only (commit 8603c59c). Tracked as a follow-up on the [rollout page](https://uipath.atlassian.net/wiki/spaces/~712020f280189e56f64c78bc3f288430f26209/pages/90827849894/), will be addressed in the per-service follow-up PR.

src/services/orchestrator/processes/processes.ts
```diff
@@ -187,9 +188,11 @@ export class ProcessService extends FolderScopedService implements ProcessServic
       startInfo: apiRequest
     };
 
-    // Prefix all query parameter keys with '$' for OData
-    const keysToPrefix = Object.keys(queryOptions);
-    const apiOptions = addPrefixToKeys(queryOptions, ODATA_PREFIX, keysToPrefix);
+    // Rewrite SDK field names → API field names inside OData strings, then
+    // prefix all query parameter keys with '$' for OData.
+    const rewrittenQueryOptions = rewriteODataRequestFields(queryOptions, ProcessMap);
```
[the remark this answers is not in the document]
ninja-shreyash: Out of scope for this PR — the wiring for this service was reverted when scope narrowed to Jobs only (commit 8603c59c). Tracked as a follow-up on the [rollout page](https://uipath.atlassian.net/wiki/spaces/~712020f280189e56f64c78bc3f288430f26209/pages/90827849894/), will be addressed in the per-service follow-up PR.

tests/unit/services/orchestrator/assets.test.ts
```diff
@@ -229,6 +251,19 @@ describe('AssetService Unit Tests', () => {
...
+        createMockTransformedAssetCollection(),
+      );
+
+      await assetService.getAll({ filter: "createdTime gt 2025-01-01" });
+
+      expect(PaginationHelpers.getAll).toHaveBeenCalledWith(
+        expect.objectContaining({ fieldMap: AssetMap }),
+        expect.objectContaining({ filter: "createdTime gt 2025-01-01" }),
+      );
+    });
```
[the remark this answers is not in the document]
ninja-shreyash: Out of scope for this PR — the wiring for this service was reverted when scope narrowed to Jobs only (commit 8603c59c). Tracked as a follow-up on the [rollout page](https://uipath.atlassian.net/wiki/spaces/~712020f280189e56f64c78bc3f288430f26209/pages/90827849894/), will be addressed in the per-service follow-up PR.

tests/unit/services/orchestrator/processes.test.ts
```diff
@@ -151,13 +152,27 @@ describe('ProcessService Unit Tests', () => {
...
+      });
+
+      await service.getAll({ filter: "processName eq 'X'", orderby: 'createdTime desc' });
+
+      expect(PaginationHelpers.getAll).toHaveBeenCalledWith(
+        expect.objectContaining({ fieldMap: ProcessMap }),
+        expect.objectContaining({ filter: "processName eq 'X'", orderby: 'createdTime desc' }),
+      );
+    });
+
```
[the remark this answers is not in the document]
ninja-shreyash: Out of scope for this PR — the wiring for this service was reverted when scope narrowed to Jobs only (commit 8603c59c). Tracked as a follow-up on the [rollout page](https://uipath.atlassian.net/wiki/spaces/~712020f280189e56f64c78bc3f288430f26209/pages/90827849894/), will be addressed in the per-service follow-up PR.

tests/unit/utils/transform.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  it('does not partially match longer identifiers', () => {
+    expect(
+      rewriteODataIdentifiers(
+        "processNameSuffix eq 'X'",
+        { processName: 'releaseName' },
+      ),
+    ).toBe("processNameSuffix eq 'X'");
+  });
+
+  it('does not rewrite operators that happen to overlap map keys', () => {
```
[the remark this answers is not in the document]
ninja-shreyash: Fixed in 1647ee0b — renamed to clarify the test asserts the rewriter *does* rewrite operator-named keys (caller responsibility to avoid collisions).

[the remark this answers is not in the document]
ninja-shreyash: Fixed in 1647ee0b — renamed to clarify the test asserts the rewriter *does* rewrite operator-named keys (caller responsibility to avoid collisions).

tests/unit/utils/transform.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    expect(out).toEqual({
+      filter: "releaseName eq 'X'",
+      pageSize: 10,
+      folderId: 42,
+    });
+  });
+
+  it('ignores non-string values on OData keys', () => {
+    const out = transformOptions(
+      { filter: undefined, orderby: null as unknown as string },
```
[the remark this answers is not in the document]
ninja-shreyash: Fixed in 1647ee0b — dropped the cast.

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -87,6 +93,11 @@ export class JobService extends FolderScopedService implements JobServiceModel {
       return createJobWithMethods(rawJob, this);
     };
 
+    // Rewrite renamed SDK field names → API names inside OData strings
+    // before delegating, mirroring the transformRequest pattern used for
+    // request bodies.
+    const apiOptions = options ? transformOptions(options, JobMap) as T : options;
```
deepeshrai-tech: remove as T , transformOptions already returns T

deepeshrai-tech (approved):

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: Good catch — fixed. The explicit `as T` was redundant since `transformOptions<T extends Record<string, any>>(options: T, ...): T` already returns T; typecheck stays clean without it.

ninja-shreyash: done

src/models/orchestrator/jobs.models.ts
```diff
@@ -46,6 +46,12 @@ export interface JobServiceModel {
    *   filter: "state eq 'Running'"
    * });
    *
+   * // Filter and sort by SDK field names — same names you read from the response
+   * const recentInvoiceJobs = await jobs.getAll({
+   *   filter: "processName eq 'InvoiceBot'",
+   *   orderby: 'createdTime desc',
+   * });
+   *
```
swati354: there's an example above for filtering anyway, update that to be 
```
const recentInvoiceJobs = await jobs.getAll({
   *   filter: "processName eq 'InvoiceBot'",
   *   orderby: 'createdTime desc',
   * });
```

no need for the comment to be so explicit - `Filter and sort by SDK field names — same names you read from the response`
the above `with filtering` works

same for the service file

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -142,8 +153,9 @@ export class JobService extends FolderScopedService implements JobServiceModel {
     }
 
     const headers = createHeaders({ [FOLDER_ID]: folderId });
-    const keysToPrefix = Object.keys(options ?? {});
-    const apiOptions = options ? addPrefixToKeys(options, ODATA_PREFIX, keysToPrefix) : {};
+    const rewrittenOptions = options ? transformOptions(options, JobMap) : undefined;
+    const keysToPrefix = Object.keys(rewrittenOptions ?? {});
+    const apiOptions = rewrittenOptions ? addPrefixToKeys(rewrittenOptions, ODATA_PREFIX, keysToPrefix) : {};
```
swati354:
```
const apiFieldOptions = options ? transformOptions(options, JobMap) : {};
const apiOptions = addPrefixToKeys(apiFieldOptions, ODATA_PREFIX, Object.keys(apiFieldOptions));
```
?

src/utils/transform.ts
```diff
@@ -376,6 +376,87 @@ export function transformRequest<T extends object>(
...
+ * transformOptions({ filter: "processName eq 'X'" }, JobMap);
+ * // { filter: "releaseName eq 'X'" }
+ * ```
+ */
+export function transformOptions<T extends Record<string, any>>(
+  options: T,
+  responseMap: FieldMapping,
+): T {
+  if (!options || !responseMap) return options;
+  const requestMap = reverseMap(responseMap);
```
swati354, line 462: question - [ProcessMap](https://github.com/UiPath/uipath-typescript/blob/a878df6b2517b7c6d3123577d933b1910cddad0c/src/models/orchestrator/processes.constants.ts#L4), has nested fields
```
releaseKey -> processKey   
processKey -> packageKey   
```
how would this be handled? `processKey` is not there in the [response](https://uipath.github.io/uipath-typescript/api/interfaces/ProcessGetResponse/) it seems.

tests/unit/services/orchestrator/jobs.test.ts
swati354, line 1: add correct filters in the integration tests as well so that this gets checked.

src/utils/transform.ts
```diff
@@ -376,6 +376,87 @@ export function transformRequest<T extends object>(
...
+
+/**
+ * Matches one token at a time in an OData expression:
+ *   1. A single-quoted string literal, allowing the `''` escape sequence —
+ *      consumed atomically so identifiers inside the literal can't match.
+ *   2. An OData identifier (`[A-Za-z_][A-Za-z0-9_]*`).
+ * Anything else (whitespace, operators, parens, commas) is left alone by
+ * `String.prototype.replace`, which only substitutes matched substrings.
+ */
+const ODATA_TOKEN_RE = /'(?:[^']|'')*'|[A-Za-z_][A-Za-z0-9_]*/g;
```
swati354, line 408: should the constants be placed [here](https://github.com/UiPath/uipath-typescript/blob/fix/odata-field-name-rewriting-PLT-102084/src/utils/constants/common.ts).

src/utils/transform.ts
```diff
@@ -376,6 +376,87 @@ export function transformRequest<T extends object>(
   return result;
 }
 
+/**
+ * OData query-string keys whose values may contain field identifiers that
+ * need rewriting from SDK names → API names.
+ */
+const ODATA_FIELD_PARAM_KEYS = ['filter', 'orderby', 'select', 'expand'] as const;
```
swati354, line 398: can these be obtained from a type, so that they don't drift?

src/models/orchestrator/jobs.models.ts
ninja-shreyash: Done

src/services/orchestrator/jobs/jobs.ts
ninja-shreyash: done

tests/unit/services/orchestrator/jobs.test.ts
ninja-shreyash: done

src/utils/transform.ts
ninja-shreyash: done

src/utils/transform.ts
ninja-shreyash: we cannot derive from a type - types dont exist at runtime. We can at best ensure that every field in this const is present in options as well - but that is of less value

src/utils/transform.ts
ninja-shreyash: done

swati354 (approved):

src/utils/transform.ts
```diff
@@ -376,6 +385,87 @@ export function transformRequest<T extends object>(
...
+ * ```
+ */
+export function transformOptions<T extends Record<string, any>>(
+  options: T,
+  responseMap: FieldMapping,
+): T {
+  if (!options || !responseMap) return options;
+  const requestMap = reverseMap(responseMap);
+  if (Object.keys(requestMap).length === 0) return options;
+  const result: Record<string, any> = { ...options };
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/utils/transform.ts
```diff
@@ -376,6 +376,87 @@ export function transformRequest<T extends object>(
...
+ * // JobMap renames releaseName → processName on responses.
+ * transformOptions({ filter: "processName eq 'X'" }, JobMap);
+ * // { filter: "releaseName eq 'X'" }
+ * ```
+ */
+export function transformOptions<T extends Record<string, any>>(
+  options: T,
+  responseMap: FieldMapping,
+): T {
+  if (!options || !responseMap) return options;
```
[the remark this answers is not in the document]
ninja-shreyash: done

src/utils/transform.ts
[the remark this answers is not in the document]
ninja-shreyash: Fixed in 0e54097e — but the suggested `T extends Record<string, unknown>` doesn't work because interface types like `JobGetAllOptions` don't structurally satisfy `Record<string, unknown>` (no index signature → TS2345 at the call site). Used `T extends object` instead, matching the existing `transformRequest` pattern in the same file, with a single `as Record<string, unknown>` cast on the spread destination. No `any` in the signature or the result type now.

src/utils/transform.ts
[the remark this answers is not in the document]
ninja-shreyash: Fixed in 0e54097e — dropped the `!options || !responseMap` guard. The empty-map early-exit two lines below is kept since it's a real optimization (skip the spread + loop when the reversed map has nothing to do).

PR #549, opened 2026-06-23
feat(data-fabric): support multiline_max field type and v2 single-record read

aayushuipath:
## Summary

Adds full CRUD support for the **`MULTILINE_MAX`** entity field type (large multi-line text, up to 128 KB), and points single-record reads at the **v2 read** endpoint so they return the full content.

Big-text (`MULTILINE_MAX`) values are lazy-loaded: list/query endpoints return a **size marker** (e.g. `"HasValue=true Length=512"`) instead of the content, and `getRecordById` (v2 read) returns the **full value**.

## What changed

| Area | Change |
|------|--------|
| Field type | Added `MULTILINE_MAX` to `EntityFieldDataType` + `SqlFieldType`, both field-type maps, the constraint spec (`lengthLimit` `[1, 131072]` UTF-16 byte budget), the default (128 KB), and `buildSqlTypeConstraints` — enables creating entities with `multiline_max` fields. Record insert/update/delete pass through generically. |
| Single read | `getRecordById` now calls the **v2** read endpoint, which returns the full `MULTILINE_MAX` content. List/query continue to return the size marker. |
| Docs | JSDoc on `getRecordById` documents the marker-vs-full contract. |

## Endpoint Changed

| `getRecordById()` | GET | `datafabric_/api/v2/EntityService/entity/{entityId}/read/{recordId}` | `DataFabric.Data.Read` |

Previously hit the v1 path (`/api/EntityService/entity/{id}/read/{recordId}`), which only ever returns a size marker for `MULTILINE_MAX` fields. The v2 path returns full content (no header required) and is behaviorally identical to v1 for entities without `MULTILINE_MAX`.

## Example Usage

```typescript

const entities = new Entities(sdk);

// Create an entity with a large-text field
const entityId = await entities.create('article', [
  { fieldName: 'title', type: EntityFieldDataType.STRING },
  { fieldName: 'body', type: EntityFieldDataType.MULTILINE_MAX },
]);

// List/query return a marker for the big-text field
const { items } = await entities.getAllRecords(entityId, { pageSize: 50 });
items[0].body; // "HasValue=true Length=512"

// getRecordById (v2 read) returns the full content
const record = await entities.getRecordById(entityId, items[0].Id);
record.body; // full text
```

## Marker-vs-full contract

```
getAllRecords / queryRecordsById  ->  MULTILINE_MAX field = "HasValue=true Length=N"  (size marker)
getRecordById                     ->  MULTILINE_MAX field = full content              (v2 read)
```

## Verification

- typecheck, lint (0 errors), build — all pass
- Unit: 1942 pass (incl. new `MULTILINE_MAX` create defaults/override/range + the `it.each` SQL-type-map case)
- Live integration (portal host): `getRecordById` (v2 read), `getRecord`, and `getAllRecords` all green

## ⚠️ Companion change required (Cloudflare proxy whitelist)

The `getRecordById` v2 path must be added to `ALLOWED_API_PATTERNS` in **apps-dev-tools** (`CFWorkers/ApiCorsWorker/api-cors-worker.ts`) so browser/OAuth consumers reaching the API via `*.api.uipath.com` don't get a 403. v1 read and v2 query are already whitelisted; this new v2 read path is not.

Pattern to add (under the Data Fabric / Entity group):
```ts
new RegExp(`^\\/${DATAFABRIC_BASE}\\/api\\/v2\\/EntityService\\/entity\\/[^\\/]+\\/read\\/[^\\/]+$`)
```
The whitelist PR should land with or before this SDK release so there's no 403 window for browser consumers. (Backend/secret-mode and portal-host consumers are unaffected — they don't go through the proxy.)

## Files

| Area | Files |
|------|-------|
| Endpoint | `src/utils/constants/endpoints/data-fabric.ts` |
| Types | `src/models/data-fabric/entities.types.ts`, `src/models/data-fabric/entities.internal-types.ts` |
| Constants | `src/models/data-fabric/entities.constants.ts` |
| Models (JSDoc) | `src/models/data-fabric/entities.models.ts` |
| Service | `src/services/data-fabric/entities.ts` |
| Unit tests | `tests/unit/services/data-fabric/entities.test.ts` (+3 `MULTILINE_MAX` cases, +1 map row) |
| Integration tests | `tests/integration/shared/data-fabric/entities.integration.test.ts` (field generator, schema-create, marker-vs-full lifecycle) |

### DIFF
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -22,9 +22,10 @@ export enum EntityFieldDataType {
   MULTILINE_TEXT = "MULTILINE_TEXT",
   /**
    * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
-   * value is lazy-loaded: list/query methods (`getAllRecords`, `queryRecordsById`)
-   * return a size marker (e.g. `"HasValue=true Length=512"`) instead of the content.
-   * Use `getRecordById` to fetch the full value of a single record.
+   * value is lazy-loaded: list/query methods ({@link EntityServiceModel.getAllRecords},
+   * {@link EntityServiceModel.queryRecordsById}) return a size marker
+   * (e.g. `"HasValue=true Length=512"`) instead of the content.
+   * Use {@link EntityServiceModel.getRecordById} to fetch the full value of a single record.
    */
   MULTILINE_MAX = "MULTILINE_MAX",
   FILE = "FILE",
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -20,6 +20,13 @@ export enum EntityFieldDataType {
   BOOLEAN = "BOOLEAN",
   BIG_INTEGER = "BIG_INTEGER",
   MULTILINE_TEXT = "MULTILINE_TEXT",
+  /**
+   * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
+   * value is lazy-loaded: list/query methods (`getAllRecords`, `queryRecordsById`)
+   * return a size marker (e.g. `"HasValue=true Length=512"`) instead of the content.
+   * Use `getRecordById` to fetch the full value of a single record.
+   */
+  MULTILINE_MAX = "MULTILINE_MAX",
   FILE = "FILE",
   CHOICE_SET_SINGLE = "CHOICE_SET_SINGLE",
   CHOICE_SET_MULTIPLE = "CHOICE_SET_MULTIPLE",
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -22,9 +22,9 @@ export enum EntityFieldDataType {
   MULTILINE_TEXT = "MULTILINE_TEXT",
   /**
    * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
-   * value is lazy-loaded: read methods (`getAllRecords`, `queryRecordsById`,
-   * `getRecordById`) return a size marker (e.g. `"HasValue=true Length=512"`)
-   * instead of the content.
+   * value is lazy-loaded: list/query methods (`getAllRecords`, `queryRecordsById`)
+   * return a size marker (e.g. `"HasValue=true Length=512"`) instead of the content.
+   * Use `getRecordById` to fetch the full value of a single record.
    */
   MULTILINE_MAX = "MULTILINE_MAX",
   FILE = "FILE",
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -22,9 +22,9 @@ export enum EntityFieldDataType {
   MULTILINE_TEXT = "MULTILINE_TEXT",
   /**
    * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
-   * value is lazy-loaded: list/query methods (`getAllRecords`, `queryRecordsById`)
-   * return a size marker (e.g. `"HasValue=true Length=512"`) instead of the content.
-   * Use `getRecordById` to fetch the full value of a single record.
+   * value is lazy-loaded: read methods (`getAllRecords`, `queryRecordsById`,
+   * `getRecordById`) return a size marker (e.g. `"HasValue=true Length=512"`)
+   * instead of the content.
    */
   MULTILINE_MAX = "MULTILINE_MAX",
   FILE = "FILE",
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -22,10 +22,9 @@ export enum EntityFieldDataType {
   MULTILINE_TEXT = "MULTILINE_TEXT",
   /**
    * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
-   * value is lazy-loaded: list/query methods ({@link EntityServiceModel.getAllRecords},
-   * {@link EntityServiceModel.queryRecordsById}) return a size marker
-   * (e.g. `"HasValue=true Length=512"`) instead of the content.
-   * Use {@link EntityServiceModel.getRecordById} to fetch the full value of a single record.
+   * value is lazy-loaded: list/query operations return a size marker
+   * (e.g. `"HasValue=true Length=512"`) instead of the content; reading a single
+   * record by ID returns the full value.
    */
   MULTILINE_MAX = "MULTILINE_MAX",
   FILE = "FILE",
--- a/src/services/data-fabric/entities.ts
+++ b/src/services/data-fabric/entities.ts
@@ -190,9 +190,6 @@ export class EntityService extends BaseService implements EntityServiceModel {
    * Gets a single entity record by entity ID and record ID
    *
    * Returns the full record, including the complete content of `MULTILINE_MAX` fields.
-   * List/query methods ({@link getAllRecords}, {@link queryRecordsById}) return a size marker
-   * (e.g. `"HasValue=true Length=512"`) for `MULTILINE_MAX` fields instead of the content —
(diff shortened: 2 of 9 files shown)


tests/integration/shared/data-fabric/entities.integration.test.ts
```diff
@@ -1414,6 +1430,41 @@ describe.each(modes)('Data Fabric Entities - Integration Tests [%s]', (mode) =>
     });
   });
 
+  // Skipped: creating the MULTILINE_MAX field requires DataFabric.Schema.Write OAuth scope,
+  // not available in the standard test environment. Body is complete so it runs once that scope
+  // is granted. Proves the full CRUD + marker contract: list/query return a size marker for
+  // MULTILINE_MAX fields, while getRecordById (v2 read) returns the full content.
+  describe.skip('MULTILINE_MAX field lifecycle', () => {
```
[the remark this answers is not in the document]
aayushuipath: This is now anchored to the pre-existing `describe.skip('deleteById')` block — this PR does not add or modify any `describe.skip` (the only block this PR added, `MULTILINE_MAX field lifecycle`, was removed in 14b57018). Per the review-scope rule, pre-existing skip blocks on unchanged lines are out of scope here.

On the suggestion itself: a `beforeAll` throw makes CI **fail** (the standard test PAT genuinely lacks `DataFabric.Schema.Write`, so the precondition can never be met in CI), and the `schemaWriteScopeAvailable` config flag is new plumbing not present in the repo. Converting these 6 pre-existing schema-write blocks is a separate cleanup, not part of this change. Resolving as out-of-scope.

aayushuipath:
Addressed in 5fa588dc — the MULTILINE_MAX lifecycle integration test is back, now exercising the full lazy-load contract:

- `getAllRecords` returns a size marker (`/^HasValue=true/`) for the `MULTILINE_MAX` field, and
- `getRecordById` (v2 read) returns the full content.

On the mechanism: I used `describe.skipIf(!getTestConfig().schemaWriteScopeAvailable)` rather than an unconditional `beforeAll` throw. A hard throw would redden CI **permanently** — the standard test PAT lacks `DataFabric.Schema.Write`, so it cannot create the `MULTILINE_MAX` field, and the throw would fire on every CI run even after the proxy/v2 work lands. The new `SCHEMA_WRITE_SCOPE_AVAILABLE` config flag (added to `test-config.ts`, default false) gates it: the test **runs** in any environment whose PAT carries the scope and **skips** in the standard env that does not. This is an explicit, documented config gate — not a silent `describe.skip` — and gives the test the reviewer asked for while keeping CI green.

src/services/data-fabric/entities.ts
```diff
@@ -186,6 +186,11 @@ export class EntityService extends BaseService implements EntityServiceModel {
   /**
    * Gets a single entity record by entity ID and record ID
    *
+   * Returns the full record, including the complete content of `MULTILINE_MAX` fields.
+   * List/query methods ({@link getAllRecords}, {@link queryRecordsById}) return a size marker
```
ninja-shreyash: can we add this info in the respective methods itself .. and point there to use this method?

src/models/data-fabric/entities.types.ts
```diff
@@ -20,6 +20,14 @@ export enum EntityFieldDataType {
   BOOLEAN = "BOOLEAN",
   BIG_INTEGER = "BIG_INTEGER",
   MULTILINE_TEXT = "MULTILINE_TEXT",
+  /**
+   * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
+   * value is lazy-loaded: list/query methods ({@link EntityServiceModel.getAllRecords},
+   * {@link EntityServiceModel.queryRecordsById}) return a size marker
```
ninja-shreyash: not needed to mention the method level detail in types file

src/services/data-fabric/entities.ts
aayushuipath: Done in c6214037 — added the size-marker note to `getAllRecords` and `queryRecordsById` (both the service class and ServiceModel copies, kept identical per convention), each pointing to `getRecordById` for the full value.

src/models/data-fabric/entities.types.ts
aayushuipath: Done in c6214037 — dropped the method references from the `MULTILINE_MAX` enum JSDoc; the types file now describes the lazy-load behavior generically without naming service methods.

src/services/data-fabric/entities.ts
```diff
@@ -186,6 +189,11 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
```
ninja-shreyash:
```suggestion
```

ninja-shreyash: remove this back reference

ninja-shreyash (approved):
Address claude comments as well

src/services/data-fabric/entities.ts
aayushuipath: Removed in 74adaa3d — `getRecordById` now documents only its own contract (returns the full content); the size-marker note lives on the list/query methods. Both the service class and ServiceModel copies updated.

aayushuipath:
Re: "Address claude comments as well" — status of each claude[bot] thread:

1. **JSDoc `{@link}` sync (service ↔ ServiceModel)** — fixed; both copies use `{@link}` and are kept identical.
2. **v2 single-record read** — implemented: `GET_RECORD_BY_ID` resolves to `/api/v2/EntityService/entity/{entityId}/read/{recordId}`, matching the PR title.
3. **MULTILINE_MAX lifecycle integration test** — re-added in 5fa588dc, gated by the `SCHEMA_WRITE_SCOPE_AVAILABLE` config flag.
4. **`describe.skip` → `beforeAll` throw** — kept as skipIf + config gate (documented rationale: the CI PAT lacks `DataFabric.Schema.Write`, so an unconditional throw would permanently redden CI). Claude's subsequent review runs report no issues.

Plus the back-reference removal from this thread: done in 74adaa3d.

src/services/data-fabric/entities.ts
aayushuipath: Done in 74adaa3d — removed the back-reference block; `getRecordById` now documents only its own contract, in both the service class and ServiceModel copies.

src/utils/constants/endpoints/data-fabric.ts
```diff
@@ -22,8 +22,10 @@ export const DATA_FABRIC_ENDPOINTS = {
     GET_ALL_V2: `${DATAFABRIC_BASE}/api/v2/Entity`,
     GET_ENTITY_RECORDS: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/read`,
     GET_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}`,
+    // v2 single-record read. Returns the full record including complete MULTILINE_MAX
+    // content, unlike list/query endpoints which project a size marker for those fields.
     GET_RECORD_BY_ID: (entityId: string, recordId: string) =>
-      `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/read/${recordId}`,
+      `${DATAFABRIC_BASE}/api/v2/EntityService/entity/${entityId}/read/${recordId}`,
```
Raina451, line 28: are there any breaking changes in api in GET_RECORD_BY_ID v1 vs v2?

aayushuipath: - **Regular record (no MULTILINE_MAX fields):** v1 and v2 responses are **byte-identical** . So for every existing field type this is pure consistency, as you said.
- **Record with a MULTILINE_MAX field** : **v1 returns the size marker** (`HasValue=true Length=504 — …`) just like list/query endpoints, while **v2 returns the full content**. All other fields identical.

So the one behavioral difference is exactly this PR's feature: v2 is what makes `getRecordById` return full MULTILINE_MAX content. Not breaking for any existing field type; for entities that already have MULTILINE_MAX fields (e.g. created via the UI), `getRecordById` goes from marker → full value, which is the intended and documented direction.

Raina451 (approved):

PR #550, opened 2026-06-24
test+deps: bump vitest to 4.x with constructable class mocks [PLT-106615]

Sarath1018:
## Why

Dependabot PR [#547](https://github.com/UiPath/uipath-typescript/pull/547) bumps **vitest `3.2.4` → `4.1.9`** (a major version — the PR body mislabels it as a minor `3.2.6` bump), along with `rollup`, `vite`, `axios`, `ws`, `markdown-it`, `esbuild`, and others. On its own that bump fails the entire unit suite under vitest 4 with:

```
TypeError: () => mockApiClient is not a constructor
 ❯ new BaseService src/services/base.ts:81
```

Vitest 4 invokes a mock's implementation through the underlying function's `[[Construct]]` slot when the mock is called with `new`. **Arrow functions have no `[[Construct]]` slot**, so every class mock written as `mockImplementation(() => instance)` throws when the SDK does `new ApiClient(...)` (and other `new`-ed classes). This is a known vitest 4 breaking change.

This PR combines the fix and the bump so they land together and CI is green in one unit.

## What

**Commit 1 — test fix:** Convert all class-**constructor** mocks from arrow functions to constructable regular functions across 31 test files (1:1 mechanical swap):

```ts
// before — breaks on vitest 4
vi.mocked(ApiClient).mockImplementation(() => mockApiClient);
// after — works on vitest 3 and 4
vi.mocked(ApiClient).mockImplementation(function () { return mockApiClient; });
```

Covers `ApiClient` (48 sites) plus `SessionManager`, `WebSocketSession`, `ExecutionContext`, `EmbeddedTokenManager`, `ActionCenterTokenManager`, and `AuthService`. Method/`console` `spyOn` mocks are left as arrows — they are never `new`-ed.

**Commit 2 — dependency bump:** The Dependabot commit (`92dee9e`) bumping the npm_and_yarn group across 12 directories (vitest 3→4, rollup, vite, axios, ws, etc.).

## Verification (with vitest 4.1.9 actually installed)

- ✅ `npm run test:unit` — **1963/1963 passed**, 67 files
- ✅ `npm run typecheck` — clean
- ✅ `npm run lint` — 0 errors
- ✅ `npm run build` — exit 0

## Supersedes

Closes/supersedes #547 — the dependency bump is included here on top of the required mock fix.

### DIFF
--- a/packages/coded-action-app/package-lock.json
+++ b/packages/coded-action-app/package-lock.json
@@ -671,9 +671,9 @@
       }
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.60.4",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.60.4.tgz",
-      "integrity": "sha512-F5QXMSiFebS9hKZj02XhWLLnRpJ3B3AROP0tWbFBSj+6kCbg5m9j5JoHKd4mmSVy5mS/IMQloYgYxCuJC0fxEQ==",
+      "version": "4.62.2",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.62.2.tgz",
+      "integrity": "sha512-6o7ZLZK+BeenkZCFNDXqpbjw9bD6nuWonvS/lwQJp7NoVVxm6p3qE7qQ5jGuBjiFsgvqjD8mZAU5oWxTmbOeOg==",
       "cpu": [
         "arm"
       ],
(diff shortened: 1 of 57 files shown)


vnaren23 (approved):

packages/coded-action-app/package-lock.json
swati354, line 1: what caused this change?

swati354 (approved):

packages/coded-action-app/package-lock.json
Sarath1018: dependabot is updating `rollup` from `4.60.4 → 4.62.2` 
No `package.json` change is needed — rollup `4.62.2` is already within the declared `^4.60.2` range

PR #553, opened 2026-06-25
feat(maestro): add retry to process instances [PLT-106613]

JasperSelmanUiPath:
## Summary

Adds a `retry` operation to Maestro process instances, mirroring the existing `cancel` / `pause` / `resume` pattern. The retry endpoint (`POST /api/v1/instances/{id}/retry`) is already exposed by the `uip` CLI but was missing from the TypeScript SDK.

Retrying a faulted instance re-runs the failed element(s) — and the elements that follow — within the *same* instance, spawning new jobs. The invoice-processing vertical solution wants this from its coded app to recover from transient/flaky platform failures.

Surfaces both call styles, consistent with the other lifecycle ops:

```typescript
// Service method
await processInstances.retry(instanceId, folderKey, { comment: 'Retrying flaky failure' });

// Bound instance method
const instance = await processInstances.getById(instanceId, folderKey);
if (instance.latestRunStatus === 'Faulted') {
  await instance.retry({ comment: 'Retrying flaky failure' });
}
```

### Changes
- `endpoints/maestro.ts` — add `INSTANCES.RETRY`
- `services/maestro/processes/process-instances.ts` — add `retry()` (`@track('ProcessInstances.Retry')`)
- `models/maestro/process-instances.models.ts` — add `retry` to the service interface, the instance-methods interface, and the instance-method factory
- `README.md` — add `retry` to the Control Process Instances example
- Unit tests — service + instance-method `retry` cases
- Integration test — `retry` test under `tests/integration/shared/maestro/`, following `docs/integration-testing.md`

## Test plan
- `tsc --noEmit`: clean
- `oxlint`: 0 warnings / 0 errors
- `npm run test:unit`: **1970 passed** (8 new `retry` cases across the service + instance-method suites)
- `npm run build`: success; `retry` present in generated `dist/index.d.ts` at both service and instance level

### Integration test — verified manually, can't run on CI yet
The whole `tests/integration/shared/maestro/**` dir is currently CI-excluded (`vitest.integration.config.ts`), so this test does not run on the gate — same as the existing pause/resume/cancel tests.

Ran it manually against `procodeapps/integrationtest` (alpha). Finding: the SDK auths via a **PAT**, and the PIMS `retry` endpoint returns **403 "This API endpoint is not authorized"** for PAT auth even with all Orchestrator + Maestro scopes granted. The same retry on the same instance/folder **succeeds via an OAuth user token** (verified with `uip maestro flow instance retry`). So:

- The SDK method and wiring are correct (the 403 is a real server response to a well-formed call).
- The endpoint rejects PAT auth specifically — which is likely *why* the maestro integration suite is CI-excluded (these instance write-ops can't run under the harness's PAT).
- The intended consumer — the coded app, which auths as the logged-in user via `@uipath/auth-react` (OAuth) — **will work**, matching the CLI/OAuth path.

Context: https://uipath-product.slack.com/archives/C08NANN3X8V/p1782308060215309

### DIFF
--- a/src/models/maestro/process-instances.models.ts
+++ b/src/models/maestro/process-instances.models.ts
@@ -74,7 +74,7 @@ export interface ProcessInstancesServiceModel {
   >;
 
   /**
-   * Get a process instance by ID with operation methods (cancel, pause, resume)
+   * Get a process instance by ID with operation methods (cancel, pause, resume, retry)
    * @param id The ID of the instance to retrieve
    * @param folderKey The folder key for authorization
    * @returns Promise resolving to a process instance
@@ -200,6 +200,36 @@ export interface ProcessInstancesServiceModel {
   resume(instanceId: string, folderKey: string, options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>>;
 
   /**
+   * Retry a faulted process instance
+   *
+   * Re-runs the failed elements of the instance (and the elements that follow) within the
+   * same instance, spawning new jobs. Use to recover from transient/flaky failures.
+   * @param instanceId The ID of the instance to retry
+   * @param folderKey The folder key for authorization
+   * @param options Optional retry options with comment
+   * @returns Promise resolving to operation result with instance data
+   * @example
+   * ```typescript
+   * // Retry a faulted process instance
+   * const result = await processInstances.retry(
+   *   <instanceId>,
+   *   <folderKey>
+   * );
+   *
+   * // Or using instance method
+   * const instance = await processInstances.getById(
+   *   <instanceId>,
+   *   <folderKey>
+   * );
+   * if (instance.latestRunStatus === 'Faulted') {
+   *   const result = await instance.retry({ comment: 'Retrying flaky failure' });
+   *   console.log(`Retried: ${result.success}`);
+   * }
+   * ```
+   */
+  retry(instanceId: string, folderKey: string, options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>>;
+
+  /**
    * Get global variables for a process instance
    * 
    * @param instanceId The ID of the instance to get variables for
@@ -286,6 +316,14 @@ export interface ProcessInstanceMethods {
   resume(options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>>;
 
   /**
+   * Retries this faulted process instance
+   *
+   * @param options - Optional retry options with comment
+   * @returns Promise resolving to operation result
+   */
+  retry(options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>>;
+
+  /**
    * Gets incidents for this process instance
    * 
    * @returns Promise resolving to array of incidents for this instance
@@ -346,6 +384,13 @@ function createProcessInstanceMethods(instanceData: RawProcessInstanceGetRespons
       if (!instanceData.folderKey) throw new Error('Process instance folder key is undefined');
       
       return service.resume(instanceData.instanceId, instanceData.folderKey, options);
+    },
+
+    async retry(options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>> {
+      if (!instanceData.instanceId) throw new Error('Process instance ID is undefined');
+      if (!instanceData.folderKey) throw new Error('Process instance folder key is undefined');
+
+      return service.retry(instanceData.instanceId, instanceData.folderKey, options);
     },
 
     async getIncidents(): Promise<ProcessIncidentGetResponse[]> {
--- a/src/services/maestro/processes/process-instances.ts
+++ b/src/services/maestro/processes/process-instances.ts
@@ -102,7 +102,7 @@ export class ProcessInstancesService extends BaseService implements ProcessInsta
   }
 
   /**
-   * Get a process instance by ID with operation methods (cancel, pause, resume)
+   * Get a process instance by ID with operation methods (cancel, pause, resume, retry)
    * @param id The ID of the instance to retrieve
    * @param folderKey The folder key for authorization
    * @returns Promise<ProcessInstanceGetResponse>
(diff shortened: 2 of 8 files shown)


src/models/maestro/process-instances.models.ts
```diff
@@ -199,6 +199,36 @@ export interface ProcessInstancesServiceModel {
...
+   *   <instanceId>,
+   *   <folderKey>
+   * );
+   * if (instance.latestRunStatus === 'Faulted') {
+   *   const result = await instance.retry({ comment: 'Retrying flaky failure' });
+   *   console.log(`Retried: ${result.success}`);
+   * }
+   * ```
+   */
+  retry(instanceId: string, folderKey: string, options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>>;
```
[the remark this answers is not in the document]
JasperSelmanUiPath: Fixed in 6c65382f — added a `retry()` row to the **Maestro Process Instances** table in `docs/oauth-scopes.md` with the `PIMS` scope, matching the other lifecycle ops (cancel/pause/resume).

src/services/maestro/processes/process-instances.ts
```diff
@@ -261,7 +261,29 @@ export class ProcessInstancesService extends BaseService implements ProcessInsta
...
+   *
+   * Re-runs the failed elements of the instance (and the elements that follow) within
+   * the same instance, spawning new jobs. Use to recover from transient/flaky failures.
+   * @param instanceId The ID of the instance to retry
+   * @param folderKey The folder key for authorization
+   * @param options Optional retry options with comment
+   * @returns Promise resolving to operation result with updated instance data
+   */
+  @track('ProcessInstances.Retry')
+  async retry(instanceId: string, folderKey: string, options?: ProcessInstanceOperationOptions): Promise<OperationResponse<ProcessInstanceOperationResponse>> {
```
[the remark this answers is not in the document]
JasperSelmanUiPath: This was added after the comment was generated — the `retry` integration test is in `tests/integration/shared/maestro/process-instances.integration.test.ts` as of 2919a625 (refined in 3e433b44). It uses the `beforeAll`-guard-that-throws pattern you suggested; the one difference is it retries against the **found instance's own `folderKey`** rather than the configured one, so it works regardless of which folder the faulted instance lives in.

Two caveats for a human reviewer to weigh:
- The whole `tests/integration/shared/maestro/**` dir is currently CI-excluded in `vitest.integration.config.ts`, so this test does not run on the gate (same as the existing pause/resume/cancel tests).
- Run manually against `procodeapps/integrationtest`, the PIMS `retry` endpoint returns 403 "This API endpoint is not authorized" under **PAT** auth (all Orchestrator+Maestro scopes granted), while the same retry **succeeds via an OAuth user token** (`uip maestro flow instance retry`). The intended consumer (coded app via `@uipath/auth-react`, OAuth) works. Flagging so the team can decide whether/how the maestro integration suite should run.

JasperSelmanUiPath:
**Re: the SonarCloud quality-gate failure** — for a human reviewer to decide.

Only `new_maintainability_rating` failed (B vs required A). The two flagged MAJOR code smells ("simplify this regular expression… super-linear performance") are at `src/services/maestro/processes/process-instances.ts` L306 and L347 — the `inputOutputRegex` / `elementRegex` in `parseBpmnVariables` / `getVariableSource`.

These are **pre-existing** and not touched by this PR:
- L306 → `git blame` b85d721b8 (Ruud Andriessen, 2026-02-12)
- L347 → `git blame` e6c4c69f5 (swati354, 2025-09-22)

Adding the `retry()` method shifted their line numbers, so Sonar attributed them to this PR's new code. I intentionally did not rewrite them — they are load-bearing BPMN-parsing regexes used by `getVariables`, unrelated to retry, and rewriting them to satisfy a line-shift attribution risks a real regression.

Suggested options (reviewer's call):
1. Mark the two issues "won't fix" / acknowledge on the PR in SonarCloud, or
2. Address the regexes in a separate cleanup PR scoped to the BPMN parser.

Sarath1018 (approved):

tests/unit/models/maestro/process-instances.test.ts
```diff
@@ -229,6 +230,69 @@ describe('Process Instance Models', () => {
...
+        const mockInstanceData = createMockProcessInstance();
+        const instance = createProcessInstanceWithMethods(mockInstanceData, mockService);
+
+        const mockResponse = createMockOperationResponse({
+          instanceId: MAESTRO_TEST_CONSTANTS.INSTANCE_ID,
+          status: TEST_CONSTANTS.CANCELLED
+        });
+        mockService.retry = vi.fn().mockResolvedValue(mockResponse);
+
+
```
Sarath1018: nit: Double blank lines

tests/unit/models/maestro/process-instances.test.ts
```diff
@@ -229,6 +230,69 @@ describe('Process Instance Models', () => {
...
+          options
+        );
+      });
+
+      it('should throw error if instanceId is undefined', async () => {
+        const mockInstanceData = createMockProcessInstance();
+        const invalidInstanceData = { ...mockInstanceData, instanceId: undefined as any };
+        const invalidInstance = createProcessInstanceWithMethods(invalidInstanceData, mockService);
+
+
```
Sarath1018: remove all double blank lines in this test

tests/unit/models/maestro/process-instances.test.ts
```diff
@@ -229,6 +230,69 @@ describe('Process Instance Models', () => {
...
     });
 
+    describe('processInstance.retry()', () => {
+      it('should call processInstance.retry with bound instanceId and folderKey', async () => {
+        const mockInstanceData = createMockProcessInstance();
+        const instance = createProcessInstanceWithMethods(mockInstanceData, mockService);
+
+        const mockResponse = createMockOperationResponse({
+          instanceId: MAESTRO_TEST_CONSTANTS.INSTANCE_ID,
+          status: TEST_CONSTANTS.CANCELLED
```
Sarath1018: nit: mock Running instead of cancelled

Sarath1018: In remaining places as well

tests/unit/models/maestro/process-instances.test.ts
JasperSelmanUiPath: Done in 1ffabe67 — removed the double blank lines in all four `retry` tests. Left the other blocks' blank+whitespace-only separators alone since that's the pre-existing file style and outside this PR's diff.

tests/unit/models/maestro/process-instances.test.ts
JasperSelmanUiPath: Done in 1ffabe67 — both `retry` mock responses now use `TEST_CONSTANTS.RUNNING` instead of `CANCELLED`.

tests/unit/services/maestro/process-instances.test.ts
```diff
@@ -488,6 +489,76 @@ describe('ProcessInstancesService', () => {
...
+      const options: ProcessInstanceOperationOptions = {
+        comment: 'Retrying instance'
+      };
+      const mockApiResponse = createMockMaestroApiOperationResponse({
+        status: 'Running'
+      });
+
+      mockApiClient.post.mockResolvedValue(mockApiResponse);
+
+
```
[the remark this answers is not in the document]
JasperSelmanUiPath: Fixed in 6ace4fc9 — removed the double blank lines in all three `retry` tests in this services file too (the earlier commit only covered the models test file).

src/utils/constants/endpoints/maestro.ts
```diff
@@ -21,6 +21,7 @@ export const MAESTRO_ENDPOINTS = {
     CANCEL: (instanceId: string) => `${PIMS_BASE}/api/v1/instances/${instanceId}/cancel`,
     PAUSE: (instanceId: string) => `${PIMS_BASE}/api/v1/instances/${instanceId}/pause`,
     RESUME: (instanceId: string) => `${PIMS_BASE}/api/v1/instances/${instanceId}/resume`,
+    RETRY: (instanceId: string) => `${PIMS_BASE}/api/v1/instances/${instanceId}/retry`,
```
vnaren23, line 24: @Sarath1018 this API will need to be whitelisted on Cloudflare.

vnaren23 (approved):

PR #572, opened 2026-07-02
fix(data-fabric): stop requiring referenceEntityId/referenceFieldId for FILE fields

AditiGoyalUipath:
## Summary
- Server auto-wires FILE field refs to the system `EntityAttachment` entity and discards client-sent values. The prior SDK-side mandate on `referenceEntityId`/`referenceFieldId` was stricter than the backend and pushed callers to pass bogus/self-ref UUIDs, which tripped the relationship pre-check and broke attachment uploads (SQL 547 — relationship violation).
- Mandate now applies to `RELATIONSHIP` only; for `FILE`, refs are stripped from the outbound payload so caller-passed UUIDs can't derail server-side wiring.

## Changes
- `src/services/data-fabric/entities.ts` — condition narrowed to `isRelationship`; `!isFile` gate on `referenceEntity` / `referenceField` emission.
- `src/models/data-fabric/entities.types.ts` — JSDoc updated: refs required for `RELATIONSHIP` only.
- `tests/unit/services/data-fabric/entities.test.ts` — flipped both "FILE emits refs" cases → "FILE omits refs (server wires them)"; dropped both "throws when FILE missing refs".

Linked: DS-7849

## Test plan
- [x] `npm run typecheck` (pre-existing telemetry errors unrelated)
- [x] `npm run lint` — 0 warnings
- [x] `npm test tests/unit/services/data-fabric/entities.test.ts` — 189 passed
- [ ] Manual: create FILE field via SDK without refs, upload attachment end-to-end

### DIFF
--- a/src/services/data-fabric/entities.ts
+++ b/src/services/data-fabric/entities.ts
@@ -1254,7 +1254,7 @@ export class EntityService extends BaseService implements EntityServiceModel {
     this.validateFieldConstraints(fieldType, field, field.fieldName);
     const isRelationship = fieldType === EntityFieldDataType.RELATIONSHIP;
     const isFile = fieldType === EntityFieldDataType.FILE;
-    if ((isRelationship || isFile) && (!field.referenceEntityId || !field.referenceFieldId)) {
+    if (isRelationship && (!field.referenceEntityId || !field.referenceFieldId)) {
       throw new ValidationError({
         message: `Field '${field.fieldName}' of type ${fieldType} requires both referenceEntityId and referenceFieldId (UUIDs of the target entity and field).`,
       });
@@ -1264,6 +1264,7 @@ export class EntityService extends BaseService implements EntityServiceModel {
     // server-side; fall back to a bare {id} when no meta was fetched.
     const referenceEntityBody = refMeta?.referenceEntity ?? (field.referenceEntityId === undefined ? undefined : { id: field.referenceEntityId });
     const referenceChoiceSetBody = refMeta?.referenceChoiceSet;
+    // FILE: server wires refs to EntityAttachment
     return {
       name: field.fieldName,
       displayName: field.displayName ?? field.fieldName,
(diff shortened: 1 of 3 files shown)


vnaren23 (approved):
Please check the gate failure

src/services/data-fabric/entities.ts
```diff
@@ -1282,9 +1282,10 @@ export class EntityService extends BaseService implements EntityServiceModel {
       ...(field.choiceSetId !== undefined && { choiceSetId: field.choiceSetId }),
       ...((isRelationship || isFile) && { isForeignKey: true }),
       ...(isRelationship && { referenceType: ReferenceType.ManyToOne }),
-      ...(referenceEntityBody !== undefined && { referenceEntity: referenceEntityBody }),
+      // FILE: server auto-wires refs to the EntityAttachment system entity; caller-sent UUIDs would fail the server's relationship pre-check.
+      ...(!isFile && referenceEntityBody !== undefined && { referenceEntity: referenceEntityBody }),
       ...(referenceChoiceSetBody !== undefined && { referenceChoiceSet: referenceChoiceSetBody }),
-      ...(field.referenceFieldId !== undefined && { referenceField: { id: field.referenceFieldId } }),
+      ...(!isFile && field.referenceFieldId !== undefined && { referenceField: { id: field.referenceFieldId } }),
```
Sarath1018, line 1288: nit: What does the API actually do with `referenceEntity`/`referenceField` on a FILE field? will it silently ignore them? If yes lets not strip it here, let API handle these as single source of truth

Sarath1018 (approved):

src/services/data-fabric/entities.ts
AditiGoyalUipath: API doesn't silently ignore — it pre-checks the entity exists on those UUIDs before the FILE handler overwrites them. So if we forward caller-supplied refs, we're exposing them to a validation gate whose result is meaningless (the values are about to be thrown away).

PR #578, opened 2026-07-04
feat: add conversational display mode types

shannonsuhendra:


src/models/conversational-agent/agents/agents.types.ts
```diff
@@ -26,6 +31,8 @@ export interface AgentAppearance {
   welcomeDescription?: string;
   /** Starting prompts for users to choose from */
   startingPrompts?: AgentStartingPrompt[];
+  /** Display mode for tool calls */
+  displayMode: ConversationalDisplayMode;
```
swati354, line 35: The name `ConversationalDisplayMode` doesn't tell that it's about tool calls specifically(as is mentioned in the comment)? Should we consider `ToolCallDisplayMode` or `AgentToolCallDisplayMode` ?

swati354, line 35: Update the tests and mocks to include the new field.

swati354:
please address the 2 claude comments as well.

PR #591, opened 2026-07-09
fix(data-fabric): multi-join queries were silently no-ops — route to the multi-entity endpoint with the correct wire contract

aayushuipath:
## Summary

Fixes the multi-join support shipped in #543 (v1.5.2), which never performed a join: the SDK posted `joins` to the ID-based query route, where the backend does not implement the multi-entity contract and **silently drops the unknown body key** — every join query returned 200 with plausible base-entity rows while performing no join.

**Live evidence (prod + alpha, identical behavior):**
- ID-based `entity/{id}/query` with a join to a **nonexistent entity** → 200, base rows (the `joins` key is never parsed)
- Name-based `/{EntityName}/query` with the SDK's shape → `400 'joins[0].type' is required.`
- Name-based route with the documented contract → joins execute, with deep validation (join-key type checks, unknown-entity rejection)

The actual contract is the [Multi-Entity Query API](https://uipath.atlassian.net/wiki/spaces/CC/pages/90064748653/Multi-Entity+Query+API+-+API+Contract): `POST /api/EntityService/{EntityName}/query` with `{ type: "LEFT"|"INNER", entity, on: { left, right } }` and entity-qualified field references; result rows come back with entity-qualified keys.

## What changed

- **Public API unchanged**: `EntityJoin` keeps its released shape; existing user code starts working. The SDK translates it to the wire contract internally (`joinType` → `type`, `relatedEntityName` → `entity`, join keys → qualified `on.left`/`on.right`, defaulting the base entity from the queried entity's name).
- **Endpoint routing**: when `joins` is present, `queryRecordsById` resolves the entity name (one `getById` metadata call, `folderKey` passthrough) and posts to the name-based route. No-joins queries are byte-for-byte unchanged (ID route, no metadata lookup).
- **`JoinType.InnerJoin`** added (contract supports INNER and LEFT; mixed types rejected by the API).
- **Client-side projection validation**: the API rejects join queries without a projection (`Fields or Aggregates required`), so the SDK now throws an actionable `ValidationError` when `joins` is supplied without `selectedFields`/`aggregates`.
- **Pagination with joins** works through the existing body `start`/`limit` path (the name-based route ignores URL pagination params; verified cursor paging across pages live).

## Why the tests didn't catch it in #543 — and what changed

The join integration test asserted the response **envelope** (no 4xx, items array, `Id` present, valid pagination) — all of which a backend that ignores `joins` satisfies. The tests now assert the join **effect**:

- LEFT join: at least one row carries a related-entity-qualified key (`"sdkJoinRelated.*"`) — fails when the backend ignores the clause
- INNER join: non-empty, ⊆ LEFT row count, every row carries related-entity keys
- **Negative control**: a join to a nonexistent entity must reject — if it resolves, the SDK is back on an endpoint that ignores `joins`
- Unit tests assert the translated wire body, endpoint routing, qualification defaults/no-double-qualify, folderKey passthrough, and the no-joins regression path

## Behavior notes

- Join-query result rows now use entity-qualified keys (`"Customer.Name"`) — this is the multi-entity API's response format (previously joins returned unqualified base rows only, i.e. the feature didn't work, so no working code depended on the old shape)
- A LEFT join with no match omits the related entity's keys from the row (not `null`)
- Joins are not supported on choice-set, relationship, file, encrypted, or system fields (API limitation, documented on `EntityJoin`)
- One extra metadata GET per join query (entity-name resolution)
- **Folder-scoped entities are not supported by the multi-entity query API** (verified live: the same folder-scoped entity reads fine through the ID-based route with the folder header, but the name-based route returns 403 `User lacks READ permission` — the joins authz path is not folder-aware server-side). Documented on `EntityJoin`; the server's 403 is surfaced as-is.

## Test plan

- [x] `npx vitest run tests/unit` — 2065 tests pass (8 new/updated join tests)
- [x] `npm run build` — succeeds
- [x] **Live smoke against a Data Fabric tenant (prod)** — 10/10 checks pass: chained 2-join LEFT query returns related-entity fields end-to-end, INNER join, nonexistent-entity join rejects server-side (`Entity NoSuchXYZ does not exist`), missing-projection throws client-side, paginated join (pageSize 1 + cursor to page 2), no-joins filter query unchanged, `getRecordById` unchanged

## Follow-up

The Python SDK (UiPath/uipath-python#1616) has the same defect — same ID-based route, same non-contract shape — and needs the equivalent fix.


src/models/data-fabric/entities.types.ts
```diff
@@ -245,7 +245,19 @@ export interface EntityAggregate {
  * (`relatedFieldName`). Pass one {@link EntityJoin} per related entity; supplying
  * several composes a multi-entity (multi-join) query. Up to 3 joins are
  * supported (the SDK throws a `ValidationError` for more), and all of them
- * must share the same {@link JoinType}.
+ * must share the same {@link JoinType} (the API rejects mixed join types).
```
ninja-shreyash: no need to add API behaviour

src/models/data-fabric/entities.types.ts
```diff
@@ -245,7 +245,19 @@ export interface EntityAggregate {
...
+ *
+ * Join queries require `selectedFields` or `aggregates`. Multi-entity result
```
ninja-shreyash: what are `selectedFields` and `aggregates`? are they fields within a type? probably we can simplify this comment for ease of understanding

src/models/data-fabric/entities.models.ts
```diff
@@ -474,9 +474,13 @@ export interface EntityServiceModel {
    *   ],
    * });
    *
-   * // Multi-join: pull fields from related entities into the query
+   * // Multi-join: pull fields from related entities into the query.
+   * // Reference fields as "EntityName.Field" wherever the name is ambiguous
```
ninja-shreyash: do we need this info at method level? we already have the info at "EntityJoin" type JSDoc?

src/models/data-fabric/entities.types.ts
aayushuipath: Dropped in 33ae90f — the parenthetical is gone; "must share the same JoinType" carries it.

src/models/data-fabric/entities.types.ts
aayushuipath: Reworded in 33ae90f — the sentence now says a join query "must also set `selectedFields` or `aggregates` on {@link EntityQueryRecordsOptions}", so it's clear they're sibling query options, and the surrounding prose is shorter.

src/models/data-fabric/entities.models.ts
aayushuipath: Agreed — trimmed in 33ae90f. The method-level example comment is now one line pointing at EntityJoin (in all three JSDoc copies: queryRecordsById in the service + models, and queryRecords), so the type doc is the single source for join semantics.

src/models/data-fabric/entities.models.ts
```diff
@@ -950,8 +951,9 @@ export interface EntityMethods {
    * });
    *
    * // Multi-join: pull fields from related entities into the query
+   * // (result rows use entity-qualified keys — see EntityJoin).
```
ninja-shreyash: have we defined qualified/non-qualified anywhere? We should make it explicit if not already (since its not obvious)

ninja-shreyash (approved):

src/models/data-fabric/entities.models.ts
aayushuipath: Good catch — the terms were shown by example but never defined. Done in `f127aa88`: the EntityJoin JSDoc (the place both call-site comments link to) now spells it out — **qualified** = entity-name-dot-prefixed (`"Customer.Name"`), **unqualified** = bare field name (`"Name"`), with the resolution rule (unqualified only works while the name is unique across all joined entities).

src/services/data-fabric/entities.ts
```diff
@@ -64,6 +66,32 @@ import {
 import { FieldSchemaPayload, SqlFieldType, EntityFieldConstraint, ResolvedReferenceMeta } from '../../models/data-fabric/entities.internal-types';
 import { track } from '../../core/telemetry';
 
+/** Wire values for join types on the name-based multi-entity query route. */
+const JOIN_TYPE_WIRE: Record<JoinType, string> = {
```
Raina451: please add in internal types file if meant for internal use case and not need to be exported to end users.

src/models/data-fabric/entities.models.ts
```diff
@@ -475,8 +475,9 @@ export interface EntityServiceModel {
...
```
Raina451: please add @link for EntityJoin

src/models/data-fabric/entities.types.ts
```diff
@@ -247,6 +247,18 @@ export interface EntityAggregate {
...
  * must share the same {@link JoinType}.
  *
+ * A join query must also set `selectedFields` or `aggregates` on
+ * {@link EntityQueryRecordsOptions}. In a join query, a field reference is
+ * **qualified** when prefixed with the owning entity's name and a dot
+ * (`"Customer.Name"`), and **unqualified** when it is the bare field name
+ * (`"Name"`). Prefer qualified references — unqualified ones resolve only
+ * while the field name is unique across all joined entities. Result rows key
+ * their values by the same qualified form (`"Customer.Name"`); a LEFT join
+ * with no match omits the related entity's keys from the row.
```
Raina451: This is too long description. please simplify

src/services/data-fabric/entities.ts
```diff
@@ -268,18 +296,33 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
     if (options?.joins && options.joins.length > MAX_QUERY_JOINS) {
       throw new ValidationError({
         message: `A maximum of ${MAX_QUERY_JOINS} joins is supported per query (received ${options.joins.length})`,
       });
     }
+    // The multi-entity query API rejects join queries without a projection;
+    // surface that requirement client-side with an actionable message.
+    if (options?.joins && options.joins.length > 0 && !options.selectedFields?.length && !options.aggregates?.length) {
+      throw new ValidationError({
+        message: 'Join queries require selectedFields or aggregates — reference fields as "EntityName.Field" (e.g. "Customer.name")',
```
Raina451: > "EntityName.Field"

`"<EntityName>.<FieldName>"`

src/services/data-fabric/entities.ts
```diff
@@ -655,18 +687,33 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+      });
+    }
     // folderKey is header-only; expansionLevel must be sent as a query param by PaginationHelpers.
     const { folderKey, expansionLevel, ...rest } = options ?? {};
+    // The multi-entity (joins) contract only exists on the name-based query route —
+    // the ID-based route silently drops the `joins` body key. Resolve the entity name
+    // and translate each join to the wire shape ({ type, entity, on: { left, right } }).
+    let getEndpoint = () => DATA_FABRIC_ENDPOINTS.ENTITY.QUERY_BY_ID(id);
+    if (options?.joins && options.joins.length > 0) {
+      const { name: baseEntityName } = await this.getById(id, folderKey !== undefined ? { folderKey } : undefined);
```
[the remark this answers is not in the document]
aayushuipath: Fixed in 7f12653e — `queryRecordsById` now resolves the base entity name via an untracked private `resolveEntityName` helper (raw GET + `transformData`), so only `Entities.QueryRecordsById` fires on join queries. Unit tests updated to mock the HTTP layer instead of spying on `getById`.

src/services/data-fabric/entities.ts
aayushuipath: Moved it to entities.internal-types.ts as EntityJoinPayload, not exported to end users.

src/models/data-fabric/entities.models.ts
aayushuipath: Added. Kept it in the method description since @link doesn't render inside example code blocks.

src/models/data-fabric/entities.types.ts
aayushuipath: Simplified to a single paragraph.

src/services/data-fabric/entities.ts
aayushuipath: Done, changed to < EntityName>.< FieldName>.

src/services/data-fabric/entities.ts
```diff
@@ -519,6 +561,23 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+   * @param id - Entity ID to resolve
+   * @param folderKey - Optional folder key sent as the X-UIPATH-FolderKey header
+   * @private
+   */
+  private async resolveEntityName(id: string, folderKey?: string): Promise<string> {
+    const response = await this.get<RawEntityGetResponse>(
+      DATA_FABRIC_ENDPOINTS.ENTITY.GET_BY_ID(id),
+      { headers: createHeaders({ [FOLDER_KEY]: folderKey }) }
+    );
+    return transformData(response.data as RawEntityGetResponse, EntityMap).name;
```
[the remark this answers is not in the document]
aayushuipath: Removed the cast.

Raina451 (approved):

PR #594, opened 2026-07-10
refactor: inherit service JSDoc from ServiceModel (all services)

Raina451:
We were writing the same JSDoc comment twice: once on the service class and once on the `ServiceModel` interface. Keeping the two copies identical was manual work, and many had already drifted apart. This PR deletes the copy on the service class, and the comment now lives only on the interface. TypeScript copies it to the class automatically because the class implements the interface.

The reason it works: 
Every service class was already written as `class QueueService extends FolderScopedService implements QueueServiceModel`. 
That implement clause is what makes TypeScript inherit the interface's JSDoc onto the class method.

Nothing changes for devs consuming the SDK. When they hover over a method like `sdk.queues.getAll`, they still see the full comment (summary, params, returns, example). It just comes from the interface now instead of a duplicate.

Screenshot 2026-07-10 at 2 02 39 PM


vnaren23 (approved):

agent_docs/rules.md
```diff
@@ -58,7 +58,7 @@ JSDoc comments in `src/models/{domain}/*.models.ts` are the **source of truth fo
...
 - Keep JSDoc in sync with method names.
-- **Keep JSDoc on service class methods in sync with `{Entity}ServiceModel`** — both the models file and the service implementation file must have identical JSDoc for each public method. `ServiceModel` is the source of truth for docs, but the service class copy aids developer navigation and IDE tooltips. When updating JSDoc on one, update the other.
+- **Do NOT duplicate JSDoc on service class methods** — service classes `implements {Entity}ServiceModel`, and TypeScript inherits the interface's JSDoc onto the implementing method's IDE hover when the class method has no JSDoc of its own. Write method JSDoc **only** on `{Entity}ServiceModel`; leave the service class method bare (keep the `@track` decorator and body). This makes the interface the single source of truth for both the docs site (TypeDoc renders the interface) and IDE tooltips (inherited via `implements`), so the two can't drift. A partial JSDoc comment on the class method silently disables inheritance — so class methods must carry **no** method JSDoc at all. Methods not declared on the interface keep their own JSDoc (nothing to inherit).
```
vnaren23, line 61: Shouldnt this also be added to onboarding skill?

Raina451: skill uses rules.md only from agents doc:https://github.com/UiPath/uipath-typescript/blob/main/.claude/skills/onboard-api/SKILL.md#step-4-implement

swati354 (approved):

PR #595, opened 2026-07-10
ci: add sample-app PR gate

Raina451:
## Sample-app PR gate

Adds `scripts/check-samples.mjs` and a Sample App Checks workflow that enforces the sample-app conventions.

### Rules
| Rule | Applies to |
|------|-----------|
| has `uipath.json.example` | all apps |
| `.gitignore` present and ignores `.uipath` | all apps |
| README has a `Preview` heading with an image/gif/video | all apps |
| `index.html` favicon `<link>` points at a file that exists | apps with `index.html` |
| a coded-apps harness dep (`@uipath/coded-apps-dev`, or `@uipath/coded-action-app` for action apps) | apps with `index.html` |
| `vite.config` sets `base: './'` and (non-action apps) imports `@uipath/coded-apps-dev` | apps with a vite config |

Apps are auto-discovered (any tracked `package.json` under `samples/`). Rules are pure functions with a `--selftest`. Run all apps locally: `npm run check:samples`.

### How the files are read
The favicon rule reads an attribute out of `index.html`, so it parses the HTML with `node-html-parser` rather than a regex. The rest are grep-style checks (does the README mention media, does the vite config contain `base: './'`), which is what regex and `String.includes` are for. The vite check greps the source on purpose: loading the config to read `base` would need `npm ci` per app, which is too heavy for a structural gate.

### Scope: changed apps only (for now)
The CI job runs `--changed`, gating only the sample apps touched in a PR. Existing apps currently have **47 violations** (mostly missing demo gifs/favicons and per-app config the app owners need to supply). Rather than block every PR on that backlog, each app becomes enforced the next time it is changed. Flip CI to a full `npm run check:samples` once the backlog is cleared.

Generated with Claude Code


.github/workflows/sample-checks.yml
```diff
@@ -0,0 +1,29 @@
+name: Sample App Checks
```
vnaren23: Do we need a separate Lint? Cant this be a set in an existing Linter?

Raina451: added in existing pr checks now.

package.json
```diff
@@ -274,6 +275,7 @@
         "@vitest/ui": "^4.1.9",
         "builtin-modules": "^3.3.0",
         "dotenv": "^17.2.0",
+        "node-html-parser": "^9.0.0",
```
vnaren23, line 278: Why is this needed on SDK package?

.github/workflows/pr-checks.yml
```diff
@@ -70,3 +70,39 @@ jobs:
...
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      # Always run the rules' own self-test, so a rule-logic regression is
+      # caught even on a PR that changes no sample apps.
```
vnaren23: Do we need to run always?

package.json
Raina451: It's a devDependency, so it isn't shipped to consumers, it's only used by the sample-check script and npm run check:samples for local dev too.

.github/workflows/pr-checks.yml
Raina451: Moved the self-test into the unit test suite, so it runs with the normal tests. this job now only gates the changed apps.

vnaren23 (approved):

PR #596, opened 2026-07-10
fix(data-fabric): make FILE fields creatable end-to-end

AditiGoyalUipath:
## Summary
- FILE fields could not be created via the SDK — server returned `"Target entity is not provided in request"`.
- Fix mirrors the web UI's wire payload: emit `type:"relationship"` for FILE, drop `isForeignKey:true` for FILE.

## Root cause
`buildSchemaFieldPayload` was sending:
```json
{ "sqlType": {"name":"UNIQUEIDENTIFIER","lengthLimit":300}, "fieldDisplayType":"File", "isForeignKey": true }
```

Two problems vs. the UI request:
1. **Missing `type` discriminator.** The UI sends `type:"relationship"` alongside `fieldDisplayType:"File"` — that pair is how the server routes to the attachment auto-wire path.
2. **`isForeignKey:true` present.** With `isForeignKey` set but no `referenceEntity`/`referenceField`, the server enters its relationship pre-check and demands a target entity — the exact failure users hit.

Follow-up to #572 (which removed the client-side `referenceEntityId`/`referenceFieldId` requirement but didn't fix the outgoing payload).

## Change
```diff
- ...((isRelationship || isFile) && { isForeignKey: true }),
+ ...(isRelationship && { isForeignKey: true }),
  ...(isRelationship && { referenceType: ReferenceType.ManyToOne }),
+ ...(isFile && { type: 'relationship' as const }),
```
Plus `type?: string` on `FieldSchemaPayload` so TS accepts the new wire key.

## Test plan
- [x] Unit tests updated (198/198 pass) — assertions now match the corrected wire payload.
- [x] E2E verified against `cloud.uipath.com`: create entity with `{"fieldName":"attachment","type":"FILE"}` → succeeds; insert record → succeeds; `files.uploadAttachment` → succeeds; delete cleanup → succeeds.

Screenshot 2026-07-10 at 4 13 18 PM

### DIFF
--- a/src/services/data-fabric/entities.ts
+++ b/src/services/data-fabric/entities.ts
@@ -1316,12 +1316,7 @@ export class EntityService extends BaseService implements EntityServiceModel {
       isHiddenField: field.isHiddenField ?? false,
       ...(field.defaultValue !== undefined && { defaultValue: field.defaultValue }),
       ...(field.choiceSetId !== undefined && { choiceSetId: field.choiceSetId }),
-      // Relationship fields are routed server-side by `isForeignKey: true`
-      // (fieldDisplayType is downstream-only). FILE goes through a separate
-      // guarded branch keyed on `fieldDisplayType: "File"` that auto-wires the
-      // internal EntityAttachment reference unconditionally — so FILE must
-      // NOT emit `isForeignKey` (that would divert it into the relationship
-      // branch and trigger the "Target entity is not provided" pre-check).
+      // FILE routes via fieldDisplayType; isForeignKey would divert it to the relationship pre-check.
       ...(isRelationship && { isForeignKey: true, referenceType: ReferenceType.ManyToOne }),
       ...(!isFile && referenceEntityBody !== undefined && { referenceEntity: referenceEntityBody }),
       ...(referenceChoiceSetBody !== undefined && { referenceChoiceSet: referenceChoiceSetBody }),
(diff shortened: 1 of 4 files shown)


src/services/data-fabric/entities.ts
```diff
@@ -1316,9 +1316,13 @@ export class EntityService extends BaseService implements EntityServiceModel {
       isHiddenField: field.isHiddenField ?? false,
       ...(field.defaultValue !== undefined && { defaultValue: field.defaultValue }),
       ...(field.choiceSetId !== undefined && { choiceSetId: field.choiceSetId }),
-      ...((isRelationship || isFile) && { isForeignKey: true }),
+      ...(isRelationship && { isForeignKey: true }),
       ...(isRelationship && { referenceType: ReferenceType.ManyToOne }),
-      // FILE: server auto-wires refs to the EntityAttachment system entity; caller-sent UUIDs would fail the server's relationship pre-check.
+      // FILE: sent as a relationship-typed field with fieldDisplayType=File so the
```
vnaren23: can we add an explicit integration test to cover this scenario please.

AditiGoyalUipath: Modified existing test to have file field

vnaren23 (approved):

ninja-shreyash (approved):
PR desc looks stale

PR #599, opened 2026-07-12
feat(cas): add helper method for authenticated citation sources

scottcmg:
## Details
Customers have reported issues with the generated PDF citations that are returned as part of the conversational agent response where clicking the inline citation or the source in the source list returns a 401 in a new tab. This is because these citations require an authorized fetch from ECS to display.

This PR is to add a helper method that customers can use to add support for these citation links so they do not have to be responsible for independently determining the wiring needed for support. It adds some security hardening against (1) inline HTML execution and (2) hijacked download requests sending user credentials to a non-UiPath-origin source.


scottcmg:
Pushed fixes for the review 👇

- Telemetry key → `DownloadCitationSource` (PascalCase)
- Origin check now fails closed (`!base || ...`)
- Test param typed `Partial<CitationSourceMedia>` (no more `any`)
- Added network-failure + malformed-URL branch tests (fwiw `'not a url :://'` actually resolves against the base, so used `'https://['`)
- oauth-scopes: checked the ECS `ReferenceController` — it's just `ORG_MEMBER_OR_S2S`, no specific scope, so the entry says "none (authenticated org-member token)"

Skipped the integration test for now — conv-agent has no integration suite at all yet, so a one-off here felt out of place. Can do it as a separate pass if we want to stand that up.

docs/oauth-scopes.md
```diff
@@ -145,6 +145,7 @@ The `ConversationalAgents` scope is required for real-time WebSocket sessions (`
 |--------|-------------|
 | `getAll()` | `OR.Execution` or `OR.Execution.Read` |
 | `getById()` | `OR.Execution` or `OR.Execution.Read` |
+| `downloadCitationSource()` | None (any authenticated org-member token) |
```
vnaren23: Lets just mark it `NA`.

docs/oauth-scopes.md
```diff
@@ -145,6 +145,7 @@ The `ConversationalAgents` scope is required for real-time WebSocket sessions (`
 |--------|-------------|
 | `getAll()` | `OR.Execution` or `OR.Execution.Read` |
 | `getById()` | `OR.Execution` or `OR.Execution.Read` |
+| `downloadCitationSource()` | `NA` |
```
vnaren23: NA without quotes. Dont want an LLM to confuse it as an actual scope name 🤦

vnaren23 (approved):

docs/oauth-scopes.md
scottcmg: Oh my bad 😅 fixed~

src/services/conversational-agent/conversational-agent.ts
```diff
@@ -42,6 +47,9 @@ export class ConversationalAgentService extends BaseService implements Conversat
   /** Service for reading and updating the current user's profile/context settings. See {@link UserSettingsServiceModel}. */
   public readonly user: UserSettingsService;
 
+  /** Configured SDK origin, used to keep citation downloads on the tenant's host. */
+  private readonly baseUrl?: string;
```
swati354: `instance.config.baseUrl` is typed string on `BaseConfig`, so this can drop the `?` (and then we could remove the `!base` check in line 114)

src/services/conversational-agent/conversational-agent.ts
```diff
@@ -80,6 +89,59 @@ export class ConversationalAgentService extends BaseService implements Conversat
...
+    // server-provided; if one were ever malformed or injected to point at
+    // another host, do not send the user's token to that host.
+    const base = this.baseUrl ? new URL(this.baseUrl) : undefined;
+    let target: URL;
+    try {
+      target = new URL(source.downloadUrl, base);
+    } catch {
+      throw new ValidationError({
+        message: `Invalid citation downloadUrl`
+      });
```
swati354: nit : there's no interpolation, could be plain single quotes around the message. Same for line 116.
Also shouldn't both be `ServerError` instead?

tests/unit/services/conversational-agent/conversational-agent.test.ts
```diff
@@ -34,6 +36,16 @@ vi.mock('@/services/conversational-agent/conversations/session/session-manager',
...
 }));
 
+// Minimal non-OK Response stub for the error-mapping tests.
+const errorResponse = (status: number, statusText: string) => ({
+  ok: false,
+  status,
+  statusText,
+  json: async () => ({ message: statusText }),
+  text: async () => statusText,
+  headers: { get: () => null },
```
swati354: are these two read anywhere?

tests/unit/services/conversational-agent/conversational-agent.test.ts
```diff
@@ -346,6 +358,108 @@ describe('ConversationalAgentService Unit Tests', () => {
...
+        mediaSource({ mimeType: 'image/png', title: 'chart' }),
+      );
+
+      expect(blob.type).toBe('image/png');
+    });
+
+    it('throws a ValidationError when the source has no downloadUrl', async () => {
+      await expect(
+        conversationalAgent.downloadCitationSource(mediaSource({ downloadUrl: undefined })),
+      ).rejects.toThrow(/downloadUrl/);
```
swati354: Every other error test in this block uses `.rejects.toBeInstanceOf(ValidationError)`. This one uses `toThrow(/downloadUrl/)`. Any particular reason?

src/models/conversational-agent/conversational-agent.models.ts
```diff
@@ -141,6 +141,40 @@ export interface ConversationalAgentServiceModel {
...
+   * HTML content is intentionally returned as `application/octet-stream` (a
+   * download) rather than `text/html`, so previewing the blob inline can't
+   * execute citation markup in your app's origin.
+   *
+   * @param source - A media citation source (`CitationSourceMedia`) with a `downloadUrl`
+   * @returns Promise resolving to the document as a `Blob`
+   * @throws ValidationError if the source has no `downloadUrl`
+   * @throws A typed HTTP error for error responses — e.g. `AuthenticationError`
+   *         (401), `AuthorizationError` (403), `NotFoundError` (404) — or
+   *         `NetworkError` on a connection failure, matching other SDK calls
```
swati354: Does any method document the errors with the `@throw` tag? Could you verify. 
wherever it is used, I don't think it is in the docs surface.

src/models/conversational-agent/conversational-agent.models.ts
```diff
@@ -141,6 +141,40 @@ export interface ConversationalAgentServiceModel {
...
+   * Media citation sources expose a `downloadUrl` pointing at an auth-gated
+   * UiPath endpoint; opening it directly sends no token and fails with `401`.
+   * This performs the request with the SDK's access token and returns the bytes,
+   * with the `Blob` type resolved from the source `mimeType` (falling back to the
+   * response Content-Type, then the title's file extension). Use `source.title`
+   * as the file name.
+   *
+   * HTML content is intentionally returned as `application/octet-stream` (a
+   * download) rather than `text/html`, so previewing the blob inline can't
+   * execute citation markup in your app's origin.
```
swati354: could this description be made a little concise?

src/services/conversational-agent/conversational-agent.ts
```diff
@@ -80,6 +89,59 @@ export class ConversationalAgentService extends BaseService implements Conversat
...
+    } catch (error) {
+      throw ErrorFactory.createNetworkError(error);
+    }
+
+    if (!response.ok) {
+      const errorInfo = await errorResponseParser.parse(response);
+      throw ErrorFactory.createFromHttpStatus(response.status, errorInfo);
+    }
+
+    const blob = await response.blob();
```
swati354: should this be inside try-catch as well?
could we extend the existing try/catch to cover the body read too

something like 
```
try {
    const response = await fetch(target.href, {
      headers: { Authorization: `Bearer ${token}` }
    });

    if (!response.ok) {
      const errorInfo = await errorResponseParser.parse(response);
      throw ErrorFactory.createFromHttpStatus(response.status, errorInfo);
    }

    blob = await response.blob();
  } catch (error: any) {
    if (error.type && error.type.includes('Error')) throw error;
    throw ErrorFactory.createNetworkError(error);
  }
```

tests/unit/services/conversational-agent/conversational-agent.test.ts
scottcmg: Nope, residual from an earlier version I didn't remove - removed in latest commit.

tests/unit/services/conversational-agent/conversational-agent.test.ts
scottcmg: Consistency makes the most sense, changed.

src/models/conversational-agent/conversational-agent.models.ts
scottcmg: Oh yup - that tag isn't used, removed.

src/services/conversational-agent/conversational-agent.ts
scottcmg: I implemented something similar to cover the read - I kept it as separate try/catch blocks so that inner throws don't get caught and rethrown and so the error type is easy to scope. Let me know if you'd prefer a different implementation here though I'm happy to update.

swati354 (approved):

PR #607, opened 2026-07-15
[PLT-106969]feat(buckets): accept bucket id or name on file ops; add folderKey/folderPath to getAll

deepeshrai-tech:
## Summary

- Widens 5 `BucketService` file-op methods (`deleteFile`, `getFileMetaData`, `getFiles`, `getReadUri`, `uploadFile`) to accept `bucket: number | string` — either a numeric Id or a bucket Name. Names are resolved via a lean OData `$filter=Name eq '…'&$select=Id&$top=1` lookup on the folder-scoped Buckets collection.
- Extends `BucketGetAllOptions` with `folderKey?: string` and `folderPath?: string` so `getAll()` can be scoped by any of the three folder refs. `getAll()` with no folder options **remains a cross-folder query** — deliberately does not fall back to the SDK init-time meta-tag folderKey (non-breaking).
- Extracts shared name-lookup logic into a private `findByNameInFolder` helper on `FolderScopedService`; both `getByNameLookup` (full-resource fetch) and the new `resolveIdByName` (Id-only projection) delegate to it.

## Public API surface

```typescript
// Before
buckets.deleteFile(bucketId: number, path, options?);
buckets.uploadFile(bucketId: number, path, content, options?);
buckets.getAll(options?: { folderId?: number, ...odata });

// After
buckets.deleteFile(bucket: number | string, path, options?);
buckets.uploadFile(bucket: number | string, path, content, options?);
buckets.getAll(options?: {
  folderId?: number,
  folderKey?: string,
  folderPath?: string,
  ...odata,
});
```

Positional forms are the primary API. Deprecated options-only overloads (`uploadFile({ bucketId, ... })`, `getReadUri({ bucketId, ... })`) remain numeric-only for `bucketId` — they're already on the way out.

## Design notes

- Numeric ids skip the resolver entirely — no extra API call.
- Name resolution runs inside the same folder scope the caller provides (`folderId` / `folderKey` / `folderPath`); missing name → `NotFoundError` with folder hint.
- Resolver is not `@track`-decorated; the outer public method's `@track` fires once (avoids the delegation double-telemetry anti-pattern).
- `coerceBucketId` distinguishes missing-value from invalid-value error messages (`bucket is required` vs `bucket must be a positive numeric Id`).
- `getAll()` picks the endpoint based on whether any folder ref is in options — not on `folderId` alone — so `folderKey`/`folderPath` correctly trigger the folder-scoped endpoint. Endpoint constants unchanged.

## Test plan

- [x] Unit: 87 tests in `buckets.test.ts` (all passing). New coverage: bucket-name resolution for all 5 methods, `NotFoundError` propagation, `numeric id skips name lookup` regression guard, positive-Id validation, `folderKey`/`folderPath` on `getAll`, cross-folder-default preserved when SDK has init-time folderKey.
- [x] Integration: added `File operations - by bucket name` block exercising all 5 methods end-to-end + folderKey scoping on the by-name path; added folderKey/folderPath cases to the `getAll` describe.
- [x] Typecheck + lint clean.
- [x] Full unit-test suite: 2071/2071 passing.
- [x] Rollup build produces `dist/` cleanly.
- [x] Verified locally against alpha Orchestrator via a React sample app (kept local; not shipped with this PR).

### DIFF
--- a/src/models/orchestrator/buckets.models.ts
+++ b/src/models/orchestrator/buckets.models.ts
@@ -19,26 +19,37 @@ import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '.
  */
 export interface BucketServiceModel {
   /**
-   * Gets all buckets across folders with optional filtering
-   * 
+   * Gets all buckets across folders, or scoped to a specific folder when
+   * folder context is provided.
+   *
+   * Folder scoping is optional and can be supplied as `folderId`, `folderKey`,
+   * or `folderPath` in the options. When none of them is provided the request
+   * is a cross-folder query. Unlike file-operation methods, `getAll()` does
+   * **not** fall back to the SDK's init-time folder key — no folder in options
+   * always means cross-folder.
+   *
    * The method returns either:
    * - A NonPaginatedResponse with data and totalCount (when no pagination parameters are provided)
    * - A paginated result with navigation cursors (when any pagination parameter is provided)
-   * 
-   * @param options - Query options including optional folderId and pagination options
+   *
+   * @param options - Optional folder scoping (`folderId` / `folderKey` / `folderPath`), OData query options, and pagination options
    * @returns Promise resolving to either an array of buckets NonPaginatedResponse<BucketGetResponse> or a PaginatedResponse<BucketGetResponse> when pagination options are used.
    * {@link BucketGetResponse}
    * @example
    * ```typescript
-   * // Get all buckets across folders
+   * // Cross-folder — every bucket the caller can see
    * const allBuckets = await buckets.getAll();
    *
-   * // Get buckets within a specific folder
-   * const folderBuckets = await buckets.getAll({
-   *   folderId: <folderId>
-   * });
+   * // Scoped by folder ID
+   * const folderBuckets = await buckets.getAll({ folderId: <folderId> });
    *
-   * // Get buckets with filtering
+   * // Scoped by folder key (GUID)
+   * await buckets.getAll({ folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   *
+   * // Scoped by folder path
+   * await buckets.getAll({ folderPath: 'Shared/Finance' });
+   *
+   * // Filtering
    * const filteredBuckets = await buckets.getAll({
    *   filter: "name eq 'MyBucket'"
    * });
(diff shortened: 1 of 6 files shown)


src/models/orchestrator/buckets.models.ts
```diff
@@ -236,15 +269,15 @@ export interface BucketServiceModel {
...
+    bucket: number | string,
     path: string,
     content: Blob | Uint8Array<ArrayBuffer> | File,
     options?: BucketUploadFileRequestOptions,
   ): Promise<BucketUploadResponse>;
   /**
    * Uploads a file to a bucket — options-only form.
    *
-   * @deprecated Use the positional form: `uploadFile(bucketId, path, content, options?)`. See {@link BucketUploadFileRequestOptions} for the supported options.
+   * @deprecated Use the positional form: `uploadFile(bucket, path, content, options?)`. See {@link BucketUploadFileRequestOptions} for the supported options.
```
vnaren23, line 280: This shouldnt be changing right?

src/services/orchestrator/buckets/buckets.ts
```diff
@@ -122,34 +143,36 @@ export class BucketService extends FolderScopedService implements BucketServiceM
...
       folderId = optionsOrFolderId;
       restOptions = (legacyOptions ?? {}) as Omit<T, 'folderId' | 'folderKey' | 'folderPath'>;
     } else {
-      // Preferred form: getFileMetaData(bucketId, options?)
+      // Preferred form: getFileMetaData(bucket, options?)
       const opts = optionsOrFolderId ?? ({} as T);
       ({ folderId, folderKey, folderPath, ...restOptions } = opts);
     }
 
+    const bucketId = await this.coerceBucketId(
```
ninja-shreyash, line 170: this extra would need OR.Buckets.Read. Pls update oauth-scopes.md

src/services/orchestrator/buckets/buckets.ts
```diff
@@ -448,19 +483,23 @@ export class BucketService extends FolderScopedService implements BucketServiceM
...
-    }
-
+  async deleteFile(bucket: number | string, path: string, options?: BucketDeleteFileOptions): Promise<void> {
     if (!path) {
       throw new ValidationError({ message: 'path is required for deleteFile' });
     }
 
+    const { folderId, folderKey, folderPath } = options ?? {};
+
+    const bucketId = await this.coerceBucketId(
```
ninja-shreyash, line 493: This would require OR.Buckets.Read permission. Pls verify/update oauth-scopes.md

src/services/orchestrator/buckets/buckets.ts
```diff
@@ -499,4 +538,35 @@ export class BucketService extends FolderScopedService implements BucketServiceM
...
+  private async coerceBucketId(
+    bucket: number | string,
+    folder: { folderId?: number; folderKey?: string; folderPath?: string },
+    callerLabel: string,
+  ): Promise<number> {
+    if (bucket === null || bucket === undefined) {
+      throw new ValidationError({ message: `bucket is required for ${callerLabel}` });
+    }
+    if (typeof bucket === 'number') {
+      if (!Number.isFinite(bucket) || bucket <= 0) {
```
ninja-shreyash, line 557:
```suggestion
      if (!Number.isInteger(bucket) || bucket <= 0) {
```

PR #619, opened 2026-07-22
feat(data-fabric): align entity create field names with get response

niket0503:
## Summary

The entity **create** request used different names than the **get** response for the same concepts. This aligns the create side to the get/API vocabulary — **pure renames only**, the flat/ergonomic create shape is unchanged.

| Concept | Before (create) | After (create) | Matches get |
|---|---|---|---|
| Field name | `fieldName` | `name` | `name` |
| Remove-field name | `fieldName` | `name` | `name` |
| Analytics flag | `isAnalyticsEnabled` | `isInsightsEnabled` | `isInsightsEnabled` |

## Intentionally NOT changed
- **Flattened fields** — scalar `type` + flat constraints (`lengthLimit`, `maxValue`…) and bare-id references (`referenceEntityId`/`referenceFieldId`) vs the nested `fieldDataType`/`referenceEntity`/`referenceField` in the response. These are deliberate create-input ergonomics; the SDK/backend bridge the gap (SDK maps `type` ⇄ `sqlType`+`fieldDisplayType`; backend resolves reference ids to full objects on read).
- **`folderKey` vs `folderId`** — `folderKey` is the SDK-wide folder-scoping input used by every service; `folderId` is the entity's stored attribute in the response. Renaming would break cross-SDK consistency.

## ⚠️ Breaking change
Public create API renames: `EntityCreateFieldOptions.fieldName`→`name`, `EntityRemoveFieldOptions.fieldName`→`name`, `EntityCreateOptions.isAnalyticsEnabled`→`isInsightsEnabled`.

**CLI follow-up needed:** the `data-fabric-tool` `df entities create` command validates `fieldName` and its examples use it — it needs a paired `fieldName`→`name` update when it bumps the SDK. (`isAnalyticsEnabled` isn't used by the CLI create.)

### DIFF
--- a/src/models/data-fabric/entities.types.ts
+++ b/src/models/data-fabric/entities.types.ts
@@ -348,8 +348,9 @@ export interface EntityCreateFieldOptions extends EntityFieldBase {
   /**
    * Field name — must start with a letter and contain only
    * letters, numbers, and underscores (e.g., `"productName"`).
+   * Matches the `name` field returned by `getById`.
    */
-  fieldName: string;
+  name: string;
   /** Field data type — one of the {@link EntityFieldDataType} values (default: STRING) */
   type?: EntityFieldDataType;
   /** Choice set ID for choice-set fields */
@@ -393,7 +394,12 @@ export interface EntityCreateOptions extends EntityFolderScopedOptions {
   description?: string;
   /** Whether role-based access control is enabled for this entity (default: false) */
   isRbacEnabled?: boolean;
-  /** Whether Analytics integration is enabled for this entity (default: false) */
+  /**
+   * Whether Analytics integration is enabled for this entity (default: false).
+   * Surfaced by `getById` as `isAnalyticsEnabled`.
+   *
+   * @experimental Analytics integration is in preview — the contract may change.
+   */
   isAnalyticsEnabled?: boolean;
   /** External field source definitions (default: empty) */
   externalFields?: ExternalField[];
@@ -412,7 +418,7 @@ export interface EntityFieldUpdateOptions extends EntityFieldBase {
  */
 export interface EntityRemoveFieldOptions {
   /** Name of the field to remove */
-  fieldName: string;
+  name: string;
 }
 
 /**
@@ -729,7 +735,12 @@ export interface RawEntityGetResponse {
   usedStorageSizeInMB?: number;
   attachmentSizeInByte?: number;
   isRbacEnabled: boolean;
-  isInsightsEnabled?: boolean;
+  /**
+   * Whether Analytics integration is enabled for this entity.
+   *
+   * @experimental Analytics integration is in preview — the contract may change.
+   */
+  isAnalyticsEnabled?: boolean;
   id: string;
   createdBy: string;
   createdTime: string;
--- a/src/services/data-fabric/entities.ts
+++ b/src/services/data-fabric/entities.ts
@@ -455,7 +455,7 @@ export class EntityService extends BaseService implements EntityServiceModel {
 
     // Filter out removed fields
     if (options.removeFields?.length) {
-      const removeSet = new Set(options.removeFields.map(r => r.fieldName));
+      const removeSet = new Set(options.removeFields.map(r => r.name));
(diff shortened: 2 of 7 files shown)


AditiGoyalUipath (approved):

src/services/data-fabric/entities.ts
```diff
@@ -511,7 +511,9 @@ export class EntityService extends BaseService implements EntityServiceModel {
           fields: [...fields, ...newFields],
           folderId: raw.folderId ?? DATA_FABRIC_TENANT_FOLDER_ID,
           isRbacEnabled: raw.isRbacEnabled ?? false,
-          isInsightsEnabled: raw.isInsightsEnabled ?? false,
+          // `raw` is the untransformed GET response, so read the wire field `isInsightsEnabled`
+          // directly (EntityMap renames it to `isAnalyticsEnabled` only on the transformed read path).
```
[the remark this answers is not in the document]
niket0503: added

src/services/data-fabric/entities.ts
```diff
@@ -553,31 +511,16 @@ export class EntityService extends BaseService implements EntityServiceModel {
...
+          isInsightsEnabled: (raw as { isInsightsEnabled?: boolean }).isInsightsEnabled ?? false,
```
AditiGoyalUipath, line 516: This also should be isAnalyticsEnabled

niket0503: This is the upsert payload for api, which needs isInsightsEnabled

src/models/data-fabric/entities.types.ts
```diff
@@ -401,8 +394,8 @@ export interface EntityCreateOptions extends EntityFolderScopedOptions {
   description?: string;
   /** Whether role-based access control is enabled for this entity (default: false) */
   isRbacEnabled?: boolean;
-  /** Whether Analytics integration is enabled for this entity (default: false) */
-  isAnalyticsEnabled?: boolean;
+  /** Whether Insights integration is enabled for this entity (default: false). Matches the `isInsightsEnabled` field returned by `getById`. */
+  isInsightsEnabled?: boolean;
```
vnaren23: Please revert this. It was intentionally meant to be `isAnalyticsEnabled`.

AditiGoyalUipath: @vnaren23 We intentionally kept it as isAnalyticsEnabled But it is creating more confusion . Since Get API respond with isInsightsEnabled and create accept isAnalyticsEnabled. Instead of remapping the response of get Method , We felt if we can keep this flag same as isInsightsEnabled. Even in Python SDK we are exposing it as isInsightsEnabled only. Its is better to keep the source of truth same as API's unless necessary. Otherwise will cause more confusion if we keep it different for people using from coding agent/coded agent/coded app etc

vnaren23: This is internal mapping. External contracts do not see the mapping we are doing. We intentionally took the route to keep external facing properties to be generic and not leak internal implementation details. I dont see a reason why we should revert this now.

AditiGoyalUipath: In that case get Response should also map it back to isAnalyticsEnabled right ?  My concern is We can keep same naming convention for Create Method and Get Method.

vnaren23: I'm assuming get response is already done.

src/models/data-fabric/entities.types.ts
```diff
@@ -739,7 +730,8 @@ export interface RawEntityGetResponse {
   usedStorageSizeInMB?: number;
   attachmentSizeInByte?: number;
   isRbacEnabled: boolean;
-  isInsightsEnabled?: boolean;
+  /** Whether Analytics integration is enabled for this entity. */
```
vnaren23: is this comment needed?

vnaren23 (approved):

src/models/data-fabric/entities.types.ts
niket0503: removed

AditiGoyalUipath (approved):

PR #621, opened 2026-07-22
feat(cas): add anonymous chat steps to docs

scottcmg:
## Description
Conversational Agents provides support for anonymous chat using client credential token types from app-scoped confidential external apps, but the docs didn't have any information on how to do this.

This PR adds this functionality info and example in the conversational agent, and expands the auth docs to include information on how to set up an external app to use with conversational agents for anonymous chat.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -63,6 +63,26 @@ To Generate a PAT Token:
 4. Give it a name and expiration date
 5. Provide relevant scopes
 
+Conversational Agents support anonymous, sign-in-free chat via tokens issued to confidential, app-scoped External Apps. In this SDK, this token type is only guaranteed to work with Conversational Agents.
+
+To create a confidential External App:
+
+1. In UiPath Cloud: **Admin** → **External Applications**
+2. Click **Add Application** → **Confidential Application**
+3. Configure:
+    - **Name**: Your app name
+    - **Redirect URI**: For example, `http://localhost:3000` (for development)
+    - **Scopes**: Select the permissions you need ([see scopes guide](/uipath-typescript/oauth-scopes)). All scopes must be added as **Application** scopes, not **User** scopes.
+4. Save and copy the **Client ID** and the **Client Secret** — you will not be shown the Client Secret again
+5. Follow the [official UiPath documentation](https://docs.uipath.com/automation-cloud/automation-cloud/latest/api-guide/accessing-uipath-resources-using-external-applications) on how to request client-credentials tokens for a confidential app with application scopes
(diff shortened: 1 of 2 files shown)


src/models/conversational-agent/conversational-agent.models.ts
```diff
@@ -98,6 +98,44 @@ import type { ConnectionStatus } from '@/core/websocket';
...
+ * });
+ * ```
+ *
+ * The SDK forwards this identifier on every HTTP request and real-time WebSocket session. Each distinct
+ * `externalUserId` — scoped to the client ID of the External App the token was issued for — gets its own
+ * conversation history and user settings, and the same value always maps back to the same user.
+ *
+ * ### Limitations
+ *
+ * - **App-scoped tokens only.** `externalUserId` takes effect only when the SDK is authenticated
```
Raina451, line 130: > only when the SDK is authenticated

does this mean that confidential app client secret needs to be stored on customer backend and authenticate sdk via the token generated by this client id and secret like this?: 
https://github.com/UiPath/uipath-typescript/blob/2e2b7e9988888bc5eff6fa2038000249e25bda02/docs/authentication.md?plain=1#L55-L56

if yes, I would suggest to improve the not here to mention confidential client can also be used and sdk can be suthenticated with secret parameter.

docs/authentication.md
```diff
@@ -63,6 +63,26 @@ To Generate a PAT Token:
 4. Give it a name and expiration date
 5. Provide relevant scopes
 
+Conversational Agents support anonymous, sign-in-free chat via tokens issued to confidential, app-scoped External Apps. In this SDK, this token type is only guaranteed to work with Conversational Agents.
```
Raina451: and if above [comment](https://github.com/UiPath/uipath-typescript/pull/621#discussion_r3635912746) is true then we can frame this genericaly and not specific to conversational agent

src/models/conversational-agent/conversational-agent.models.ts
scottcmg: What this block is for is to explain more about the limitations on externalUserIds which is an additional field required to use app-scoped tokens in conversational agents service. I think this block is very specific to the conversational agents service itself and should be separate from the general SDK note "Using externally obtained tokens". Let me know if you feel differently though or if anything here you feel will be unclear to users.

docs/authentication.md
scottcmg: Sure, changed it to be generic. I'm not sure if it's usable with every service though, it might not work the same as a user token in some APIs.

Raina451 (approved):

vnaren23 (approved):

PR #628, opened 2026-07-30
ci: add option to publish only to GitHub Packages

Sarath1018:
## What

Adds a **`github_packages_only`** boolean input (checkbox) to the **Publish SDK package** workflow. When checked, the run publishes **only to GitHub Packages** (`npm.pkg.github.com`) and skips everything that belongs to an official release.

## Why

Prerelease / test builds (e.g. `1.5.5-dev`) are currently published to GitHub Packages manually from developer machines with personal PATs, because the existing workflow always publishes to the public npm registry in the same run. This option makes those prereleases go through the same audited workflow instead.

## Behavior

| Step / Job | Default run | `github_packages_only` checked |
|---|---|---|
| Build + telemetry version bake | ✅ | ✅ |
| Publish to npm (public registry) | ✅ | ⏭️ skipped (all 3 packages: ts-sdk, coded-action-app-sdk, telemetry) |
| Publish to GitHub Packages | ✅ | ✅ |
| `deploy-docs` job | ✅ | ⏭️ skipped |
| `trigger-cf-worker` job | ✅ | ⏭️ skipped |
| Run summary | unchanged | notes npm was skipped |

- Default is `false`, so existing release runs behave exactly as before.
- The input is passed to summary scripts via `env:` (not inline interpolation), per Actions injection-safety guidance.
- Docs deploy and Cloudflare worker deploy are skipped because they should only accompany official npm releases.

## Usage

Actions → **Publish SDK package** → Run workflow → select package → check **Publish only to GitHub Packages** → run from the branch containing the prerelease version in `package.json`.

## Files

| Area | Files |
|------|-------|
| CI | `.github/workflows/publish.yml` |

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,6 +12,11 @@ on:
           - ts-sdk
           - coded-action-app-sdk
           - telemetry
+      github_packages_only:
+        description: 'Publish only to GitHub Packages (skip npm, docs deploy, and CF worker)'
+        required: false
+        default: false
+        type: boolean
 
 permissions: {}
 
@@ -58,6 +63,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Publish to npm
+        if: ${{ !inputs.github_packages_only }}
         run: |
           echo "@uipath:registry=https://registry.npmjs.org" > .npmrc
           npm publish --provenance --access public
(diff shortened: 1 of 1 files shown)


ninja-shreyash (approved):

.github/workflows/publish.yml
ninja-shreyash, line 75: can we use `--tag dev` so that it doesnt get assigned "latest" tag in github packages?

ninja-shreyash, line 75: also we should add a check on sdk version so that it does indeed point to prerelease version whenever `inputs.github_packages_only` is true. We should fail the workflow if user selects `github_packages_only` = true but tries to release public version.

Raina451 (approved):

.github/workflows/publish.yml
Sarath1018: addressed here https://github.com/UiPath/uipath-typescript/pull/638

Sarath1018: addressed here https://github.com/UiPath/uipath-typescript/pull/638

PR #629, opened 2026-07-31
feat(platform): add Platform service for user-scoped platform settings

kittyyueli:
Adds a new modular `Platform` service exposing the two Setting endpoints available to external apps, via the `@uipath/uipath-typescript/platform` subpath.

Named `Platform` per review — "Identity" is the internal service name, and Platform Management expects further surface (organization settings, licensing, tenants) in the same namespace. The methods say which scope they act on for the same reason.

## API

```typescript
import { Platform, PlatformSettingKey } from '@uipath/uipath-typescript/platform';

await sdk.initialize();

const platform = new Platform(sdk);

// GET identity_/api/Setting?key=...&userId=...
const settings = await platform.getUserSettings([PlatformSettingKey.UserTheme], userId, { organizationId });
const theme = settings.find(s => s.key === PlatformSettingKey.UserTheme)?.value;

// PUT identity_/api/Setting
const updated = await platform.updateUserSettings(
  [{ key: PlatformSettingKey.UserTheme, value: 'dark' }],
  userId,
  organizationId
);
```

| Method | Endpoint | OAuth scope |
|--------|----------|-------------|
| `getUserSettings()` | `GET identity_/api/Setting` | `PM.Setting` or `PM.Setting.Read` |
| `updateUserSettings()` | `PUT identity_/api/Setting` | `PM.Setting` or `PM.Setting.Write` |

Both are bulk operations. Keys with no stored value are omitted from the read response, so the result may be shorter than the list of keys requested.

## Design notes

**Organization-level URL routing.** The Setting URLs carry no tenant segment, but `ApiClient` unconditionally inserts the organization and tenant path segments. `IDENTITY_API_BASE` uses the same `../` URL-normalization pattern already documented on `NOTIFICATION_BASE` to collapse the tenant segment, so `{org}/{tenant}/../identity_/api/Setting` resolves to `{org}/identity_/api/Setting`. A unit test pins this. The org-level shape is corroborated by `token-manager.ts`, which already calls `{baseUrl}/{orgName}/identity_/connect/token`.

**Keys are a closed set.** `PlatformSettingKey` enumerates the eight supported keys; anything else is a compile error on both operations. `PlatformSetting.key` is typed as the enum too, so response keys can be compared without a cast. A unit test pins the members and their wire values, since adding or renaming one is a public API change.

**Every operation is user-scoped.** The settings store distinguishes organization-scoped from user-scoped rows by whether `userId` is present in the request — omitting it targets the organization-wide row rather than falling back to the calling user. This SDK exposes only the user-scoped path: `userId` is a required argument on both methods and is always sent (query string on reads, body on writes), and both throw `ValidationError` if it is missing, so no call can silently become organization-scoped. This is why the methods are named `getUserSettings` / `updateUserSettings`, and why `PlatformSetting.userId` is a non-null `string`.

**The organization must be named on reads too.** Omitting `organizationId` makes the API target the **host** partition rather than the caller's own organization, which an external application is not authorized for — the request comes back `403`. This is why `getUserSettings` takes it as an option and the integration suite always passes it; it is a required positional argument on writes.

**`partitionGlobalId` is renamed to `organizationId`.** The API calls the organization a "partition", but every other service in this package names that GUID `organizationId` (see the traces types, where it is documented as the organization/account GUID). The wire contract is unchanged — `partitionGlobalId` is still what goes over HTTP — but it is confined to the request boundary via `PlatformSettingMap` + `transformData`, with `RawPlatformSetting` in `platform.internal-types.ts` describing the wire shape. `partitionGlobalId` does not appear anywhere in the published `.d.ts`.

**Values are always strings.** Numbers and booleans arrive as `'8'` / `'false'`; the `UserCase.*` keys hold JSON strings the caller parses.

**Names kept on `identity`.** `IDENTITY_BASE` / `IDENTITY_API_BASE` and the `identity_/api/Setting` URL describe the wire, not the SDK surface, so they are unchanged; `PLATFORM_SETTING_ENDPOINTS` lives in a new `endpoints/platform.ts`, leaving `endpoints/identity.ts` to the OAuth `connect/*` endpoints it held before this PR. The `IDENTITY_TEST_USER_ID` env var and its `identityTestUserId` config field also keep their names, tracking the already-provisioned `UIPATH_IDENTITY_TEST_USER_ID` repository secret.

The Conversational Agent user service keeps its own `getSettings()` / `updateSettings()` — different service, untouched by this PR.

## Validation

Request and response shapes were checked against a live response and against the first-party portal clients for the same endpoint, which corrected several initial assumptions:

- `key` is a repeated query param (`key=a&key=b`), not a comma-joined list
- the response carries `id`, the organization GUID, and `userId` in addition to `key` and `value`
- the PUT body is an object (`settings`, `partitionGlobalId`, `userId`), not a bare array, and scope travels in the body rather than the query string
- the PUT responds with the stored rows, so `updateUserSettings` returns `PlatformSetting[]` rather than a wrapper echoing the submitted payload

## Verification

`npm run typecheck`, `npm run lint` (0 errors), `npm run test:unit` (2177 passing, 27 for Platform), `npm run build`, and `npm run docs:api` all pass. The build emits `dist/platform/{index.mjs,index.cjs,index.d.ts}` exporting `PlatformService as Platform` plus `PlatformSettingKey` as a runtime value. Platform source is at 100% statements/branches/functions, and the suite additionally covers the subpath barrel entry point, all eight enum members through the request path, empty responses, and bulk writes.

**The integration suite passes in CI.** `tests/integration/shared/platform/platform.integration.test.ts` — 8 tests, green. No scope grant and no proxy change were needed; the earlier `403`s were caused by the suite not sending the organization GUID on reads, fixed in 54df90d.

Docs updated: `docs/oauth-scopes.md`, `mkdocs.yml` nav, `package.json` exports, `rollup.config.js` entry.

## Open items for review

- **`keys` is required on `getUserSettings`.** Every observed call passes keys, so a keys-less "fetch everything" call is not exposed. Easy to relax if the endpoint does support it.
- **Enum member JSDoc is inferred from key names** (e.g. `UserAlert.AlertDuration` as "how long alerts remain visible"). Worth a check from someone who knows the exact semantics.
- **Cloudflare Workers whitelisting** may still be wanted for consumers going through `alpha.api.uipath.com`, even though CI does not need it. `apps-dev-tools` is not reachable from this environment; the pattern would be `/^\/[^/]+\/identity_\/api\/Setting$/`.
- `DELETE api/Setting` and `POST api/Setting/VerifySmtpSettings` exist on the same controller but are out of scope here; happy to add either.

https://claude.ai/code/session_013Q3NWsanguD7ZWnYjFSEf6


kittyyueli:
Thanks — taking the two findings separately.

**1. Options type name — fixed** in 6c301d0. `IdentitySettingsGetOptions` → `IdentitySettingGetOptions`. Correct call: every options type in the package uses the singular entity name (`AssetGetByIdOptions`, `AgentGetSummaryOptions`, `AgentMemoryGetTimelineOptions`), and every other type in this module already did too (`IdentitySetting`, `IdentitySettingUpsert`, `IdentitySettingKey`, `IdentitySettingMap`, `RawIdentitySetting`). That one was the outlier.

**2. `describe.skip` vs `beforeAll` + throw — leaving as-is for a human call, for three reasons.**

*The missing-scope premise is no longer supported by the evidence.* I based the `PM.Setting` explanation on the run where Identity 403'd. In the very next run, with Identity skipped, the same 403 appeared in a suite this PR does not touch:

```
AuthorizationError: Forbidden (403)
❯ EntityService.getAll  src/services/data-fabric/entities.ts:276
❯ tests/integration/shared/data-fabric/entities.integration.test.ts:169
```

Data Fabric `getAll` passed in the earlier run and failed 25 minutes later, so the 403 is roaming between suites rather than sitting on one service. That points at an environment-level authorization problem, not a per-service scope gap — which means my own stated reason for the skip is probably wrong, and a `PM.Setting` grant may not be the fix.

*The suggested snippet conflicts with another rule.* `agent_docs/rules.md` also says "**NEVER** wrap integration test API calls in try/catch — let errors propagate naturally. Silent catches mask real failures." The proposed `try { ... } catch (err) { throw new Error(...) }` around `getSettings` is that pattern, and it also replaces a typed `AuthorizationError` with a generic `Error`, losing `statusCode` for anyone debugging.

*Avoiding a flip-flop.* The previous commit moved this from a throw to a skip. Moving it back in the next commit, then possibly forward again once the 403 is understood, is churn on a decision that hinges on facts we don't have yet — whether the 403 is a permanent scope gap (skip is right) or transient environment breakage (throw is right, and the test would pass once the environment recovers).

Happy to switch to the guard the moment someone confirms which it is. Flagging for the reviewer's call rather than guessing.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
Following up on my previous comment — the evidence resolved, and it resolved in favour of the review. **`describe.skip` is removed in d9c5595**; the suite is back with its `beforeAll` guard.

Three consecutive coverage runs each 403'd in a *different, unrelated* suite:

| Run | Failing suite | Call |
|---|---|---|
| 05:08 | Identity | `getSettings` |
| 05:31 | Data Fabric | `EntityService.getAll` |
| 05:58 | Agent Feedback | `deleteCategory`, `submit` |

Data Fabric passed in the third run after failing in the second. A 403 that moves between unrelated services run to run is environment-level authorization trouble, not a missing `PM.Setting` scope on one endpoint. My original diagnosis was wrong, and the skip built on it was hiding a suite for a reason that does not hold — Identity is in exactly the same position as every other integration suite, none of which are skipped.

The guard still does not wrap the call in try/catch, per "**NEVER** wrap integration test API calls in try/catch — let errors propagate naturally." An `AuthorizationError` with `statusCode: 403` propagating unmodified is more useful than a re-thrown generic `Error`, and it reads identically to how the Data Fabric and Feedback failures above surfaced.

**This does mean the coverage job will stay red while the environment misbehaves**, and that is not something this PR can fix. Worth noting the failures are concentrated in write operations against the dev test organization (`deleteCategory`, `submit`, `getAll`), so it may be worth checking whether the CI application's credentials or scope grants changed recently — it is affecting suites well beyond this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
Correcting myself again — the fourth run separates what I had conflated into one cause. There are **two independent problems**, and neither is fixable from this PR.

| Run | Identity | Agent Feedback | Data Fabric |
|---|---|---|---|
| 05:08 | **403** | pass | pass |
| 05:31 | *skipped* | pass | **403** |
| 05:58 | *skipped* | **403** ×3 | pass |
| 06:06 | **403** | **401** | pass |

**1. Identity looks like a genuine scope gap after all.** It has returned 403 on both runs where it actually executed — 2 for 2. That is consistent behaviour, not the roaming pattern I described in my previous comment. My "environment-wide" explanation over-generalised from runs where Identity was skipped and therefore contributed no data. The `PM.Setting` grant is back to being the most likely fix.

**2. Something separate is genuinely unstable.** Agent Feedback failed with **403** at 05:58 and **401** at 06:06 on identical code, and Data Fabric failed once then passed twice. An error code that changes between 403 and 401 across runs is not a scope configuration — that is a credential or token-exchange problem, and it affects suites this PR does not touch.

**I am not re-adding `describe.skip`.** Two reasons. It would be the third flip on one decision, which is churn on a call that should be made once with facts in hand. More importantly it would not work: coverage would still be red from the Agent Feedback 401, so no change to this PR's code can turn this job green.

What would actually unblock it:

- grant **`PM.Setting`** to the CI external application, which should settle Identity, and confirms or refutes the scope theory either way;
- separately, look at why the dev test organization's credentials are returning 401/403 inconsistently across Feedback and Data Fabric — that predates and outlives this PR.

Everything else on this PR is green: lint, unit tests (2140, 27 for Identity), build, docs generation, and the review bot's last two passes. The Identity integration suite is written in full and will run unchanged once the scope lands.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
The run after the rebase settles this. **The environment instability has cleared, and Identity still 403s** — so there is one problem left, not two, and it is a stable authorization gap rather than flakiness.

| Run | Base | Identity | Agent Feedback | Data Fabric |
|---|---|---|---|---|
| 05:08 | `e006b90` | **403** | pass | pass |
| 05:31 | `e006b90` | *skipped* | pass | **403** |
| 05:58 | `e006b90` | *skipped* | **403** ×3 | pass |
| 06:06 | `e006b90` | **403** | **401** | pass |
| 17:47 | `0d637be` | **403** | pass | pass |

The latest run is otherwise completely clean — **302 passed, 0 failures** apart from Identity's `beforeAll`, with Feedback and Data Fabric both green. Identity is now 403 on **3 of 3** runs where it actually executed, spanning ~12 hours and two different base commits. That rules out the credential flapping I blamed earlier and confirms the CI external application is not authorized for `identity_/api/Setting`.

**Granting `PM.Setting` to the CI application should be the whole fix.** Nothing in this PR's code can substitute for it — a wrong user or organization would return an empty array with 200, not a 403.

Two other things from this run worth recording:

- **The rebase onto `0d637be` is clean.** All of main's new tests pass (302 up from 295, including the new Functions suite), and the conflict resolutions in `package.json`, `rollup.config.js`, `test-config.ts`, `unified-setup.ts`, and `coverage.yml` are verified to have kept main's additions intact alongside this branch's.
- **Skipping the Identity suite would now turn this job green**, which was not true earlier when Feedback was also failing. I have deliberately not done that: it has already been flipped twice, the convention argument against it stands, and hiding a suite is worse than a red check that names a real missing scope. Happy to skip it if maintainers would rather merge now and re-enable when the scope lands — just say so.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integration/shared/identity/identity.integration.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+const modes: InitMode[] = ['v1'];
+
+const ALL_KEYS = Object.values(IdentitySettingKey);
+
+// Identity settings are scoped to (organization, user), so this GUID is only meaningful in
+// the organization the test PAT authenticates against. Set IDENTITY_TEST_USER_ID to
+// override when running against a different organization — a GUID from another
+// organization returns no rows rather than an error.
+const DEFAULT_TEST_USER_ID = '81a27926-9d8d-4c62-84e5-df1c51c0b676';
```
Sarath1018: Drop the hardcoded fallback and use IDENTITY_TEST_USER_ID which is present in the github secrets already

kittyyueli: Done in 5fff732 — the fallback is gone and the suite now throws when `IDENTITY_TEST_USER_ID` is absent. Thanks, and the workflow plumbing to read it was already added in this PR (`UIPATH_IDENTITY_TEST_USER_ID_DEV || UIPATH_IDENTITY_TEST_USER_ID`, matching the pattern the other fixtures use).

Worth flagging that your comment may also explain the failure I have been misdiagnosing. I had been reporting the suite's 403 as the CI application missing the `PM.Setting` scope, resting on my assumption that a wrong user would return `200` with an empty array rather than a `403`. That assumption was never verified. The fallback GUID I had hardcoded belongs to a user in a **different organization** than the CI PAT authenticates against, and since settings are keyed on (organization, user), the API plausibly rejects a foreign `userId` outright — in which case the 403 was self-inflicted and this change fixes it.

The next coverage run should tell us which it is:

- suite passes → it was the wrong user, and no scope grant is needed;
- still 403 with the real secret → the `PM.Setting` gap is real after all, and my earlier request stands.

Either way the diagnosis stops being guesswork, so I have not raised the scope request again pending that result.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli: Result is in, and it settles it: **the scope gap is real.** The foreign-`userId` theory I floated above is dead.

Two things the `5fff732` run proves:

1. **The secret exists and reaches the test.** The new "IDENTITY_TEST_USER_ID is not configured" guard (lines 32–36) did **not** fire — execution reached the API call at line 41. So your point was right and the workflow plumbing reads the secret correctly.
2. **With the correct CI user, `getSettings` still returns 403.**

```
AuthorizationError: Forbidden
❯ IdentityService.getSettings src/services/identity/identity.ts:55:22
❯ tests/integration/shared/identity/identity.integration.test.ts:41:22
Serialized Error: { statusCode: 403, ... }
```

That is 4 of 4 runs where the suite executed, now including one with a user that genuinely belongs to the CI organization. Agent Feedback and Data Fabric passed in the same runs, so the environment is healthy. The only remaining explanation is that the CI external application is not authorized for `identity_/api/Setting`.

**Ask: please grant `PM.Setting` (or `PM.Setting.Read` + `PM.Setting.Write`) to the CI external application.** No code change on this PR can substitute for it, and the suite should pass unmodified once it lands.

If you would rather merge before that happens, I can mark the suite `describe.skip` with the reason recorded — that would turn the coverage job green today, since Identity is now the only failure. I have not done it unprompted because it has already been flipped twice on this PR and hiding the suite is worse than a red check naming a real missing scope. Your call.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
**Retracting my `PM.Setting` request — it was wrong.** @Sarath1018 had the real cause in Slack and I under-weighted it. The remaining 403 is the Cloudflare Workers proxy rejecting an un-whitelisted endpoint, not a scope gap.

The `e1f39e7` run closes the loop. Both scope values now reach the API — the `beforeAll` guard requiring `IDENTITY_TEST_USER_ID` **and** `UIPATH_ORGANIZATION_ID` did not fire, so execution got to the call at line 44 — and it is still 403, while Sarath ran the same suite locally with the same two values and all 8 tests passed. The only thing that differs between those two environments is that CI goes through the proxy and a local run hits the API directly.

That also matches the shape of every observation: consistent regardless of user or organization, and confined to Identity while Feedback, Data Fabric and Functions pass in the same runs — their paths are already whitelisted.

So what my earlier comments read as a stable authorization gap in the Identity service was a stable *routing* block on a new path. Apologies for sending anyone toward a scope grant that would not have helped.

### Remaining work, which is outside this repo

An `identity_/api/Setting` pattern needs adding to `ALLOWED_API_PATTERNS` in `apps-dev-tools`, following [apps-dev-tools#135](https://github.com/UiPath/apps-dev-tools/pull/135). I do not have access to that repository, so here is a draft to paste — the endpoint is organization-level, with no tenant segment:

```typescript
// Identity Service
new RegExp(`^\\/${IDENTITY_BASE}\\/api\\/Setting$`),
```

and a matching case in `CFWorkers/ApiCorsWorker/tests/api-cors-worker.test.ts` under `Path Validation - Allowed Paths`:

```typescript
it('should allow Identity Setting endpoint', async () => {
    const request = createMockRequest('https://api.uipath.com/org/identity_/api/Setting');
    const response = await worker.fetch(request, env as any, ctx);
    expect(response.status).not.toBe(403);
    expect(fetchSpy).toHaveBeenCalled();
});
```

One pattern covers both operations, since `GET` and `PUT` share the path. Worth checking it against however the worker strips leading segments before matching — I am writing this from the conventions rather than from the worker source, which I cannot read.

Once that lands, this suite should pass unmodified. Everything else here is green: lint, 2177 unit tests, build, docs, and four consecutive clean review passes.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/oauth-scopes.md
```diff
@@ -226,6 +226,13 @@ The `ConversationalAgents` scope is required for real-time WebSocket sessions (`
 | `getPolicyTraces()` | `Insights.RealTimeData Insights OR.Folders.Read` |
 | `getOperationSummary()` | `Insights.RealTimeData Insights OR.Folders.Read` |
 
+## Identity
```
vnaren23: Identity is very internal naming. Should we rename it to `Platform` on whatever is developer facing?

kittyyueli: Fair challenge, and the repo's own convention backs the premise — `agent_docs/conventions.md` says *"NEVER use internal product/team names in service file paths or class names — use the user-facing domain name instead."* The `PM.Setting` scope (Platform Management) is corroborating evidence that "Identity" is the internal service name here.

Two things give me pause about `Platform` specifically, so I'd rather get a decision than rename unilaterally:

1. **Every other service in the SDK is named for its domain, not its platform layer** — `./tasks`, `./assets`, `./queues`, `./entities`, `./cases`, `./notifications`, `./governance`, `./functions`. `Platform` names a layer, and it's broad enough that it reads like a home for anything Platform Management ever ships. This PR adds exactly two methods, both of which read/write **user settings** (`UserTheme.Theme`, `UserLanguage.Language`, `UserCase.*`).
2. **The subpath is the part that's expensive to get wrong.** Renaming after publish is a breaking change for consumers, so whatever we pick should be the name we're willing to keep.

Three candidates, in the order I'd rank them:

| Option | Import | Class | Reads as |
|---|---|---|---|
| **A. `user-settings`** | `@uipath/uipath-typescript/user-settings` | `UserSettings` | Describes exactly what the two methods do; every operation here is user-scoped by design |
| **B. `platform`** | `@uipath/uipath-typescript/platform` | `Platform` | Your suggestion — matches the `PM.*` scope family, leaves room for future Platform Management endpoints |
| **C. keep `identity`** | `@uipath/uipath-typescript/identity` | `Identity` | Matches the `identity_/api/Setting` URL, but that's the internal service name the convention warns against |

I lean **A** — it's the only one that describes the capability rather than the team or the layer, which is what the convention is asking for. But **B** is the better call if Platform Management intends to land more surface here (org settings, licensing, tenants) and would rather have one namespace than several, and you have context on that roadmap that I don't.

Whichever you pick I'll do the full rename in one commit — subpath in `package.json` + `rollup.config.js`, class, the seven exported `Identity*` type names, the enum, file/dir names under `src/` and `tests/`, this scopes table, and the `mkdocs.yml` nav entry. Nothing is published yet, so it's a cheap change today and a breaking one later.

---
_Generated by [Claude Code](https://claude.ai/code)_

vnaren23: Lets go with option B since we expect more surfaces to land there like org settings, licensing, tenants etc.

kittyyueli: Done in d13930a — renamed to `Platform` throughout.

```ts
import { Platform, PlatformSettingKey } from '@uipath/uipath-typescript/platform';
```

What moved:

- subpath `./identity` → `./platform` (`package.json`, `rollup.config.js`)
- `IdentityService as Identity` → `PlatformService as Platform`
- the seven exported types → `PlatformServiceModel`, `PlatformSetting`, `PlatformSettingKey`, `PlatformSettingUpsert`, `PlatformSettingGetOptions`, `PlatformSettingMap`, `RawPlatformSetting`
- `@track('Identity.GetSettings' | 'Identity.UpdateSettings')` → `@track('Platform.*')`
- `src/{models,services}/identity/` → `.../platform/`, files renamed to match; same for the unit, integration, mock and constant files
- this scopes section, and the `mkdocs.yml` nav entry (moved to keep the `P`s alphabetical)

Three things I deliberately left on the `identity` name, since they describe the wire and not the SDK surface — say the word if you disagree on any:

1. **`IDENTITY_BASE` / `IDENTITY_API_BASE` and the `identity_/api/Setting` URL.** The endpoint is still served by `identity_`; renaming the constant would make it lie about the path it builds. I did split `PLATFORM_SETTING_ENDPOINTS` out into a new `endpoints/platform.ts`, so `endpoints/identity.ts` is back to holding only the OAuth `connect/*` endpoints it had before this PR.
2. **`IDENTITY_TEST_USER_ID` and its `identityTestUserId` config field.** These track the `UIPATH_IDENTITY_TEST_USER_ID` repository secret that's already provisioned; renaming them needs a secret change and buys nothing developer-facing. There's a comment at the read site explaining why the name differs.
3. **The Cloudflare whitelist entry** is unaffected — the request URL didn't change, so the draft pattern in the PR description still stands as written.

One thing worth deciding now rather than later, given you expect org settings in this namespace: the methods are `getSettings()` / `updateSettings()` and are user-scoped by design (`userId` is always sent — mandatory, no org-scoped variant exposed). When organization settings arrive, `platform.getSettings()` reads ambiguously next to a sibling that fetches org-wide values. If you'd rather pre-empt that, `getUserSettings()` / `updateUserSettings()` is a one-line change today. I haven't done it because you didn't ask for it and the current names are what was already reviewed — happy either way.

Verified after the rename: typecheck clean, oxlint 0 errors, 2177 unit tests pass, `npm run build` emits `dist/platform/{index.mjs,index.cjs,index.d.ts}` exporting `PlatformService as Platform` plus `PlatformSettingKey` as a runtime value, and no `dist/identity/`.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
Correction, and it cuts against something I wrote ~30 minutes ago: **the Platform integration suite passes in CI, and no Cloudflare whitelisting is needed for it.**

I had just rewritten the PR description to name the un-whitelisted proxy path as the blocker. That was wrong. The evidence:

```
✓ tests/integration/shared/platform/platform.integration.test.ts (8 tests) 3057ms
```

and PR Checks on `54df90d` (2026-08-06) was a **full success**. So the 403 was fixed by 54df90d — the commit that started sending the organization GUID on reads — exactly as `PlatformSettingGetOptions.organizationId` documents: omit it and the API targets the *host* partition, which an external app is not authorized for. The fix was in the test, not in anyone's configuration.

That makes three wrong diagnoses of this one 403 from me: first a missing `PM.Setting` scope, then the Cloudflare whitelist, both stated with more confidence than the evidence supported. The description is corrected. **No action is needed from @Sarath1018 or anyone else** — please disregard both the earlier scope request and the whitelist ask. (A whitelist entry may still be wanted for consumers hitting `alpha.api.uipath.com` directly, but that is a separate question and not a blocker here.)

### What is actually failing

3 tests in `tests/integration/shared/orchestrator/processes.integration.test.ts` (`getAll` with filter, `getById`, `start`), all `403 Forbidden`. This PR does not touch Orchestrator Processes, and that file passed on `54df90d`.

Orchestrator folder-scoped reads are 403ing intermittently across unrelated branches — [`feat/plt-107228-cloud-logout`](https://github.com/UiPath/uipath-typescript/actions/runs/31317760885) failed the same job on 2026-08-09 with `403`s on Orchestrator **Buckets** `getById`, `getAll`, and file metadata. A cloud-logout branch cannot break Orchestrator Buckets.

I am **not** claiming a root cause this time — I have not established one. What is verified is that it reproduces on branches that cannot have caused it, and that it is not in code this PR changes. I will re-run the job to see whether it clears; if it does not, it needs someone with visibility into the CI application's Orchestrator folder permissions.

---
_Generated by [Claude Code](https://claude.ai/code)_

vnaren23 (approved):

kittyyueli:
Re-run cleared it — **all 22 checks green**, including `coverage / coverage` and `test-and-build`. The Orchestrator Processes 403s were transient, so nothing needs looking into on the CI application's Orchestrator permissions after all. Thanks for the approval, @vnaren23.

Two things left open in the description, neither blocking — both worth a moment before this publishes, since the second is cheap now and breaking afterwards:

- **Enum member JSDoc is inferred from key names** (e.g. `UserAlert.AlertDuration` documented as "how long alerts remain visible for the user"). Worth a glance from whoever knows the exact semantics.
- **`getSettings()` / `updateSettings()` are user-scoped by design.** If organization settings land in this namespace as you expect, `getUserSettings()` / `updateUserSettings()` would read less ambiguously alongside them. Happy to rename now; say the word.

Not merging — leaving that to you.

---
_Generated by [Claude Code](https://claude.ai/code)_

vnaren23:
> * `getUserSettings()` / `updateUserSettings()`

Renaming to getUserSettings() / updateUserSettings() would be better

kittyyueli:
Done in 3931234 — `getUserSettings()` / `updateUserSettings()`.

Also renamed the `@track` names (`Platform.GetUserSettings`, `Platform.UpdateUserSettings`), the `ValidationError` messages, and the scopes table. The Conversational Agent user service keeps its own `getSettings()` / `updateSettings()` — I caught a sweep that had renamed its scopes-table rows too and reverted them.

Typecheck, lint (0 errors), and 2177 unit tests pass. Heads up that this reset your approval.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
`coverage` is red again, and I said I'd raise it rather than keep re-running, so: **this needs someone who owns the CI environment. It is not this PR.**

The rename itself is validated against the live API — the Platform suite has passed in every run, including both runs on `3931234` with the new method names:

```
✓ tests/integration/shared/platform/platform.integration.test.ts (8 tests) 3695ms
```

### The failure moves between runs

| Run | Failing test | Error |
|---|---|---|
| `d13930a` | Orchestrator Processes ×3 — `getAll` w/ filter, `getById`, `start` | 403 |
| `d13930a` re-run | *(none — all green)* | — |
| `3931234` | Data Fabric Entities `:617` — `entity.insertRecords` → `getById` | 403 |
| `3931234` re-run | Data Fabric Entities `:390` — `getAllRecords > should handle pagination with cursor` | 403 |

Always exactly one failing file, always `403 Forbidden`, and a **different operation each time**. In the latest run the files that failed in earlier runs all passed: Orchestrator Processes (24 ✓), Buckets, Assets, Queues, Jobs (14 ✓), Data Fabric attachments (6 ✓), and the Data Fabric smoke test. It also reproduces on branches that cannot have caused it — [`feat/plt-107228-cloud-logout`](https://github.com/UiPath/uipath-typescript/actions/runs/31317760885) failed with 403s on Orchestrator Buckets on 2026-08-09.

A genuinely missing scope would fail the same operation every time. This does not, so it is unlikely to be permissions — but I am **not** claiming a cause, having already been wrong twice about a 403 on this PR. Possibly worth someone checking whether the shared test environment throttles under the full suite's load, since Data Fabric Entities is by far the heaviest file and has now been hit twice in a row.

### What this means for the PR

Nothing here is fixable from this branch. `test-and-build` is just a mirror of the coverage result, so both red checks are this one failure. Everything else is green: typecheck, lint, unit (2177), build, docs, `regen`, `check-samples`, SonarQube gate, and the automated review.

@Sarath1018 @vnaren23 — happy to re-run again if you'd rather chase green, but it looks like a pre-existing environment issue worth tracking separately rather than something to keep spinning on here.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/platform/platform.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // Scope travels in the query string on reads, but in the body on writes
+    const params: Record<string, string | PlatformSettingKey[]> = { key: keys, userId };
+    if (options?.organizationId) params.partitionGlobalId = options.organizationId;
+
+    const response = await this.get<RawPlatformSetting[]>(
+      PLATFORM_SETTING_ENDPOINTS.SETTINGS,
+      { params }
+    );
+    return response.data.map(toPlatformSetting);
```
Sarath1018: please use `transformData ` for this

src/services/platform/platform.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!organizationId) {
+      throw new ValidationError({ message: 'organizationId is required for updateUserSettings' });
+    }
+
+    const response = await this.put<RawPlatformSetting[]>(PLATFORM_SETTING_ENDPOINTS.SETTINGS, {
+      settings,
+      partitionGlobalId: organizationId,
+      userId,
+    });
+    return response.data.map(toPlatformSetting);
```
Sarath1018: Same, use `transformData` for this

src/services/platform/platform.ts
kittyyueli: Done in 9ec59cc, both call sites.

It was already going through `transformData`, but indirectly — via a module-level `toPlatformSetting` helper that wrapped it and was applied per row with `.map()`. Your comment sent me to look at the signature, and the helper turned out to be redundant: `transformData` takes `T | T[]` and maps element-wise itself, so the array can go straight in.

```typescript
return transformData(response.data, PlatformSettingMap) as unknown as PlatformSetting[];
```

The helper is gone, which also matches the existing idiom — `FeedbackService.getById` does the same thing. Typecheck, lint (0 errors), and 2177 unit tests pass, with the 27 Platform tests still asserting the `partitionGlobalId` → `organizationId` rename on both operations.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/platform/platform.ts
```diff
@@ -0,0 +1,85 @@
...
+ * read-only / write-only access).
+ */
+export class PlatformService extends BaseService implements PlatformServiceModel {
+  @track('Platform.GetUserSettings')
+  async getUserSettings(
+    keys: PlatformSettingKey[],
+    userId: string,
+    options?: PlatformSettingGetOptions
+  ): Promise<PlatformSetting[]> {
+    if (keys.length === 0) {
```
Sarath1018:
```suggestion
    if (!keys?.length) {
```

src/services/platform/platform.ts
```diff
@@ -0,0 +1,85 @@
...
+    return response.data.map(toPlatformSetting);
+  }
+
+  @track('Platform.UpdateUserSettings')
+  async updateUserSettings(
+    settings: PlatformSettingUpsert[],
+    userId: string,
+    organizationId: string
+  ): Promise<PlatformSetting[]> {
+    if (settings.length === 0) {
```
Sarath1018:
```suggestion
    if (!settings?.length) {
```

src/services/platform/platform.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (keys.length === 0) {
+      throw new ValidationError({ message: 'keys must contain at least one setting key' });
+    }
+    if (!userId) {
+      throw new ValidationError({ message: 'userId is required for getUserSettings' });
+    }
+
+    // Scope travels in the query string on reads, but in the body on writes
+    const params: Record<string, string | PlatformSettingKey[]> = { key: keys, userId };
+    if (options?.organizationId) params.partitionGlobalId = options.organizationId;
```
Sarath1018, line 54: Lets validate for non empty `organizationId` as well

src/services/platform/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @example
+ * ```typescript
+ * import { UiPath } from '@uipath/uipath-typescript/core';
+ * import { Platform, PlatformSettingKey } from '@uipath/uipath-typescript/platform';
+ *
+ * const sdk = new UiPath(config);
+ * await sdk.initialize();
+ *
+ * const platform = new Platform(sdk);
+ * const settings = await platform.getUserSettings([PlatformSettingKey.UserTheme], '<userId>');
```
Sarath1018: Isn't this require org id as well

src/services/platform/platform.ts
kittyyueli: Applied in b53005d, along with the matching one on `settings` below. Matches `folder-scoped.ts:123`, which already uses this form.

Added a test per case (`undefined` and `null`, both arguments). I checked they actually cover the guard by reverting it — all four fail with `TypeError: Cannot read properties of undefined`, and pass with `?.` restored. So the change is doing real work for untyped JS consumers, not just reading defensively.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/platform/platform.ts
kittyyueli: Good catch — this was a real footgun. Fixed in b53005d.

The old `if (options?.organizationId)` silently *dropped* an empty string, which is the worst outcome: the request went out with no scope, hit the host partition, and came back `403`. The caller would have had no idea their own empty value caused it.

```typescript
if (options?.organizationId !== undefined && !options.organizationId) {
  throw new ValidationError({ message: 'organizationId must not be empty when provided' });
}
```

Checking against `undefined` rather than falsiness keeps the distinction that matters: omitting the option entirely is still legal (that's what makes it optional), only an explicitly empty value is rejected. Two tests cover it — the empty string rejecting, and options-omitted still sending no scope param — and I confirmed the first fails with the guard reverted.

`updateUserSettings` already validated `organizationId`, since there it's a required positional argument.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/platform/index.ts
kittyyueli: Yes — that example was wrong, and by my own documentation it would have 403'd. Fixed in b53005d:

```typescript
const settings = await platform.getUserSettings([PlatformSettingKey.UserTheme], '<userId>', {
  organizationId: '<organizationId>',
});
```

I swept the rest while I was in there: the three `@example` blocks on `PlatformServiceModel` already passed `organizationId`, so this module-level one was the only place teaching the broken call.

---
_Generated by [Claude Code](https://claude.ai/code)_

kittyyueli:
One more data point on the coverage failure, because it sharpens the signature considerably — then I'll leave it alone.

The run on `b53005d` failed **5 tests across 3 unrelated files**, and the 403s are clustered in time rather than tied to any operation:

| Time | Failure |
|---|---|
| 20:11:01.698 | Orchestrator Buckets `getByName` (folderKey) — 403 |
| 20:11:01.874 | Orchestrator Buckets `getByName` (folderPath) — 403 |
| 20:11:02.959 | Orchestrator Buckets file-ops `getAll` — 403 |
| 20:12:09.838 | Action Center Task Catalogs `fetchById` — 403 |
| — | Data Fabric Entities `getAll` w/ folderKey — **timed out at 30s** |

Three 403s inside 1.3 seconds, across two different services, plus a 30-second timeout elsewhere in the same run. Buckets had passed cleanly in the immediately preceding run; Task Catalogs had passed too. Across the four runs on this branch the failing set has been Orchestrator Processes → *(clean)* → Data Fabric Entities → Buckets + Task Catalogs + a DF timeout.

A burst of 403s across unrelated services in the same second, with a timeout alongside, doesn't look like authorization at all — it looks like the environment briefly refusing or starving requests. That's consistent with the load idea I raised earlier, but I still can't confirm a cause from here and I'm not going to keep guessing at it.

Platform passed again in this run — `✓ platform.integration.test.ts (8 tests) 3742ms` — as it has in every run.

I've stopped re-running. This needs someone with visibility into the shared alpha environment, and it's worth its own issue rather than more cycles here. Everything else on the PR is green.

---
_Generated by [Claude Code](https://claude.ai/code)_

Sarath1018 (approved):

PR #635, opened 2026-08-03
feat(data-fabric): mark entity + choiceset schema-lifecycle methods @experimental

AditiGoyalUipath:
## Summary

- **Marks entity schema-lifecycle methods `@experimental`** — `entities.create()`, `entities.updateById()` / `entity.update()`, `entities.deleteById()` / `entity.delete()`. Previously `@internal`; now surfaced in generated docs and IDE tooltips but tagged unstable.
- **Marks choiceset schema-lifecycle methods `@experimental`** — `choicesets.create()`, `choicesets.updateById()`, `choicesets.deleteById()`. Value-CRUD (`insertValueById` / `updateValueById` / `deleteValuesById`) stays **`@internal`** because those endpoints (`POST /api/EntityService/{name}/choiceset/{insert|update|delete}`) accept only the first-party `DataServiceApiUserAccess` scope — a PAT with `DataFabric.*` scopes cannot reach them.
- **Validates `entities.updateById` input** — three new `ValidationError` guards:
  - No changes at all → throws instead of a silent no-op.
  - `removeFields[i].name` that doesn't exist on the entity → throws instead of silently ignoring.
  - `updateFields[i].id` that doesn't match a non-system field → throws instead of returning the field unchanged.
- **Documents the field-name rule** on `create` / `updateById` / bound `update` JSDoc — the Data Fabric backend rejects underscores in field names; examples updated from `product_name` / `old_field` to `productName` / `oldField`.
- **OAuth scopes doc** — adds `DataFabric.Schema.Write` rows for entity `create` / `updateById` / `deleteById` and choiceset `create` / `updateById` / `deleteById`.
- **Integration test activation** — un-skipped the entity schema-CRUD blocks (create, updateById, deleteById, entity-schema-methods, MULTILINE_MAX lifecycle, RELATIONSHIP field expansion, sqlType constraint defaults) now that the test PAT carries `DataFabric.Schema.Write`. Choiceset schema-lifecycle blocks stay skipped (local run showed 403s on this env's PAT); value-CRUD blocks removed (endpoint gated on a first-party scope). Client-side `ValidationError` tests moved out of integration to unit per `Agents.md`.

Companion Cloudflare Workers whitelist PR: UiPath/apps-dev-tools#131

## Notes

- Endpoint `ENTITY.UPDATE` stays on the v1 URL `/api/Entity/{id}/metadata` — verified against the live API in both v0 and v1 init modes.
- No shared config touched (`vitest.integration.config.ts`, endpoint base paths, etc. unchanged from `main`).

## Test plan

- [x] `npm run test:unit tests/unit/services/data-fabric/entities.test.ts tests/unit/services/data-fabric/choicesets.test.ts` — 250 passed
- [x] Integration: entity `updateById` metadata (v0 + v1) — passes against v1 endpoint
- [x] Four new unit tests cover each throw scenario in `updateById`

### DIFF
--- a/docs/oauth-scopes.md
+++ b/docs/oauth-scopes.md
@@ -70,6 +70,12 @@ This page lists the specific OAuth scopes required in external app for each SDK 
 |--------|-------------|
 | `getAll()` | `DataFabric.Schema.Read` |
 | `getById()` | `DataFabric.Data.Read` |
+| `create()` | `DataFabric.Schema.Write` |
+| `updateById()` | `DataFabric.Schema.Write` |
+| `deleteById()` | `DataFabric.Schema.Write` |
+| `insertValueById()` | `DataFabric.Schema.Write` |
+| `updateValueById()` | `DataFabric.Schema.Write` |
+| `deleteValuesById()` | `DataFabric.Schema.Write` |
 
 ## Maestro Processes
 
--- a/docs/oauth-scopes.md
+++ b/docs/oauth-scopes.md
@@ -60,6 +60,9 @@ This page lists the specific OAuth scopes required in external app for each SDK 
 | `deleteAttachment()` | `DataFabric.Data.Write` |
 | `queryRecordsById()` / `queryRecords()` | `DataFabric.Data.Read` |
 | `importRecordsById()` / `importRecords()` | `DataFabric.Data.Write` |
+| `create()` | `DataFabric.Schema.Write` |
+| `updateById()` / `update()` | `DataFabric.Schema.Write` |
+| `deleteById()` / `delete()` | `DataFabric.Schema.Write` |
 
 ## ChoiceSets
 
--- a/docs/oauth-scopes.md
+++ b/docs/oauth-scopes.md
@@ -73,9 +73,6 @@ This page lists the specific OAuth scopes required in external app for each SDK 
 | `create()` | `DataFabric.Schema.Write` |
 | `updateById()` | `DataFabric.Schema.Write` |
 | `deleteById()` | `DataFabric.Schema.Write` |
-| `insertValueById()` | `DataFabric.Schema.Write` |
-| `updateValueById()` | `DataFabric.Schema.Write` |
-| `deleteValuesById()` | `DataFabric.Schema.Write` |
 
 ## Maestro Processes
 
--- a/src/utils/constants/endpoints/data-fabric.ts
+++ b/src/utils/constants/endpoints/data-fabric.ts
@@ -34,7 +34,7 @@ export const DATA_FABRIC_ENDPOINTS = {
     DELETE_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/delete-batch`,
     UPSERT: `${DATAFABRIC_BASE}/api/Entity`,
     DELETE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}`,
-    UPDATE_METADATA: (entityId: string) => `${DATAFABRIC_BASE}/api/v3/entities/${entityId}/metadata`,
+    UPDATE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}/metadata`,
     QUERY_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/query`,
     // Name-based structured query. The multi-entity (joins) contract is only
     // implemented on this route — QUERY_BY_ID silently drops the `joins` body key.
--- a/src/utils/constants/endpoints/data-fabric.ts
+++ b/src/utils/constants/endpoints/data-fabric.ts
@@ -34,7 +34,7 @@ export const DATA_FABRIC_ENDPOINTS = {
     DELETE_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/delete-batch`,
     UPSERT: `${DATAFABRIC_BASE}/api/Entity`,
     DELETE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}`,
-    UPDATE_METADATA: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}/metadata`,
+    UPDATE_METADATA: (entityId: string) => `${DATAFABRIC_BASE}/api/v3/entities/${entityId}/metadata`,
     QUERY_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/query`,
     // Name-based structured query. The multi-entity (joins) contract is only
     // implemented on this route — QUERY_BY_ID silently drops the `joins` body key.
--- a/tests/integration/shared/data-fabric/choicesets.integration.test.ts
+++ b/tests/integration/shared/data-fabric/choicesets.integration.test.ts
@@ -9,7 +9,6 @@ describe.each(modes)('Data Fabric ChoiceSets - Integration Tests [%s]', (mode) =
   const testConfig = getTestConfig();
   let testChoiceSetId: string | null = testConfig.dataFabricTestChoiceSetId || null;
   const createdChoiceSetIds: string[] = [];
-  const insertedValueIds: string[] = [];
 
   // Folder-scoped CS created in the Folder-scoped operations describe block.
   // Tracked here so the file-level afterAll can clean it up if a test failed
@@ -19,14 +18,6 @@ describe.each(modes)('Data Fabric ChoiceSets - Integration Tests [%s]', (mode) =
 
   afterAll(async () => {
     const { choiceSets } = getServices();
-
-    if (testChoiceSetId && insertedValueIds.length > 0) {
-      try {
-        await choiceSets.deleteValuesById(testChoiceSetId, insertedValueIds);
-      } catch {
-        // Ignore cleanup failures — test resources are sandboxed.
-      }
-    }
 
     if (folderScopedChoiceSetId && folderScopedFolderKey) {
       try {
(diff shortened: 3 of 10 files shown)


tests/integration/shared/data-fabric/entities.integration.test.ts
```diff
@@ -1141,7 +1141,7 @@ describe.each(modes)('Data Fabric Entities - Integration Tests [%s]', (mode) =>
 
       const entityId = await entities.create(name, [
         { name: 'title', displayName: 'Title', type: EntityFieldDataType.STRING, isRequired: true },
-        { name: 'count', displayName: 'Count', type: EntityFieldDataType.INTEGER },
```
vnaren23, line 1192: Why are we changing this?

AditiGoyalUipath: Because there was discrepancies in backend and UI. We have tightened the validation in backend now.

vnaren23: Meaning we dont support integer anymore?

AditiGoyalUipath: We never used to support from UI. UI was only consumption layer to create an entity earlier.  Validation were missing in backend API's.  We support decimal precision as 0. Supporting Integer type in UI is currently not in plan.

vnaren23 (commented):
Address the Claude comments, rest LGTM

AditiGoyalUipath:
> Address the Claude comments, rest LGTM

Few addressed . Few invalid

vnaren23 (approved):

docs/oauth-scopes.md
```diff
@@ -74,13 +60,19 @@ Before running the function, `invoke()` also acquires a Studio Web license for t
...
 
 ## ChoiceSets
 
 | Method | OAuth Scope |
 |--------|-------------|
 | `getAll()` | `DataFabric.Schema.Read` |
 | `getById()` | `DataFabric.Data.Read` |
+| `create()` | `DataFabric.Schema.Write` |
+| `updateById()` | `DataFabric.Schema.Write` |
+| `deleteById()` | `DataFabric.Schema.Write` |
```
swati354, line 75: why do choicesets not have bound methods for update and delete?
can we track this in a separate PR.

src/utils/constants/endpoints/data-fabric.ts
```diff
@@ -34,7 +34,7 @@ export const DATA_FABRIC_ENDPOINTS = {
     DELETE_BY_ID: (entityId: string) => `${DATAFABRIC_BASE}/api/EntityService/entity/${entityId}/delete-batch`,
     UPSERT: `${DATAFABRIC_BASE}/api/Entity`,
     DELETE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}`,
-    UPDATE_METADATA: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}/metadata`,
+    UPDATE: (entityId: string) => `${DATAFABRIC_BASE}/api/Entity/${entityId}/metadata`,
```
swati354, line 37: why this rename? it is technically just metadata update and `UPDATE_METADATA` sits cleaner next to `UPSERT`.

tests/integration/shared/data-fabric/choicesets.integration.test.ts
```diff
@@ -107,8 +98,9 @@ describe.each(modes)('Data Fabric ChoiceSets - Integration Tests [%s]', (mode) =
     });
   });
 
-  // Skipped: choiceset CRUD methods are tagged @internal and validated live on alpha.
-  // Re-enable locally with describe.only when a fresh tenant secret is configured.
+  // Skipped: local run shows tenant-scoped `updateById` / `deleteById` return 403 on
+  // this env (test PAT can create but not update/delete tenant choicesets); un-skip
+  // once the tenant permission model or PAT is aligned.
```
swati354, line 103: would the same not happen for entities integration tests?

tests/integration/shared/data-fabric/entities.integration.test.ts
```diff
@@ -1537,97 +1416,116 @@ describe.each(modes)('Data Fabric Entities - Integration Tests [%s]', (mode) =>
...
-  // marker, getRecordById (v2 read) returns the full content. Creating the field
-  // needs DataFabric.Schema.Write scope, absent in the standard test env — so this is
-  // gated on SCHEMA_WRITE_SCOPE_AVAILABLE and skipped there (a hard throw would redden
-  // CI permanently). Runs in any environment whose PAT carries the scope.
-  describe.skipIf(!getTestConfig().schemaWriteScopeAvailable)('MULTILINE_MAX field lifecycle', () => {
-    it('should return a marker on list and the full value via getRecordById', async () => {
+  // Verifies the MULTILINE_MAX round-trip end to end: create + insert + read the
+  // full value back via getRecordById (v2 read). The list endpoint's size-marker
+  // projection is not asserted — the server threshold is inconsistent and the
+  // consumer-facing guarantee is the full-content read.
```
swati354, line 1523: if the server threshold is inconsistent, lets update the comment [here](https://github.com/UiPath/uipath-typescript/blob/4f331f200693e2c586df26cb99dd3fff19072983/src/models/data-fabric/entities.types.ts#L25) as well.

swati354:
can we update the PR description please
1. > Switches UPDATE_METADATA to /api/v3/entities/{id}/metadata — the v1 URL returns "This API endpoint is not authorized" on Data Fabric v3 tenants.

I see we are back to v1 url. Is the concern in the description not valid now?

2. > @internal was promoted to @experimental, then both tags removed — these ship as stable public API.

still marked as experimental.

tests/integration/shared/data-fabric/entities.integration.test.ts
swati354: If `INTEGER` isn't supported, the enum, `buildSqlTypeConstraints`, and the JSDoc on `maxValue`/`minValue` should also drop it, otherwise the SDK still accepts `INTEGER` and users hit API rejection at runtime.

tests/integration/shared/data-fabric/entities.integration.test.ts
```diff
@@ -1270,49 +1150,49 @@ describe.each(modes)('Data Fabric Entities - Integration Tests [%s]', (mode) =>
...
       const name = `sdk_test_${generateRandomString(8).toLowerCase()}`;
       const entityId = await entities.create(name, [
-        { name: 'base_field', type: EntityFieldDataType.STRING },
+        { name: 'baseField', type: EntityFieldDataType.STRING },
       ]);
       createdEntityIds.push(entityId);
 
       await entities.updateById(entityId, {
-        addFields: [{ name: 'new_field', type: EntityFieldDataType.INTEGER }],
+        addFields: [{ name: 'newField', type: EntityFieldDataType.DECIMAL, decimalPrecision: 0 }],
```
swati354, line 1158: the `entities.create` JSDoc example still uses `product_name`(snake case), and a few of the changed tests still throw `'Could not find to_update field id'` (with the old snake_case name in the message, not the code). What was the reason for the rename?
Does a `new_field` submitted via `create()` gets stored as `newfield`?
 - If it was pure style: please also update the JSDoc examples and fix the stale error strings so tests, docs, and messages agree.
 - If the reason was that snake_case doesn't round-trip cleanly (fields normalized on read, DF expects camelCase, etc.): the JSDoc examples in `entities.models.ts` are misleading, users who copy `product_name` will hit the same problem the tests were just patched to avoid. Please update the JSDoc and add a note explaining the constraint.

swati354:
> > Address the Claude comments, rest LGTM
> 
> Few addressed . Few invalid

Small ask on Claude review comments - could we treat them the same way we'd treat comments from a human reviewer?  A quick 👍  (or "fixed in commit X") if it's valid, a one-liner if it isn't, then resolve. Silent resolutions read the same either way and we lose the signal.
Another reason for this is it would also help us gauge how much of the review comments by claude are just noise, so we can tune it better.

docs/oauth-scopes.md
AditiGoyalUipath: choiceset already have these method and already included in the PR. Only Choiceset value scope is still not added for choiceset API. Once they are available in prod then it will be added

src/utils/constants/endpoints/data-fabric.ts
AditiGoyalUipath: Every update is metadata update only . Technically speaking Any changes in Entity is Metadata inclusive of field changes etc are considered metadata update itself. Since We have Field update part of Upsert and We don't have any Update method why to keep specific method which is UPDATE_METADATA instead of Update

tests/integration/shared/data-fabric/choicesets.integration.test.ts
AditiGoyalUipath: No because scopes are added for entity

tests/integration/shared/data-fabric/entities.integration.test.ts
AditiGoyalUipath: This PR is not related to Multiline max field . All integration tests comments are added by claude. There is another task on owner to modify SDK for the changes at server. We also need to make some standards on how to reflect changes done in API to SDK

tests/integration/shared/data-fabric/entities.integration.test.ts
AditiGoyalUipath: There are in changes going on in datafabric.
  _ in field name wasn't supported in UI. 
Just earlier the validation was not present in backend . Since this validation is added recently . There is change here. 
Will update the Doc as well

AditiGoyalUipath:
> > > Address the Claude comments, rest LGTM
> > 
> > 
> > Few addressed . Few invalid
> 
> Small ask on Claude review comments - could we treat them the same way we'd treat comments from a human reviewer? A quick 👍 (or "fixed in commit X") if it's valid, a one-liner if it isn't, then resolve. Silent resolutions read the same either way and we lose the signal. Another reason for this is it would also help us gauge how much of the review comments by claude are just noise, so we can tune it better.

I have replied or mark resolved. Claude reviewer needs to perform better to be taken seriously here. Currently it is becoming overhead.

AditiGoyalUipath:
> I see we are back to v1 url. Is the concern in the description not valid now?
> 
> 2. >

V3 API not available in Prod. Once available we will move to v3

There is active work going in datafabric where all the APi's are being modified. We still want to keep these experimental until we are sure these are stable

AditiGoyalUipath:
@swati354 The integeration tests are failing with unrelated changes Can you please look into that as well

swati354 (approved):

PR #636, opened 2026-08-03
feat(auth): end the platform session on logout (OIDC RP-initiated logout) [PLT-107228]

shivendra6720:
https://github.com/user-attachments/assets/8e7a220f-0d2f-4b3f-a45b-07c8eaf60000

## What this adds

`sdk.logout()` previously cleared only local state: the UiPath platform (Identity) session survived, so the next sign-in silently re-authenticated — users could not actually sign out or switch accounts. `logout(options?)` now supports OIDC RP-initiated logout:

| Option | Type | Behavior |
|--------|------|----------|
| `endSession` | `boolean` | Clears local auth state, then redirects the browser to `identity_/connect/endsession`, terminating the UiPath platform session (Automation Cloud or Automation Suite). The browser returns to the configured `redirectUri` — registered with Identity by definition, so the exact-match validation always passes (at hosted runtime it comes from the platform meta tags). The next sign-in prompts for credentials. Browser-only; default `false` (previous behavior, byte-for-byte). |

### `openid` is requested automatically

Identity's end-session endpoint is unauthenticated — the OIDC ID token (`id_token_hint`) is what proves the request (confirmed by the Identity team). The SDK now appends `openid` (alongside the existing `offline_access`) to every authorization request, so new sign-ins always receive an ID token — **no configuration needed**. `openid` is an identity resource that non-confidential clients can request without it being selected on the app registration (verified on alpha with a manually-registered app that doesn't list it; confirmation for older registrations is with the Identity team).

The end-session redirect happens **only when an ID token is available**:

| | Session without an ID token (signed in before this version) | Session with an ID token |
|---|---|---|
| `logout({ endSession: true })` | local state cleared only; `console.warn("endSession skipped — no OIDC ID token is available for this session. Sign in again to enable it…")`; **no navigation** | silent sign-out via `id_token_hint`, browser returns to `redirectUri` |

`client_id` is never sent — there is no fallback path. The ID token is captured from token responses (`AuthToken.id_token` → `TokenInfo.idToken` → `TokenManager.getIdToken()`), persisted with the token across page reloads, and kept across token refreshes (a refresh response may omit `id_token`; the one from login is retained).

Secret-based (confidential) auth skips end-session entirely — confidential apps cannot hold `openid` and have no user session to end.

## Compatibility

- `logout()` with no arguments is byte-for-byte the previous behavior; the new parameter is optional, so every existing call site compiles and behaves unchanged (verified by compiling released-shape consumers against the new `dist/`).
- `TokenInfo` gains an optional `idToken` field — existing literals and consumers are unaffected.
- Authorization requests now always include `openid offline_access`. Already-issued access tokens remain valid until they expire — logout does not revoke them.

## Identity dependency — shipped

[PLT-108129](https://uipath.atlassian.net/browse/PLT-108129): Identity previously dropped `post_logout_redirect_uri` for External Application clients and fell back to the portal. The fix (merged, deployed to alpha 2026-08-06) validates the passed URI and honors it.

## Example Usage

```typescript
// Local logout only (default — previous behavior)
sdk.logout();

// Also end the UiPath session; the browser is redirected and returns
// to the configured redirectUri
sdk.logout({ endSession: true });
```

[PLT-108129]: https://uipath.atlassian.net/browse/PLT-108129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/authentication.md
```diff
@@ -181,7 +160,7 @@ useEffect(() => {
...
 - `sdk.getToken()` - Get the logged-in user's access token
-- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again)
+- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again). By default this is local-only: the Automation Cloud session is untouched and a subsequent sign-in completes silently. Pass `sdk.logout({ endSession: true })` to also terminate the Automation Cloud session — the browser is redirected to the Identity end-session endpoint, so the next sign-in prompts for credentials. **The cloud logout requires the `openid` scope in your SDK configuration**: the SDK sends the OIDC ID token as `id_token_hint` to prove the end-session request; without an ID token the SDK clears local state only, logs a warning, and skips the cloud logout. Pass `postLogoutRedirectUri` to return the user to your app afterwards — e.g. `(window.location.origin + window.location.pathname).replace(/\/$/, '')`. Identity compares the value by exact string match (a trailing-slash difference is a mismatch); when omitted or mismatched, the user lands on the Automation Cloud portal.
```
Raina451: please use concise comments. and exclude things like openid, id_token_hint etc. these will be reflected in public facing docs.

Raina451: > sdk.logout({
>   endSession: true,
>   postLogoutRedirectUri: (window.location.origin + window.location.pathname).replace(/\/$/, '')
> });

why do users need to pass postLogoutRedirectUri ? 

they have already provided redirect url at time of initialisation of sdk. we should use the same.
and at hosted runtime, it should continue to read from meta tags

samples/data-fabric-app/src/context/AuthContext.tsx
```diff
@@ -73,8 +73,30 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
-    sdk.logout()
+    sdk.logout({
+      endSession: true,
+      // Identity validates this against the app's registered redirect URIs
+      // by EXACT string match, and registrations carry no trailing slash —
+      // on localhost, pathname is "/" and "http://localhost:5173/" would be
+      // silently rejected (landing the user on the portal). Strip it.
+      postLogoutRedirectUri: (
+        window.location.origin + window.location.pathname
+      ).replace(/\/$/, ''),
```
Raina451: how does end session work in node env non browser env?

src/core/auth/types.ts
```diff
@@ -6,6 +6,42 @@ export interface TokenInfo {
...
+   * next sign-in prompts for credentials instead of silently reusing the
+   * still-active cloud session. Browser-only. Defaults to false, which
+   * preserves the previous local-only logout behavior.
+   *
+   * Requires the `openid` scope in your SDK configuration: the SDK captures
+   * the OIDC ID token and sends it as `id_token_hint`, which is what proves
+   * the end-session request. Without an ID token the cloud logout is
+   * skipped — only local state is cleared — and a warning is logged.
+   */
+  endSession?: boolean;
```
Raina451, line 28: endsession is not intutive naming, consider endCloudSession or something like that.
CC: @vnaren23 thoughts?

src/core/auth/types.ts
```diff
@@ -6,6 +6,42 @@ export interface TokenInfo {
...
+   * redirected to the Identity end-session endpoint, terminating the
+   * Automation Cloud session (and invalidating the refresh token) so the
+   * next sign-in prompts for credentials instead of silently reusing the
+   * still-active cloud session. Browser-only. Defaults to false, which
+   * preserves the previous local-only logout behavior.
+   *
+   * Requires the `openid` scope in your SDK configuration: the SDK captures
+   * the OIDC ID token and sends it as `id_token_hint`, which is what proves
+   * the end-session request. Without an ID token the cloud logout is
+   * skipped — only local state is cleared — and a warning is logged.
```
Raina451: use concise jsdoc comments.

Raina451: everywhere in this pr

Raina451 (changes requested):

src/core/auth/types.ts
vnaren23: `revokeToken` maybe

Raina451: logging out of automation cloud doesnt revoke access token. so `revokeToken` could be misleading.

vnaren23: Isnt that what identity provided? I thought that was the case.

Raina451: they just added support for post_logout_redirect_uri in endpoint afaik.

vnaren23: Nvm, I was wrong. They only fixed the post redirect url. https://uipath-product.slack.com/archives/C0175DZP4PQ/p1785230568455659

`endCloudSession` - why `cloud`? This will work on Automation Suite also right?

Raina451: right, should be the same for AS

vnaren23: Actually lets test whether the token is getting revoked since the flow is `OIDC RP-initiated logout`

Raina451: i did test, token was valid post logout of automation cloud. @shivendra6720 you can test once.

shivendra6720: Tested this: captured an API
call before logout, performed the cloud logout, replayed the
byte-identical request one minute later - the access token
is NOT revoked and stays valid until its natural expiry (1h).

samples/data-fabric-app/src/context/AuthContext.tsx
shivendra6720: it does not apply there by design, in the node env/non browser env sdk.logout() just simply returns and does not perform anything.

`public logout(options?: LogoutOptions): void {
    // Secret-based auth has no session to end — skip silently
    if (this.#config && hasSecretConfig(this.#config)) {
      return;
    }
    this.#authService?.logout(options);
    this.#initialized = false;
  }`

docs/authentication.md
```diff
@@ -181,7 +160,7 @@ useEffect(() => {
 - `sdk.isInOAuthCallback()` - Check if processing OAuth redirect
 - `sdk.completeOAuth()` - Manually complete OAuth (advanced use)
 - `sdk.getToken()` - Get the logged-in user's access token
-- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again)
+- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again). By default the Automation Cloud session stays active, so the next sign-in completes silently. Pass `sdk.logout({ endCloudSession: true })` to also sign the user out of Automation Cloud — the browser is redirected to end the session and returns to your configured `redirectUri` (override with `postLogoutRedirectUri`).
```
Raina451: should we use "Automation Cloud session" ? or UiPath session? because this will land for Automation Suite also

src/core/auth/types.ts
Raina451: what name have we finalised then? cant put endCloudSession also as it will be for Automation suite also.
what about `endUiPathSession` ?

vnaren23: Isnt `endSession` itself better? `endUiPathSession` could also be confusing between coded app session vs cloud.uipath.com (or AS) session

docs/oauth-scopes.md
```diff
@@ -2,6 +2,13 @@
 
 This page lists the specific OAuth scopes required in external app for each SDK method.
 
+## Authentication
+
+| Method | OAuth Scope |
+|--------|-------------|
+| `logout()` | None — clears local authentication state only (default behavior) |
+| `logout({ endCloudSession: true })` | `openid` — without it the cloud logout is skipped (local state is still cleared) |
```
Raina451: > `openid` 

end user consuming this sdk, shouldnt need to pass this scope. why cant we always add this openid scope?

src/core/auth/types.ts
Raina451: > Isnt endSession itself better? 

just by looking at the name, it is not clear. sdk is providing logout method for logging out of coded app, if we dont want them to differentiate between coded app vs cloud/AS then logout should be doing endsession by default.

vnaren23: In theory yes but we cant do this since this will cause app to behave differently once SDK version is upgraded
> should be doing endsession by default.

Raina451: I am okay with endSession, only because I cant think of any better name.

shivendra6720: lets go ahead with endSession only then

docs/oauth-scopes.md
shivendra6720: openid has a failure mode,if an app's External App
registration doesn't include openid in its allowed scopes, requesting it at
/authorize fails the whole login with invalid_scope - so apps that log in fine
today would break on SDK upgrade.

Identity told us openid "should be" default-allowed for
non-confidential apps, but that might not be guaranteed for existing registration or older registration, lets keep it opt-in for now (the console
warning makes the missing scope discoverable in one test), and will check with Identity
and if openid is guaranteed-allowed on ALL non-confidential registrations,
auto-appending becomes a safe, non-breaking follow-up.

docs/authentication.md
```diff
@@ -181,7 +160,7 @@ useEffect(() => {
 - `sdk.isInOAuthCallback()` - Check if processing OAuth redirect
 - `sdk.completeOAuth()` - Manually complete OAuth (advanced use)
 - `sdk.getToken()` - Get the logged-in user's access token
-- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again)
+- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again). By default the UiPath session (Automation Cloud or Automation Suite) stays active, so the next sign-in completes silently. Pass `sdk.logout({ endSession: true })` to also sign the user out of UiPath session — the browser is redirected to end the session and returns to your configured `redirectUri` (override with `postLogoutRedirectUri`).
```
Raina451:
```suggestion
- `sdk.logout()` - Logout and clear all authentication state (requires re-initialization to authenticate again). By default the UiPath session (Automation Cloud or Automation Suite) stays active, so the next sign-in completes silently. Pass `sdk.logout({ endSession: true })` to also sign the user out of UiPath session — the browser is redirected to end the session and returns to your app.
```

src/core/auth/service.ts
```diff
@@ -242,14 +254,30 @@ export class AuthService {
 
   /**
    * Clears all authentication state including tokens and stored OAuth context.
+   * With `endSession: true`, additionally redirects the browser to the
+   * Identity end-session endpoint to terminate the UiPath platform session
+   * (Automation Cloud or Automation Suite). Requires the OIDC ID token
+   * (`openid` scope) — without one the session logout is skipped and a
+   * warning is logged.
```
Raina451: have concise comment

src/core/auth/service.ts
```diff
@@ -242,14 +254,30 @@ export class AuthService {
...
+    // Capture the ID token before clearToken() wipes it.
+    const idTokenHint = options?.endSession ? this.tokenManager.getIdToken() : undefined;
+
+    // End-session is an unauthenticated Identity endpoint — id_token_hint is
+    // what proves the request, so without `openid` there is nothing to send.
+    if (options?.endSession && isBrowser && !idTokenHint) {
+      console.warn(
+        'End session skipped: no OIDC ID token is available, so only local ' +
+        "authentication state was cleared. Add the 'openid' scope to your " +
+        'SDK configuration to enable endSession.'
```
Raina451: please have actionable warning.

src/core/auth/types.ts
```diff
@@ -6,6 +6,31 @@ export interface TokenInfo {
...
+   * is cleared and a warning is logged. Defaults to false (local-only
+   * logout).
+   */
+  endSession?: boolean;
+  /**
+   * URL the user returns to after the platform session ends. Defaults to
+   * the configured `redirectUri`. Must exactly match a redirect URI
+   * registered for the external app.
+   */
+  postLogoutRedirectUri?: string;
```
Raina451: is postLogoutRedirectUri still needed ?

docs/oauth-scopes.md
Raina451: > so apps that log in fine today would break on SDK upgrade.

No why ? I am asking that sdk should when making request to authorise endpoint alway add openid scope , removing dependency on user needs to provide. 

if they upgrade sdk, new sdk version will always be using openid scope for getting token, and even logout endsession also is only available in new version, so neither way this breaks on upgrade.

docs/oauth-scopes.md
```diff
@@ -2,6 +2,13 @@
 
 This page lists the specific OAuth scopes required in external app for each SDK method.
 
+## Authentication
+
+| Method | OAuth Scope |
+|--------|-------------|
+| `logout()` | None — clears local authentication state only (default behavior) |
+| `logout({ endSession: true })` | `openid` — requested automatically by the SDK; no configuration needed |
```
Raina451: still seeing openid scope as requirement, I believe we are always passing this in request and is not needed to be passed by users.

shivendra6720: its mentioned right that its requested automatically by sdk no configuration needed.

Raina451: oh misread it.

Raina451: please remove this at all, because we only mention scopes here which user needs to grant. so lets not add it here

src/core/auth/service.ts
```diff
@@ -258,6 +283,15 @@ export class AuthService {
         console.warn('Failed to clear OAuth context from session storage', error);
       }
     }
+
+    if (options?.endSession && isBrowser && idTokenHint) {
+      window.location.href = this._buildEndSessionUrl({
+        idTokenHint,
+        // The configured redirectUri is registered with Identity by
+        // definition, so it passes the exact-match validation.
+        postLogoutRedirectUri: this.config.redirectUri
```
Raina451, line 280: please check if we need to add endsession here: https://github.com/UiPath/apps-dev-tools/blob/main/CFWorkers/ApiCorsWorker/api-cors-worker.ts#L691-L696

refer this pr: https://github.com/UiPath/apps-dev-tools/pull/39

Raina451 (approved):

samples/data-fabric-app/src/context/AuthContext.tsx
```diff
@@ -73,9 +73,15 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+  /**
+   * Signs the user out of the app AND the UiPath platform: local tokens are
+   * cleared, then the browser is redirected to end the session and returned
+   * to the configured `redirectUri`. For an app-only logout — platform
+   * session survives, next sign-in is silent — call `sdk.logout()` with no
+   * options instead.
+   */
   const logout = () => {
-    sdk.logout()
-    setIsAuthenticated(false)
```
vnaren23, line 78: Why have we removed this?

vnaren23 (approved):

samples/data-fabric-app/src/context/AuthContext.tsx
shivendra6720: to not show the signin page as soon as we click sign out and once the signout is complete a fresh mount is done and then IsAuthenticated is false only and then we see the signin page.

src/core/auth/service.ts
shivendra6720: yes needed, raised the pr (https://github.com/UiPath/apps-dev-tools/pull/153)

PR #643, opened 2026-08-05
feat(queues): add queue item operations [PLT-104203]

shivendra6720:
## Method Added

| Layer | Method | Signature |
|-------|--------|-----------|
| Service | `queues.getAllWithMethods()` | `getAllWithMethods<T extends QueueGetAllWithMethodsOptions>(options?: T): Promise<T extends HasPaginationOptions<T> ? PaginatedResponse<QueueGetWithMethodsResponse> : NonPaginatedResponse<QueueGetWithMethodsResponse>>` |
| Service | `queues.getByIdWithMethods()` | `getByIdWithMethods(id: number, options?: QueueGetByIdWithMethodsOptions): Promise<QueueGetWithMethodsResponse>` |
| Service | `queues.getByName()` | `getByName(name: string, options?: QueueGetByNameOptions): Promise<QueueGetWithMethodsResponse>` |
| Service | `queues.getByKey()` | `getByKey(key: string, options?: QueueGetByKeyOptions): Promise<QueueGetWithMethodsResponse>` |
| Service | `queues.getAllItems()` | `getAllItems<T extends QueueGetAllItemsOptions>(queueId: number, options?: T): Promise<T extends HasPaginationOptions<T> ? PaginatedResponse<QueueItem> : NonPaginatedResponse<QueueItem>>` |
| Service | `queues.insertItemByName()` | `insertItemByName(queueName: string, specificData: Record<string, QueueItemValue>, options?: QueueInsertItemOptions): Promise<QueueItem>` |
| Bound | `queue.getAllItems()` | `getAllItems(options?): Promise<…QueueItem…>` — folder comes from the queue |
| Bound | `queue.insertItem()` | `insertItem(specificData, options?): Promise<QueueItem>` — folder comes from the queue |

## Released methods are untouched and deprecated

`getAll` and `getById` are **byte-for-byte identical to the released SDK** (machine-verified against the fork commit — they render as pure unchanged context in this diff) and keep returning plain `QueueGetResponse` data: enumerable, `structuredClone`/`postMessage`-safe, exactly as customers have it today. They carry `@deprecated` JSDoc pointing at their successors (the `Buckets.uploadFile` deprecation precedent), and unit tests pin the plain-data contract so it cannot regress.

The successors `getAllWithMethods` / `getByIdWithMethods` (naming mirrors #416's `CaseGetAllWithMethodsResponse` type convention) attach the operational methods (`queue.getAllItems()`, `queue.insertItem()`) and add full folder scoping.

[...]

PR #644, opened 2026-08-05
feat(queues): add transaction operations [PLT-104203]

shivendra6720:
Stacked on #643 — **base is `feat/sdk-plt-104203`**, so the diff here is transactions only. Rebase onto `main` once #643 merges.

## Method Added

| Service | `queues.startTransaction()` | `startTransaction(queue: QueueRef, options?: QueueStartTransactionOptions): Promise<QueueItem | null>` |
| Service | `queues.completeTransaction()` | `completeTransaction(itemId: number, outcome: QueueTransactionOutcome, options?: QueueCompleteTransactionOptions): Promise<void>` |
| Bound | `queue.startTransaction()` | `startTransaction(): Promise<QueueItem | null>` |
| Bound | `queue.completeTransaction()` | `completeTransaction(itemId: number, outcome: QueueTransactionOutcome, options?: Omit<QueueCompleteTransactionOptions, 'folderId' | 'folderKey' | 'folderPath'>): Promise<void>` |

## Endpoint Called

| `startTransaction()` | POST | `/orchestrator_/odata/Queues/UiPathODataSvc.StartTransaction` | `OR.Queues` or `OR.Queues.Write` |
| `completeTransaction()` | POST | `/orchestrator_/odata/Queues({itemId})/UiPathODataSvc.SetTransactionResult` | `OR.Queues` or `OR.Queues.Write` |

- Extends `FolderScopedService` — sets the `X-UIPATH-OrganizationUnitId` header from `folderId`
- **Resource Resolution Framework** (per the [design doc](https://uipath.atlassian.net/wiki/spaces/CLD/pages/90924515367/Resource+Resolution+Framework+for+CRUD+and+Operational+SDK+APIs)): `startTransaction` takes a `QueueRef` selector — `{ name }` or `{ id }`, mutually exclusive via `never`-typing (`ResourceRef<TId>`, first use, added to `common/types.ts`). The wire API only accepts a queue *name*, so an `id` selector is resolved first via `GET /odata/Queues({id})?$select=Name` (one extra lookup)
- **Folder scoping via options**: both methods take `folderId` / `folderKey` / `folderPath` in options, routed through the shared `resolveFolderHeaders` (with the coded-app meta-tag folder-key fallback) — same pattern as `assets.updateValueById`
- `outcome` is positional on `completeTransaction` (`QueueTransactionOutcome.Successful` / `.Failed`) — it is the caller's verdict on its own processing, and `processingError` stays optional on failure (the API requires only the outcome); `QueueCompleteTransactionOptions` holds the optional completion fields; failure details go in `processingError` (typed with the `QueueExceptionType` enum), `Output`/`Analytics` payloads are attached without case conversion, `Date` fields are serialized to ISO-8601
- Required params are guarded with `ValidationError` before any HTTP call
- Both responses run through the same `transformQueueItem` transform as the queue-item methods

## `startTransaction` requires a robot session

Orchestrator allocates the item to the robot that sent the request — the spec defines `RobotIdentifier` as *"the unique key identifying the robot that sent the request"*, and the official API guide states: *"Queue items can only be processed by Robots. For example, making a POST request to the `/odata/Queues/UiPathODataSvc.StartTransaction` endpoint requires information which is only available to the Robot."* User and application identities therefore always receive `null`, however many items are waiting (verified live: four 204s with a user PAT while eligible items existed). `RobotIdentifier` is deliberately not exposed — only a robot can truthfully supply one, and a robot session already identifies itself through its token.

`null` covers both "no eligible items" and "no allocation target" — Orchestrator answers 204 in either case and the two are not distinguishable. This is documented in the JSDoc with an example, and the integration test is `it.skip` with the same explanation, per the auth-limitation rule in `agent_docs/rules.md` (the suite authenticates with a PAT).

Browser callers additionally need UiPath/apps-dev-tools#137 (open), which whitelists `StartTransaction` in the API CORS worker; `SetTransactionResult` is already whitelisted on its main.

## Example Usage

```typescript
import { Queues, QueueExceptionType, QueueTransactionOutcome } from '@uipath/uipath-typescript/queues';

await sdk.initialize();
const queues = new Queues(sdk);

// Acquire the next item (robot sessions; user identities receive null)
const transaction = await queues.startTransaction({ name: '<queueName>' }, { folderId: <folderId> });

// or select by ID — the SDK first resolves the queue's name (one extra lookup)
const byId = await queues.startTransaction({ id: <queueId> }, { folderId: <folderId> });

if (transaction) {
  // Running under a robot session: the item is now locked to this caller
  console.log(transaction.status);        // 'InProgress'
  console.log(transaction.specificData);  // the item's business payload
} else {
  // No item was acquired: no eligible items — or the caller is a
  // user/application identity with no robot session to allocate to.
}

// Report success with output data
await queues.completeTransaction(<itemId>, QueueTransactionOutcome.Successful, {
  folderId: <folderId>,
  outputData: { paymentId: 'P-778' }
});

// Report a business failure (not retried)
await queues.completeTransaction(<itemId>, QueueTransactionOutcome.Failed, {
  folderId: <folderId>,
  processingError: {
    reason: 'Vendor not found',
    type: QueueExceptionType.BusinessException
  }
});

// Or bound to a queue from getByIdWithMethods/getAllWithMethods
const queue = await queues.getByIdWithMethods(<queueId>, { folderId: <folderId> });
const item = await queue.startTransaction();
if (item) {
  await queue.completeTransaction(item.id, QueueTransactionOutcome.Successful, { outputData: { Result_Code: 'OK' } });
}
```

### Transform pipeline
Same pipeline as the queue-item methods (see #643): extract `SpecificContent`/`Output` → drop the JSON-string duplicates → `pascalToCamelCaseKeys` → `transformData(QueueItemMap)` → nested `QueueItemProcessingErrorMap` → reattach payloads verbatim. A `startTransaction` response is a `QueueItem` in `InProgress` status with `processingStartTime` set.

### Field mapping (request side, `completeTransaction`)
| SDK Option (camelCase) | API Request (PascalCase) | Change | Reason |
|------------------------|--------------------------|--------|--------|
| `outcome` (positional `QueueTransactionOutcome`) | `IsSuccessful` | Enum → boolean | caller's processing verdict |
| `processingError` | `ProcessingException` | Case + Rename | SDK error-field naming; nested `reason`/`details`/`type` → `Reason`/`Details`/`Type` |
| `deferDate` / `dueDate` (`Date`) | `DeferDate` / `DueDate` (ISO string) | Case + serialization | SDK accepts `Date` objects |
| `outputData` / `analytics` | `Output` / `Analytics` | Rename only | user-defined keys — **never case-converted** |
| `progress`, `operationId` | `Progress`, `OperationId` | Case only | — |

## Sample SDK Response

<details>
<summary><code>startTransaction()</code> — acquired item</summary>

```json
{
  "id": 22905688,
  "key": "8f4f2f6e-...",
  "status": "InProgress",
  "reviewStatus": "None",
  "priority": "Normal",
  "queueId": 346720,
  "specificData": { "InvoiceId": "INV-INLINE-1" },
  "outputData": null,
  "processingError": null,
  "reference": "PW-REF-INLINE",
  "createdTime": "2026-08-03T16:28:58.407Z",
  "processingStartTime": "2026-08-03T16:28:58.582Z",
  "processingEndTime": null,
  "retryNumber": 0,
  "folderId": 3027005
}
```

*From live E2E on alpha (trimmed) — note `status: InProgress` and `processingStartTime` set at acquisition. When no item is available (or the caller has no robot session), the method returns `null` instead.*

</details>

<details>
<summary><code>completeTransaction()</code></summary>

```json
{ "success": true }
```

*`SetTransactionResult` returns no content — the SDK returns `Promise<void>` (matching `deleteRecordById`). Verified live: success with output persisted (keys exactly as provided), BusinessException (no retry), ApplicationException (retry copy created with `ancestorId` linkage), and rejection of an outcome change on a terminal item.*

</details>

## Files

| Area | Files |
|------|-------|
| Endpoint | `src/utils/constants/endpoints/orchestrator.ts` (`START_TRANSACTION`, `SET_TRANSACTION_RESULT`) |
| Types | `src/models/orchestrator/queues.types.ts` (`QueueRef`, `QueueStartTransactionOptions`, `QueueCompleteTransactionOptions`), `src/models/common/types.ts` (`ResourceRef<TId>`) |
| Models | `src/models/orchestrator/queues.models.ts` (ServiceModel + bound methods, factory delegates) |
| Service | `src/services/orchestrator/queues/queues.ts` (`startTransaction` + `resolveQueueName`, `completeTransaction`) |
| Unit tests | `tests/unit/services/orchestrator/queues.test.ts` (37), `tests/unit/models/orchestrator/queues.test.ts` (12) |
| Integration tests | `tests/integration/shared/orchestrator/queues.integration.test.ts` (11 tests × v0/v1 modes; 20 passing / 2 skipped live against alpha) |
| Docs | `docs/oauth-scopes.md`, `agent_docs/conventions.md` |

Refs PLT-104203


src/models/orchestrator/queues.models.ts
```diff
@@ -172,6 +174,81 @@ export interface QueueServiceModel {
...
+   * } else {
+   *   // No item was acquired. This happens when the queue has no eligible
+   *   // items — and always for user/application identities (e.g. a coded app
+   *   // signed in with OAuth), which have no robot session for Orchestrator
+   *   // to allocate the item to.
+   *   console.log('Nothing to process');
+   * }
+   * ```
+   */
+  startTransactionByName(queueName: string, folderId: number): Promise<QueueItem | null>;
```
deepeshrai-tech: for operationalMethods it should be like startTransaction({name?,id?key?},options:{folderScopingOptions}), check once which identifiers are supported

src/models/orchestrator/queues.models.ts
```diff
@@ -172,6 +174,81 @@ export interface QueueServiceModel {
...
+   *   processingError: {
+   *     reason: 'Vendor not found',
+   *     type: QueueExceptionType.BusinessException
+   *   }
+   * });
+   * ```
+   */
+  completeTransaction(
+    itemId: number,
+    folderId: number,
```
deepeshrai-tech: use folderscopedoptions, check for other similar places as well

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +201,81 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+    if (!queueName) {
+      throw new ValidationError({ message: 'queueName is required for startTransactionByName' });
+    }
+    if (!folderId) {
+      throw new ValidationError({ message: 'folderId is required for startTransactionByName' });
+    }
+
+    // RobotIdentifier is deliberately not exposed: the API defines it as the key
+    // of the robot that sent the request, so only a robot can supply one, and a
+    // robot session already identifies itself through its token.
```
swati354, line 363: what if a user wants to pass a robot-identifier?

src/models/orchestrator/queues.types.ts
```diff
@@ -223,3 +223,40 @@ export interface QueueItem {
   /** Folder display path — populated when listing items; `null` on insert responses */
   folderName: string | null;
 }
+
+/**
+ * Optional details accompanying a transaction completion. The success flag is
+ * passed positionally to `completeTransaction`.
+ */
+export interface TransactionCompletionOptions {
```
swati354: `TransactionCompletionOptions` -> `QueueCompleteTransactionOptions` ?

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +201,81 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+    const response = await this.post<Record<string, unknown> | undefined>(
```
swati354, line 364: > this.post<Record<string, unknown> | undefined>

is the undefined necessary?

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +201,81 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+      Progress: options.progress,
+      OperationId: options.operationId
+    };
+    // Output/Analytics hold user-defined keys — attached unchanged.
+    if (options.outputData !== undefined) {
+      transactionResult.Output = options.outputData;
+    }
+    if (options.analytics !== undefined) {
+      transactionResult.Analytics = options.analytics;
+    }
```
swati354: why do these fields have separate checks, while the others are set unconditionally?

tests/integration/shared/orchestrator/queues.integration.test.ts
```diff
@@ -259,5 +259,46 @@ describe.each(modes)('Orchestrator Queues - Integration Tests [%s]', (mode) => {
       expect(page.items.length).toBeLessThanOrEqual(1);
       expect(page.totalCount).toBeGreaterThan(0);
     });
+
+    // skip: get-next allocation returns the item to the robot that requested it,
+    // so it cannot be exercised by the PAT-authenticated suite — a user identity
+    // is not a robot and always receives null. Body kept intact so the test runs
+    // once the suite can authenticate as a robot.
+    it.skip('should start and complete a transaction against the queue', async () => {
```
swati354, line 332: would this be solved if we allow passing robot-identifier in the body?

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +201,81 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+    return transformQueueItem(response.data);
+  }
+
+  @track('Queues.CompleteTransaction')
+  async completeTransaction(
+    itemId: number,
+    folderId: number,
+    isSuccessful: boolean,
+    options: TransactionCompletionOptions = {}
+  ): Promise<OperationResponse<void>> {
```
swati354: why not just return `Promise<void>`?

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +201,81 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+      ProcessingException: options.processingError && {
+        Reason: options.processingError.reason,
+        Details: options.processingError.details,
+        Type: options.processingError.type,
+        AssociatedImageFilePath: options.processingError.associatedImageFilePath
+      },
+      DeferDate: options.deferDate?.toISOString(),
+      DueDate: options.dueDate?.toISOString(),
+      Progress: options.progress,
+      OperationId: options.operationId
```
swati354: why can't we use `camelToPascalCaseKeys` method here? Something like -
```
const transactionResult = camelToPascalCaseKeys({
      isSuccessful,
      ...rest,
     processingException: processingError,
```

src/models/orchestrator/queues.models.ts
```diff
@@ -172,6 +174,81 @@ export interface QueueServiceModel {
...
+   * Marks the item `Successful` or `Failed` (with the failure details from
+   * {@link TransactionCompletionOptions}), and can persist output data
+   * alongside the result.
+   *
+   * Applies to items with an active transaction. Changing the outcome of an
+   * item that already reached a terminal status is rejected.
+   *
+   * @param itemId - Queue item ID of the transaction to complete
+   * @param folderId - Required folder ID
+   * @param isSuccessful - True when the item was processed successfully; false records a failure (provide `processingError` in options)
```
swati354: > @param isSuccessful - True when the item was processed successfully; false records a failure (provide `processingError` in options)

we need to pass this field? how do I know its value? And what happens if actually it had failed but I pass true?

agent_docs/conventions.md
```diff
@@ -80,7 +80,7 @@ The method attachment pattern:
 
 - **Not every service method gets bound.** Only bind methods that operate ON a specific entity after retrieval — state-changing operations (assign, cancel, complete, insert, update, delete) and contextual reads that need the entity's ID.
 - **NEVER** bind `getAll()`, `getById()`, `create()`, or cross-entity queries — these are service-level entry points. Binding them creates circular nonsense (an entity that retrieves itself).
-- **Read-only services don't bind at all** — Assets, Buckets, Processes, ChoiceSets, Cases, and ProcessIncidents have no `{Entity}Methods` interface. (Queues binds `getAllItems`/`insertItem` since queue-item support landed.)
+- **Read-only services don't bind at all** — Assets, Buckets, Processes, ChoiceSets, Cases, and ProcessIncidents have no `{Entity}Methods` interface. (Queues binds `getAllItems`/`insertItem`/`startTransaction`/`completeTransaction` since queue-item and transaction support landed.)
```
swati354: we could remove this line altogether since queues is not a read-only service now. 

> (Queues binds `getAllItems`/`insertItem`/`startTransaction`/`completeTransaction` since queue-item and transaction support landed.)

src/services/orchestrator/queues/queues.ts
shivendra6720: OperationResponse<void> is the SDK-wide contract for lifecycle operations  per conventions — keeping completeTransaction consistent

src/services/orchestrator/queues/queues.ts
shivendra6720: Yes - StartTransaction returns 204 with an empty body when no item is available, so data is genuinely absent on the wire; the union keeps the null-check honest instead of the type claiming a body always exists.

src/services/orchestrator/queues/queues.ts
shivendra6720: It's recursive, so it would also case-convert the user-defined keys inside outputData/analytics, which must be sent byte-exact

src/services/orchestrator/queues/queues.ts
shivendra6720: RobotIdentifier is defined by the API as the key of the robot that sent the
request - it identifies the caller, it doesn't route the item to another robot.A robot gets its item from its
own token without this field, and user/app identities always get null (verified
live). So there's no supported use for passing it — and if a robot-context
consumer appears later, adding it as an optional field is non-breaking.

src/services/orchestrator/queues/queues.ts
swati354: `OperationResponse` should return meaningful data,  here it returns 204 with no body so `OperationResponse<void>`  doesn't give us any value? We have [methods](https://uipath.github.io/uipath-typescript/api/interfaces/entity/#deleterecordbyid) using `promise<void>` as well so it wouldn't be new.

src/models/orchestrator/queues.models.ts
```diff
@@ -172,6 +176,92 @@ export interface QueueServiceModel {
...
+   *   folderKey: '<folderKey>',
+   *   processingError: {
+   *     reason: 'Vendor not found',
+   *     type: QueueExceptionType.BusinessException
+   *   }
+   * });
+   * ```
+   */
+  completeTransaction(
+    itemId: number,
```
swati354, line 344: I can start transaction via name or id but can complete it only via id?

src/models/orchestrator/queues.models.ts
```diff
@@ -172,6 +176,92 @@ export interface QueueServiceModel {
...
+   *
+   * Marks the item `Successful` or `Failed` (with the failure details from
+   * {@link QueueCompleteTransactionOptions}), and can persist output data
+   * alongside the result.
+   *
+   * Applies to items with an active transaction. Changing the outcome of an
+   * item that already reached a terminal status is rejected.
+   *
+   * @param itemId - Queue item ID of the transaction to complete
+   * @param isSuccessful - The caller's verdict on its own processing of the item: `true` marks it `Successful`, `false` marks it `Failed` (provide `processingError` in options). Orchestrator records the reported outcome as-is
```
swati354: again, I don't really like this param here. I get the comment tries to explain the reasoning but something like `queue.completeTransaction(item.id, true, { outputData })` doesn't really tell me what this boolean is all about. 
And secondly, if `isSuccessful` is false, is it required to pass `processingError`? Currently nothing in our code enforces that?
can we think of other ways of handling this? 
A union that enforces this condition 
```
type CompletionOutcome =
    | { outcome: 'success'; outputData?: Record<string, unknown> }
    | { outcome: 'failure'; processingError: QueueItemProcessingError };
```
 or perhaps an enum?
```
export enum QueueTransactionOutcome {
    Successful = 'Successful',
    Failed = 'Failed',
  }
```
`await queue.completeTransaction(item.id, QueueTransactionOutcome.Successful, { outputData });`

src/services/orchestrator/queues/queues.ts
```diff
@@ -199,4 +204,101 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
...
+
+  /**
+   * Resolves a queue ID to the queue's name.
+   */
+  private async _resolveQueueName(queueId: number, headers: Record<string, string>): Promise<string> {
+    const response = await this.get<{ Name: string }>(
+      QUEUE_ENDPOINTS.GET_BY_ID(queueId),
+      { headers, params: { '$select': 'Name' } }
+    );
+    return response.data.Name;
```
swati354: 1. `_resolveQueueName` -> `resolveQueueName`
2. Lets keep private methods at the bottom of the file.

src/models/orchestrator/queues.models.ts
shivendra6720: startTransaction selects a queue, and queues are addressable by name or id, so the ref supports either. completeTransaction targets one specific queue item, the one startTransaction handed back (item.id). Items have no name; the id is their only identifier, so there's nothing to resolve.

tests/unit/models/orchestrator/queues.test.ts
```diff
@@ -105,10 +108,65 @@ describe('Queue Models', () => {
...
+
+      it('should pass through null when no item is available', async () => {
+        const queue = createQueueWithMethods(createBasicQueue(), mockService);
+        vi.mocked(mockService.startTransactionByName).mockResolvedValue(null);
+
+        const result = await queue.startTransaction();
+
+        expect(result).toBeNull();
+      });
+    });
```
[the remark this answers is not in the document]
swati354: @shivendra6720

swati354 (approved):

deepeshrai-tech (approved):

PR #646, opened 2026-08-06
feat: onboard attachment upload and job attachment APIs

RakshithDattatraya:
Onboards the attachment APIs so a coded app can upload a file and attach it to a job.

Related: [APPS-35700](https://uipath.atlassian.net/browse/APPS-35700)

### What you can do now

```ts
// Upload a file the user picked
const attachment = await attachments.create(file.name, file);

// Or upload and attach it to a job in one call
const attachment = await attachments.create(file.name, file, {
  jobKey: <jobKey>,
  category: 'Invoice',
});

// List what's attached to a job
const links = await jobs.getAttachments(<jobKey>, <folderId>);

// Attach an existing attachment to another job
await jobs.linkAttachment(<attachmentId>, <jobKey>, <folderId>);
```

`getAttachments()` and `linkAttachment()` are also available on a job you've already fetched, so you don't have to pass the key and folder again:

```ts
const job = await jobs.getById(<jobKey>, <folderId>);
await job.getAttachments();
```

### Notes

- `attachments.create()` does the upload for you. Orchestrator hands back a short-lived write URL, the SDK PUTs the file to it, and that URL is deliberately left off the response — it's spent, and it's a credential.
- The `id` on the returned attachment is the handle to pass around, e.g. as the value of a `file` field in an action's output data.
- Orchestrator has no API to unlink an attachment from a job, so integration tests clean up by deleting the attachment itself.

### Testing

- Unit tests for both services, including that the write URL and OData metadata never reach the caller
- Integration tests against a live tenant, including downloading an uploaded file back and checking the bytes match
- Verified end to end against the built bundle: upload → attach → read back → clean up

[APPS-35700]: https://uipath.atlassian.net/browse/APPS-35700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/models/orchestrator/attachments.types.ts
+++ b/src/models/orchestrator/attachments.types.ts
@@ -52,3 +52,36 @@ export interface AttachmentResponse {
  * Options for getting an attachment by ID
  */
 export interface AttachmentGetByIdOptions extends BaseOptions {};
+
+/**
+ * Attachment returned when a new attachment is created.
+ *
+ * The upload URI is deliberately not exposed — the SDK uploads the content
+ * before returning, so the short-lived write credential serves no further
+ * purpose to the caller.
+ */
+export interface AttachmentCreateResponse extends Omit<AttachmentResponse, 'blobFileAccess'> {};
+
+/**
+ * Options for creating an attachment
+ */
+export interface AttachmentCreateOptions {
+  /**
+   * Key of a job to link the new attachment to. When set, the attachment is
+   * created and linked in a single call — no separate `jobs.linkAttachment()`
+   * is needed.
+   */
+  jobKey?: string;
+
+  /**
+   * Label that groups the attachment within the job. Only meaningful together
+   * with `jobKey`.
+   */
+  category?: string;
+
+  /**
+   * ID of the folder to create the attachment in. Defaults to the tenant's
+   * default folder context when omitted.
+   */
+  folderId?: number;
+}
--- a/src/models/orchestrator/jobs.models.ts
+++ b/src/models/orchestrator/jobs.models.ts
@@ -1,4 +1,4 @@
-import { JobGetAllOptions, JobGetByIdOptions, RawJobGetResponse, JobStopOptions, JobResumeOptions } from './jobs.types';
+import { JobGetAllOptions, JobGetByIdOptions, RawJobGetResponse, JobStopOptions, JobResumeOptions, JobAttachmentGetResponse, JobLinkAttachmentOptions } from './jobs.types';
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
 
 /** Combined response type for job data with bound methods. */
@@ -216,6 +216,69 @@ export interface JobServiceModel {
    * ```
    */
   restart(jobKey: string, folderId: number): Promise<JobGetResponse>;
+
+  /**
+   * Gets all attachments linked to a job.
+   *
+   * @param jobKey - The unique key (GUID) of the job whose attachments to retrieve
+   * @param folderId - The folder ID where the job resides
+   * @returns Promise resolving to an array of {@link JobAttachmentGetResponse} links, empty when the job has none
+   *
+   * @example
+   * ```typescript
+   * const attachments = await jobs.getAttachments(<jobKey>, <folderId>);
+   * ```
+   *
+   * @example
+   * ```typescript
+   * import { Attachments } from '@uipath/uipath-typescript/attachments';
+   *
+   * // Read the underlying file of each attachment linked to a job
+   * const attachmentsService = new Attachments(sdk);
+   * const links = await jobs.getAttachments(<jobKey>, <folderId>);
+   *
+   * for (const link of links) {
+   *   const attachment = await attachmentsService.getById(link.attachmentId);
+   *   console.log(link.category, attachment.blobFileAccess.uri);
+   * }
+   * ```
+   */
+  getAttachments(jobKey: string, folderId: number): Promise<JobAttachmentGetResponse[]>;
+
+  /**
+   * Links an existing attachment to a job.
+   *
+   * This does not upload a file — it associates an attachment that already
+   * exists in Orchestrator with a job.
+   *
+   * The returned link does not carry `attachmentName` or `creatorUserId` yet —
+   * read them back with {@link getAttachments}.
+   *
+   * @param attachmentId - The unique ID (GUID) of the attachment to link
+   * @param jobKey - The unique key (GUID) of the job to link the attachment to
+   * @param folderId - The folder ID where the job resides
+   * @param options - Optional settings for the link, such as its category
+   * @returns Promise resolving to the created {@link JobAttachmentGetResponse} link
+   *
+   * @example
+   * ```typescript
+   * const link = await jobs.linkAttachment(<attachmentId>, <jobKey>, <folderId>);
+   * ```
+   *
+   * @example
+   * ```typescript
+   * // Group the attachment under a category of your choosing
+   * const link = await jobs.linkAttachment(<attachmentId>, <jobKey>, <folderId>, {
+   *   category: 'Invoice',
+   * });
+   * ```
+   */
+  linkAttachment(
+    attachmentId: string,
+    jobKey: string,
+    folderId: number,
+    options?: JobLinkAttachmentOptions
+  ): Promise<JobAttachmentGetResponse>;
 }
 
 /**
@@ -278,6 +341,37 @@ export interface JobMethods {
    * @returns Promise resolving to the new {@link JobGetResponse} with full job details
    */
   restart(): Promise<JobGetResponse>;
+
+  /**
+   * Gets all attachments linked to this job.
+   *
+   * @returns Promise resolving to an array of {@link JobAttachmentGetResponse} links, empty when the job has none
+   *
+   * @example
+   * ```typescript
+   * const job = await jobs.getById(<jobKey>, <folderId>);
+   * const attachments = await job.getAttachments();
+   * ```
+   */
+  getAttachments(): Promise<JobAttachmentGetResponse[]>;
+
+  /**
+   * Links an existing attachment to this job.
+   *
+   * @param attachmentId - The unique ID (GUID) of the attachment to link
+   * @param options - Optional settings for the link, such as its category
+   * @returns Promise resolving to the created {@link JobAttachmentGetResponse} link
+   *
+   * @example
+   * ```typescript
+   * const job = await jobs.getById(<jobKey>, <folderId>);
+   * const link = await job.linkAttachment(<attachmentId>, { category: 'Invoice' });
+   * ```
+   */
+  linkAttachment(
+    attachmentId: string,
+    options?: JobLinkAttachmentOptions
+  ): Promise<JobAttachmentGetResponse>;
 }
 
 /**
@@ -308,6 +402,19 @@ function createJobMethods(jobData: RawJobGetResponse, service: JobServiceModel):
       if (!jobData.key) throw new Error('Job key is undefined');
       if (!jobData.folderId) throw new Error('Job folderId is undefined');
       return service.restart(jobData.key, jobData.folderId);
+    },
+    async getAttachments(): Promise<JobAttachmentGetResponse[]> {
+      if (!jobData.key) throw new Error('Job key is undefined');
+      if (!jobData.folderId) throw new Error('Job folderId is undefined');
+      return service.getAttachments(jobData.key, jobData.folderId);
+    },
+    async linkAttachment(
+      attachmentId: string,
+      options?: JobLinkAttachmentOptions
+    ): Promise<JobAttachmentGetResponse> {
+      if (!jobData.key) throw new Error('Job key is undefined');
+      if (!jobData.folderId) throw new Error('Job folderId is undefined');
+      return service.linkAttachment(attachmentId, jobData.key, jobData.folderId, options);
     },
   };
 }
--- a/src/models/orchestrator/jobs.types.ts
+++ b/src/models/orchestrator/jobs.types.ts
@@ -194,3 +194,69 @@ export interface JobStopOptions {
   strategy?: StopStrategy;
 }
 
+
+/**
+ * Interface for an attachment linked to a job
+ */
+export interface JobAttachmentGetResponse {
+  /**
+   * Unique identifier of the link between the job and the attachment
+   */
+  id: string;
+
+  /**
+   * Identifier of the linked attachment. Pass this to `attachments.getById()`
+   * to read the attachment's metadata and download URI.
+   */
+  attachmentId: string;
+
+  /**
+   * Key of the job the attachment is linked to
+   */
+  jobKey: string;
+
+  /**
+   * Caller-defined label that groups attachments within a job. Attachments
+   * uploaded by a running automation carry the category `JobAttachment`.
+   * Null when the link was created without a category.
+   */
+  category: string | null;
+
+  /**
+   * File name of the linked attachment. Null on the response returned when the
+   * link is first created — read it back with {@link JobServiceModel.getAttachments}.
+   */
+  attachmentName: string | null;
+
+  /**
+   * When the link was last modified
+   */
+  lastModifiedTime: string | null;
+
+  /**
+   * ID of the user who last modified the link
+   */
+  lastModifierUserId: number | null;
+
+  /**
+   * When the link was created
+   */
+  createdTime: string;
+
+  /**
+   * ID of the user who created the link
+   */
+  creatorUserId: number | null;
+}
+
+/**
+ * Options for linking an attachment to a job
+ */
+export interface JobLinkAttachmentOptions {
+  /**
+   * Label that groups the attachment within the job. Free-form — any string the
+   * calling application uses to distinguish attachments (e.g. `Input`,
+   * `Output`, `Invoice`). Defaults to null when omitted.
+   */
+  category?: string;
+}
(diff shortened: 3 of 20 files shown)


src/services/orchestrator/attachments/attachments.ts
```diff
@@ -1,12 +1,27 @@
...
+  transformRequest,
+  arrayDictionaryToRecord,
+} from '../../../utils/transform';
 import { ORCHESTRATOR_ATTACHMENT_ENDPOINTS } from '../../../utils/constants/endpoints';
 import { ODATA_PREFIX } from '../../../utils/constants/common';
+import { createHeaders } from '../../../utils/http/headers';
+import { FOLDER_ID } from '../../../utils/constants/headers';
 import { track } from '../../../core/telemetry';
 import { AttachmentsMap } from '../../../models/orchestrator/attachments.constants';
 import { BaseService } from '../../..//services/base';
```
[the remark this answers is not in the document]
RakshithDattatraya: Fixed — the extra slash is gone.

For the record this line was pre-existing on `main`, not introduced by this PR, so it was technically outside the review scope. It's a zero-risk one-character fix in a file this PR already touches, so I took it rather than leave it.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+    if (!blobFileAccess?.uri) {
+      throw new ServerError({ message: 'Attachment upload URI missing from the create response' });
+    }
+
+    // Storage returns its required headers (e.g. x-ms-blob-type) as parallel
+    // key/value arrays; sending them unconverted drops them and the PUT is rejected.
+    const rawHeaders = blobFileAccess.headers;
+    const requestHeaders: Record<string, string> =
+      rawHeaders && 'keys' in rawHeaders && 'values' in rawHeaders
+        ? arrayDictionaryToRecord(rawHeaders as unknown as { keys: string[]; values: string[] })
```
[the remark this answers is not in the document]
RakshithDattatraya: Not taking this one, for two reasons.

**The suggested code doesn't compile.** I tried it:

```
error TS2352: Conversion of type 'ResponseDictionary' to type '{ keys: string[]; values: string[]; }'
may be a mistake because neither type sufficiently overlaps with the other.
If this was intentional, convert the expression to 'unknown' first.
```

The compiler explicitly asks for `as unknown as` here — which is what the code already does. The suggestion only works in combination with the type widening, so the two halves aren't independent.

**The type widening is a breaking public API change.** `ResponseDictionary` and `BucketGetUriResponse` are both exported. Widening `headers` to `Record<string, string> | { keys: string[]; values: string[] }` means every consumer doing `headers['x-ms-blob-type']` starts failing to compile — including external SDK users, plus three internal call sites (`jobs.ts:263`, `conversations.ts:194`, and this one).

The underlying observation is correct: `headers: Record<string, string>` is simply wrong, because Azure Storage returns parallel arrays. But fixing that properly is a deliberate public-type decision that deserves its own PR, not a drive-by in a feature change. Until then this matches the existing precedent at `buckets.ts:394`, which performs the identical cast for the identical reason, and the cast carries a comment explaining why.

tests/integration/shared/orchestrator/attachments.integration.test.ts
```diff
@@ -93,5 +94,100 @@ describe.each(modes)(
         await expect(attachments.getById('')).rejects.toThrow('id is required for getById');
```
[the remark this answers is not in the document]
RakshithDattatraya: Not taking this one — that `it` block is pre-existing on `main` (`attachments.integration.test.ts:93`), not added by this PR.

The convention cited is right, and the test probably should move. But removing an unrelated existing test as part of a feature PR is churn that muddies the diff and isn't mine to delete here. Worth a separate cleanup.

src/services/orchestrator/attachments/attachments.ts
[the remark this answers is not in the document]
RakshithDattatraya: Fixed — the extra slash is gone.

For the record this line was pre-existing on `main`, not introduced by this PR, so it was technically outside the review scope. It's a zero-risk one-character fix in a file this PR already touches, so I took it rather than leave it.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+        : { ...rawHeaders };
+
+    if (blobFileAccess.requiresAuth) {
+      requestHeaders['Authorization'] = `Bearer ${await this.getValidAuthToken()}`;
+    }
+
+    const response = await fetch(blobFileAccess.uri, {
+      method: 'PUT',
```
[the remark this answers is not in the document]
RakshithDattatraya: Not taking this one — the cited precedent doesn't exist, and the convention doesn't apply to this call.

`grep -rn "TRACEPARENT" src/` returns exactly one file: `core/http/api-client.ts`. No service adds tracing headers to a raw `fetch()`, including the citation download referenced here.

More importantly, this `PUT` targets **Azure Blob Storage** (`*.blob.core.windows.net`) via a presigned SAS URL — not a UiPath service. The convention exists so calls to UiPath endpoints stay visible to UiPath's tracing infrastructure; Azure never forwards these headers there, so they'd add nothing and would leak internal correlation IDs to a third party.

The two existing blob-storage `fetch()` call sites both deliberately omit them: `buckets.ts` `_uploadToUri()` and `jobs.ts` `downloadOutputFile()`. This follows suit.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+      throw new ServerError({ message: 'Attachment upload URI missing from the create response' });
+    }
+
+    // Storage returns its required headers (e.g. x-ms-blob-type) as parallel
+    // key/value arrays; sending them unconverted drops them and the PUT is rejected.
+    const rawHeaders = blobFileAccess.headers;
+    const requestHeaders: Record<string, string> =
+      rawHeaders && 'keys' in rawHeaders && 'values' in rawHeaders
+        ? arrayDictionaryToRecord(rawHeaders as unknown as { keys: string[]; values: string[] })
+        : { ...rawHeaders };
```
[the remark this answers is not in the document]
RakshithDattatraya: Not taking this one, for two reasons.

**The suggested code doesn't compile.** I tried it:

```
error TS2352: Conversion of type 'ResponseDictionary' to type '{ keys: string[]; values: string[]; }'
may be a mistake because neither type sufficiently overlaps with the other.
If this was intentional, convert the expression to 'unknown' first.
```

The compiler explicitly asks for `as unknown as` here — which is what the code already does. The suggestion only works in combination with the type widening, so the two halves aren't independent.

**The type widening is a breaking public API change.** `ResponseDictionary` and `BucketGetUriResponse` are both exported. Widening `headers` to `Record<string, string> | { keys: string[]; values: string[] }` means every consumer doing `headers['x-ms-blob-type']` starts failing to compile — including external SDK users, plus three internal call sites (`jobs.ts:263`, `conversations.ts:194`, and this one).

The underlying observation is correct: `headers: Record<string, string>` is simply wrong, because Azure Storage returns parallel arrays. But fixing that properly is a deliberate public-type decision that deserves its own PR, not a drive-by in a feature change. Until then this matches the existing precedent at `buckets.ts:394`, which performs the identical cast for the identical reason, and the cast carries a comment explaining why.

tests/unit/models/orchestrator/jobs.test.ts
```diff
@@ -187,5 +189,78 @@ describe('Job Models', () => {
...
+
+      it('should throw when the job key is undefined', async () => {
+        const job = createJobWithMethods(createBasicJob({ key: undefined as any }), mockService);
+
+        await expect(
+          job.linkAttachment(JOB_TEST_CONSTANTS.ATTACHMENT_ID)
+        ).rejects.toThrow('Job key is undefined');
+        expect(mockService.linkAttachment).not.toHaveBeenCalled();
+      });
+    });
```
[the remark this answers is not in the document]
RakshithDattatraya: Good catch — fixed. Added the `folderId` guard test for `job.linkAttachment()`, matching the other bound methods.

The guard did exist in `createJobMethods`, but nothing exercised it, so a regression there would have gone unnoticed. Verified it passes and that the sibling `getAttachments` case still does.

src/models/orchestrator/attachments.types.ts
```diff
@@ -52,3 +52,36 @@ export interface AttachmentResponse {
...
 export interface AttachmentGetByIdOptions extends BaseOptions {};
+
+/**
+ * Attachment returned when a new attachment is created.
+ *
+ * The upload URI is deliberately not exposed — the SDK uploads the content
+ * before returning, so the short-lived write credential serves no further
+ * purpose to the caller.
+ */
+export interface AttachmentCreateResponse extends Omit<AttachmentResponse, 'blobFileAccess'> {};
```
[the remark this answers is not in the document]
RakshithDattatraya: Keeping the `interface` form — I tested both against TypeDoc and the claim that it "expands utility types correctly for both `type` and `interface` forms" doesn't hold.

**`interface extends Omit<...>`** produces a full property table:

```
## Properties
| Property | Type | Description |
| `attachmentCategory?` | `string` | Optional category for the attachment... |
| `createdTime?` | `string` | When the attachment was created |
| `id` | `string` | UUID of the attachment |
| ...8 properties total, each with its description
```

**`type = Omit<...>`** produces no property table at all — the entire generated page is:

```
> **AttachmentCreateResponse** = `Omit`<[`AttachmentResponse`](...), `"blobFileAccess"`>
```

A user reading the docs for `attachments.create()` would have to click through to `AttachmentResponse` and mentally subtract a field to learn what they get back.

This is precisely the TypeDoc breakage the convention's exception exists to prevent — the guidance says `interface extends` is required where a `type` alias "breaks TypeDoc", and an alias that renders zero properties qualifies. Reproducible with `npx typedoc` on either form.

deepeshrai-tech (commented):
Reviewed

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+    if (!content) {
+      throw new ValidationError({ message: 'content is required for create' });
+    }
+
+    // Passing jobKey here links the attachment to the job as part of creation,
+    // so no follow-up call is needed.
+    const response = await this.post<Record<string, unknown>>(
+      ORCHESTRATOR_ATTACHMENT_ENDPOINTS.CREATE,
+      camelToPascalCaseKeys(
+        transformRequest({ name, jobKey: options?.jobKey, attachmentCategory: options?.category }, AttachmentsMap)
```
deepeshrai-tech: the `category → attachmentCategory` rename happens manually here, and then this object gets run through `transformRequest(..., AttachmentsMap)`. If `AttachmentsMap` also carries that rename, one of them is dead code.Might be worth building the request object directly with the API-facing names and skipping the map for these fields — just so there's one source of truth.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+      ),
+      { headers: createHeaders({ [FOLDER_ID]: options?.folderId }) }
```
deepeshrai-tech, line 77: when `options?.folderId` is `undefined`, does `createHeaders` drop the entry, or does the header go on the wire as `X-UIPATH-OrganizationUnitId: undefined`? The JSDoc on the option says it "defaults to the tenant's default folder" when omitted, which only holds if the header is truly not sent?

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -36,4 +51,81 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
...
+
+    const response = await fetch(blobFileAccess.uri, {
+      method: 'PUT',
+      body: content,
+      headers: createHeaders(requestHeaders),
+    });
+
+    if (!response.ok) {
+      // Reported from the response body — the URI itself is a credential and is never echoed.
+      const errorInfo = await errorResponseParser.parse(response);
```
deepeshrai-tech, line 129: The upload-failure unit test uses a text mock and asserts `AuthorizationError`, so the parser clearly has a status-based fallback. But a real 403 from Azure Blob comes back as an XML body which the parser might not know what to do with. Not blocking but check once

src/services/orchestrator/jobs/jobs.ts
```diff
@@ -187,6 +187,63 @@ export class JobService extends FolderScopedService implements JobServiceModel {
...
+    const response = await this.get<Record<string, unknown>[]>(
+      JOB_ENDPOINTS.ATTACHMENTS.GET_BY_JOB_KEY,
+      {
+        params: { jobKey },
+        headers: createHeaders({ [FOLDER_ID]: folderId }),
+      }
+    );
+
+    // The endpoint answers in camelCase already, so only the semantic renames apply.
+    return transformData(response.data, JobAttachmentMap) as unknown as JobAttachmentGetResponse[];
```
deepeshrai-tech: The `as unknown as JobAttachmentGetResponse[]` force-cast is a place where TS will happily typecheck a lie if the API shape ever drifts. Unit tests cover the runtime shape thoroughly so the risk is low today, but a `RawJobAttachmentResponse` intermediate type + `satisfies` on the transform output would let the compiler prove the mapping?

src/services/orchestrator/attachments/attachments.ts
RakshithDattatraya: Fixed this

src/services/orchestrator/attachments/attachments.ts
RakshithDattatraya: `createHeaders` already omits undefined values, so the header isn't sent. I also added a test to cover this behavior since this is the only call site that passes an optional folderId.

src/services/orchestrator/jobs/jobs.ts
RakshithDattatraya: I agree it's not ideal, but this is one of the cases our conventions allow. transformData() doesn't preserve the renamed type, so without changing that utility the cast is still required. Schema drift is covered by the integration test against the real API.

src/services/orchestrator/attachments/attachments.ts
RakshithDattatraya: I verified this with a real Azure XML error response. The request fails gracefully, but the XML error details are lost because response.json() consumes the body before the fallback calls response.text(). This is an existing issue in core/errors/parser.ts. Fixed it now. Please check

src/models/orchestrator/jobs.constants.ts
```diff
@@ -15,3 +15,12 @@ export const JobMap: { [key: string]: string } = {
   processType: 'packageType',
   release: 'process',
 };
+
+/**
+ * Maps fields for attachments linked to a job to ensure consistent naming
+ * Semantic renames only — the endpoint already answers in camelCase
+ */
+export const JobAttachmentMap: { [key: string]: string } = {
```
ninja-shreyash: Why do we need the separate map when its the same Jobs service?

src/models/orchestrator/jobs.models.ts
```diff
@@ -216,6 +216,78 @@ export interface JobServiceModel {
...
    */
   restart(jobKey: string, folderId: number): Promise<JobGetResponse>;
+
+  /**
+   * Gets all attachments linked to a job.
+   *
+   * Returns one entry per link, each carrying the attachment's name, its
+   * category, and the `attachmentId` needed to read the file itself. To get a
+   * job key and its folder, first list jobs with {@link getAll} — each job
+   * exposes `key` and `folderId`.
```
ninja-shreyash:
```suggestion
   * Returns one entry per link, each carrying the attachment's name, its
   * category, and the `attachmentId` needed to read the file itself.
```

This detail is for all the methods that need a job key and folderId. No need to mention at this method level

src/models/orchestrator/jobs.models.ts
```diff
@@ -216,6 +216,78 @@ export interface JobServiceModel {
...
+   *   console.log(link.category, attachment.blobFileAccess.uri);
+   * }
+   * ```
+   */
+  getAttachments(jobKey: string, folderId: number): Promise<JobAttachmentGetResponse[]>;
+
+  /**
+   * Links an existing attachment to a job.
+   *
+   * This does not upload a file — it associates an attachment that already
```
ninja-shreyash, line 251: no need explain params here when they are defined below in @params blocks.. so this info is redundant at top-level

src/models/orchestrator/jobs.models.ts
ninja-shreyash: also there is a separate @returns section so no need to mention what method returns in top-level desc

src/models/orchestrator/jobs.types.ts
```diff
@@ -194,3 +194,69 @@ export interface JobStopOptions {
...
+  /**
+   * Caller-defined label that groups attachments within a job. Attachments
+   * uploaded by a running automation carry the category `JobAttachment`.
+   * Null when the link was created without a category.
+   */
+  category: string | null;
+
+  /**
+   * File name of the linked attachment. Null on the response returned when the
+   * link is first created — read it back with `getAttachments()` to get the name.
```
ninja-shreyash: use @link for getAttachments

deepeshrai-tech (approved):

ninja-shreyash (approved):

PR #653, opened 2026-08-10
samples: add Coded Functions sample app

amrit-agarwal-1:
Adds `samples/functions-app`, a React Coded App exercising the **Functions** service, together with the two coded functions it calls, so the sample deploys end to end rather than pointing at something the reader has to invent.

## The sample

A small shop. You browse a catalogue, add items to a basket, and enter a discount code at checkout.

The discount code is the point. **The valid codes live in an Orchestrator Secret asset**, which the browser cannot read even with the signed-in user's own token. So checking a code has to happen inside a function: if validation ran in the page, the page would need the list, and anyone could open the source and use codes they were never given.

The justification is a platform fact rather than a story:

- A Secret is **omitted entirely** from the assets listing. A query returns zero rows, not a row with a blank value.
- Its value is returned only by `GetRobotAssetByNameForRobotKey`, which requires a robot key. That key reaches your code as `ctx.robot.key`, and only a deployed job has one.

For a Secret the value arrives in **`SecretValue`**. `StringValue`, the field that works for Text assets, is an empty string, so reading that instead looks exactly like a broken feature. Worth knowing before anyone debugs this from scratch.

## What it demonstrates

| function | route | purpose |
|---|---|---|
| `list-products` | `GET /products` | the catalogue, priced server-side |
| `quote` | `POST /quote` | prices the basket and validates the promo code |

SDK surface: `Functions.getAll()`, `Functions.invoke<TInput, TOutput>()`, and `new UiPath()` with zero config relying on platform-injected meta tags.

`coded-functions/lib/contract.ts` is the single source of truth for every I/O type. The functions import it directly; the app imports it with `import type`, so both halves share one definition with nothing bundled across the boundary. 

Sample app link: https://popoc.alpha.uipath.host/promo-shop


Raina451:
please fix [PR Checks / check-samples (pull_request)](https://github.com/UiPath/uipath-typescript/actions/runs/31591190697/job/94096291061?pr=653) faiures

samples/functions-app/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc --noEmit && vite build",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@uipath/apollo-wind": "^2.32.2",
+    "@uipath/uipath-typescript": "^1.6.1",
+    "lucide-react": "^1.31.0",
```
Raina451, line 15: what are we using lucide-react for?

samples/functions-app/README.md
```diff
@@ -0,0 +1,318 @@
...
+uip tools install @uipath/codedapp-tool       # uip codedapp ...
+uip tools install @uipath/orchestrator-tool   # uip or ...
+
+uip login
+```
+
+For a non-production environment, point the login at that authority:
+
+```bash
+uip login --authority https://alpha.uipath.com
```
Raina451: dont use alpha anywhere.

samples/functions-app/README.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+## 2. Collect the values you will need
+
+You need three things, and two of them are easy to confuse.
+
+```bash
+uip or folders list --output table
+```
```
Raina451, line 89: they also need to install these tools, lets add those commands also. like:
`uip tools install @uipath/orchestrator-tool`

samples/functions-app/README.md
```diff
@@ -0,0 +1,318 @@
...
+`GET {baseUrl}/{org}/{tenant}/orchestrator_/odata/Folders` and take the `Id` field.
+
+**The base URL must be the API subdomain, not the portal domain.** The portal domain sends no CORS
+headers and the browser will block every call.
+
+| environment | base URL |
+|---|---|
+| Production | `https://api.uipath.com` |
+| Alpha | `https://alpha.api.uipath.com` |
+| Staging | `https://staging.api.uipath.com` |
```
Raina451: no where mention alpha / staging. everywhere in this pr, use prod

Raina451 (changes requested):

samples/functions-app/coded-functions/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "promo-shop-fn",
```
Raina451: also the name of app, instead of promo shop, lets use coded function sample or something

samples/functions-app/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "promo-shop",
```
Raina451: here too

samples/functions-app/README.md
amrit-agarwal-1: its added above in line 76

samples/functions-app/package.json
amrit-agarwal-1: Its being used for the icons that we show in the catalogue to make the UI more appealing. Apollo wind doesn't have those icons support.

samples/functions-app/README.md
amrit-agarwal-1: Removed from all places now.

samples/functions-app/README.md
amrit-agarwal-1: fixed now

samples/functions-app/coded-functions/package.json
amrit-agarwal-1: renamed at all places

samples/functions-app/.gitignore
```diff
@@ -0,0 +1,25 @@
+node_modules
+dist
+dist-ssr
+*.local
+.uipath
+*.nupkg
+
+# Only the app's own OAuth config — copy uipath.json.example to uipath.json.
+# Anchored, so it does not swallow coded-functions/uipath.json (the functions map).
+/uipath.json
```
Raina451: this shouldnt be added to giignore,
uipath.json doesnt contain any secrets

Raina451 (approved):

deepeshrai-tech (approved):

samples/functions-app/.gitignore
amrit-agarwal-1: fixed

PR #663, opened 2026-08-13
feat(platform): add Users service for organization user management [APPS-35699]

Sarath1018:
## Summary

First PR in the platform RBAC stack (**Users → Groups → Directory → Roles**). Adds `PlatformUserService` (exported as `Users`) to the existing `/platform` subpath.

| Method | Endpoint | Notes |
|---|---|---|
| `getAll(organizationId, options?)` | `GET /identity_/api/User/users/{organizationId}` | Paged (`pageSize`/`cursor`/`jumpToPage`) or fetch-all; `searchTerm`, `sortBy`, `sortOrder` filters |
| `getById(userId)` | `GET /identity_/api/User/{userId}` | Profile + `groupIds` memberships; bound `update()` attached |
| `updateById(userId, options)` | `PUT /identity_/api/User/{userId}` | Partial update; incremental group membership via `groupIdsToAdd`/`groupIdsToRemove` |

## Design decisions (validated against the live API)

- **Wire is camelCase** → no case transform; semantic renames only (`creationTime→createdTime`, `lastModificationTime→lastModifiedTime`, `groupIDs→groupIds`)
- **Numeric enum codes**: live API returns `type`/`category` as numbers despite the swagger declaring string enums — mapped via `PlatformUserTypeMap`/`PlatformUserCategoryMap`
- **Dropped internal fields**: `legacyId`, `bypassBasicAuthRestriction`
- **Fetch-all loop**: the API always pages (default 10, max `top=1000`), so `getAll` without pagination options iterates pages — a single request would silently truncate
- **Update semantics**: PUT is partial (null/omitted = no change) — verified live that omitted fields are not clobbered

## Bug fix included

`BaseService` pagination now honors an explicit `countParam: undefined` (via `'countParam' in paginationParams`) instead of `||`-falling back to `$count: true`. This implements the documented `COUNT_PARAM` convention; previously every non-OData service (Governance, Agents, SLA, Entities, ChoiceSets) had `$count: true` silently injected. Data Fabric integration suite re-run live to confirm no regression (82 passing; only pre-existing missing-env failures remain).

## Testing

- Unit: 2349 passing (24 new: transform completeness, enum mapping, pagination paths, `excludeFromPrefix` on both paths, error paths)
- Model tests: bound `update()` delegation + missing-ID error
- Integration (live, 9/9): fetch-all, search, sort, cursor pagination, transform validation, `updateById` round-trip with snapshot+restore on a dedicated mutable test user (`IDENTITY_MUTABLE_TEST_USER_ID`, new env var)
- E2E: scaffolded React app built against the packed tarball + Node runtime validation through dist bundles against the live API

## Docs

- `docs/oauth-scopes.md`: `PM.User` / `PM.User.Read` / `PM.User.Write`
- `docs/pagination.md`: quick-reference row
- `mkdocs.yml`: nav entry for `PlatformUserServiceModel`

## Stack

- **PR 1/4 (this)**: Users → `main`
- PR 2/4: Groups → `feat/platform-users`
- PR 3/4: Directory → `feat/platform-groups`
- PR 4/4: Roles (pap_) → `feat/platform-directory`
- Cloudflare whitelist PR in apps-dev-tools (separate repo)


src/services/platform/users.ts
```diff
@@ -0,0 +1,172 @@
...
+      const response = await this.get<RawPlatformUserListResponse>(
+        IDENTITY_USER_ENDPOINTS.GET_ALL(organizationId),
+        { params: createParams({ searchTerm, sortBy, sortOrder, top: IDENTITY_MAX_PAGE_SIZE, skip }) }
+      );
+      const { results, totalCount: reportedTotal } = response.data;
+      totalCount = reportedTotal;
+      items.push(...results.map(user => this.toUser(user)));
+
+      if (results.length === 0 || items.length >= totalCount) {
+        break;
```
amrit-agarwal-1, line 154: This loop stops when items.length >= totalCount and advances skip by the returned count. No sort is sent, so if two pages overlap even once, the duplicate pads the count, the loop exits early, and a real user never gets fetched. I tried it on the branch, pages [a,b,c] then [c,d,e] with totalCount: 6 gives you 6 items but only 5 distinct users, and nothing surfaces it.

fetchAllEntries in data-fabric/directory.tsdoes the same job but advances by top and treats a short page as terminal, which I think is the safer reading for a record offset. Can we align with that, plus a stable sortBy: Id and a dedupe by id?

src/services/base.ts
```diff
@@ -232,7 +232,11 @@ export class BaseService {
         limitedPageSize = getLimitedPageSize(params.pageSize);
         const pageSizeParam = paginationParams?.pageSizeParam || ODATA_OFFSET_PARAMS.PAGE_SIZE_PARAM;
         const offsetParam = paginationParams?.offsetParam || ODATA_OFFSET_PARAMS.OFFSET_PARAM;
-        const countParam = paginationParams?.countParam || ODATA_OFFSET_PARAMS.COUNT_PARAM;
+        // Honor an explicit `countParam: undefined` (non-OData API, no count param);
+        // fall back to the OData default only when the key is absent entirely.
+        const countParam = paginationParams && 'countParam' in paginationParams
```
amrit-agarwal-1, line 237: Good catch on ||, but nothing in tests/unit/ mentions $count at all, the only coverage is indirect via the Users suite, so someone swapping the in check back to ?? gets a green run. Should we add these cases in base.test.ts: params absent → $count=true, explicit '$count' → sent, explicit undefined → omitted.

Sarath1018: done

src/services/platform/users.ts
Sarath1018: done

src/services/platform/users.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Service for managing an organization's users.
+ *
+ * Users are organization-scoped accounts. Together with groups they form the basis of
+ * access management: put users in groups, then grant roles to the groups. Group
+ * membership is edited through {@link PlatformUserService.updateById} via
+ * `groupIdsToAdd` / `groupIdsToRemove`.
+ */
+export class PlatformUserService extends BaseService implements PlatformUserServiceModel {
+  @track('PlatformUsers.GetAll')
```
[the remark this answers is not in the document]
Sarath1018: The `@track` convention in this codebase is the plural public alias, not class-name-minus-`Service` — e.g. `TaskService` → `Tasks.*`, `EntityService` → `Entities.*`, `NotificationService` → `Notifications.*`. `ConversationalAgent` is singular only because that service's public alias is singular. `PlatformUserService` is exported as `Users`, and the sibling PRs in this stack (#664–#666) already use `PlatformGroups.*` / `PlatformDirectory.*` / `PlatformRoles.*`, as does the merged `Platform.*` settings service. Keeping `PlatformUsers.*` is the consistent choice — renaming to singular would create exactly the dashboard-query inconsistency this comment is concerned about.

amrit-agarwal-1 (approved):

src/models/platform/users.constants.ts
```diff
@@ -0,0 +1,46 @@
...
+  creationTime: 'createdTime',
+  lastModificationTime: 'lastModifiedTime',
+  groupIDs: 'groupIds',
+} as const;
+
+/**
+ * Outbound renames for `users.updateById()` — the API expects `groupIDs` casing.
+ * Used with `transformRequest()`, which reverses the map (SDK name → wire name).
+ */
+export const PlatformUserUpdateMap = {
```
ninja-shreyash: Can we add these to PlatformUserMap and use the same ? thats how we do for other services - use single map for both transforming request and response.

src/models/platform/users.models.ts
```diff
@@ -0,0 +1,174 @@
...
+import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
+
+/**
+ * A user account with entity methods attached.
+ */
+export type PlatformUserGetResponse = RawPlatformUserGetResponse & PlatformUserMethods;
+
+/**
+ * Public surface of the platform Users service. JSDoc on this interface drives
+ * the generated API reference documentation.
```
ninja-shreyash:
```suggestion
 * Public surface of the platform Users service.
```

src/models/platform/users.internal-types.ts
```diff
@@ -0,0 +1,53 @@
...
+  code: string;
+  description: string;
+}
+
+/**
+ * Wire shape of the user update result.
+ */
+export interface RawPlatformUserUpdateResult {
+  succeeded: boolean;
+  errors: RawPlatformUserUpdateError[];
```
ninja-shreyash: errors is typed as required here, but in updateById, we read it as response.data.errors ?? [] (suggesting that it can be null/undefined). Can you confirm?

ninja-shreyash (commented):
Image 

Can we place the Platform Users service within the Platform in nav?

Also check that both Platform service and Platform Users service are missing "Usage" section (every other service has it). We should add that as well.

src/models/platform/users.types.ts
```diff
@@ -0,0 +1,154 @@
...
+  /** Display name shown in the UI. */
+  displayName: string | null;
+  /** When the account was created. */
+  createdTime: string;
+  /** When the account was last modified. */
+  lastModifiedTime: string | null;
+  /** When the user last signed in. */
+  lastLoginTime: string | null;
+  /** GUIDs of the groups the user belongs to. */
+  groupIds: string[];
```
ninja-shreyash, line 90: is it possible for user to not be part of any group? what happens in that case?

src/services/platform/index.ts
```diff
@@ -21,17 +21,15 @@
  *   organizationId: '<organizationId>',
  * });
  *
- * await platform.updateUserSettings(
```
ninja-shreyash, line 24: why remove this example?

src/models/platform/users.constants.ts
Sarath1018: Done in 5ef179bc1 — merged into a single `PlatformUserMap` used by both `transformData()` (response) and `transformRequest()` (update body), `PlatformUserUpdateMap` removed.

src/models/platform/users.models.ts
Sarath1018: Applied in 5ef179bc1.

src/models/platform/users.internal-types.ts
Sarath1018: Confirmed against the Identity spec: `errors` is `nullable: true` on the wire. The raw type now reflects that (`RawPlatformUserUpdateError[] | null`) and the `?? []` read stays, with a unit test covering the null case (5ef179bc1).

src/models/platform/users.types.ts
Sarath1018: Good catch — per the spec `groupIDs` is `nullable: true`, so a user can have none. The SDK now normalizes null to `[]` in the transform so callers always get an array; raw type updated and unit test added (5ef179bc1).

src/services/platform/index.ts
Sarath1018: Restored in 5ef179bc1 — the module example now shows both `updateUserSettings` and the new `users.getAll`.

ninja-shreyash:
image
Should we call it platform only? because the jsdoc says platform service

ninja-shreyash (commented):
Rest LGTM

ninja-shreyash (approved):

src/models/platform/users.models.ts
```diff
@@ -0,0 +1,184 @@
...
+   * @example Basic usage
+   * ```typescript
+   * import { UiPath } from '@uipath/uipath-typescript/core';
+   * import { Users } from '@uipath/uipath-typescript/platform';
+   *
+   * const sdk = new UiPath(config);
+   * await sdk.initialize();
+   *
+   * const users = new Users(sdk);
+   * const allUsers = await users.getAll('<organizationId>');
```
Raina451, line 56: lets not use orgid in these methods also as discussed on other thread.

src/models/platform/users.models.ts
```diff
@@ -0,0 +1,184 @@
...
+   * membership check against a known group needs no extra call.
+   *
+   * @param organizationId - Organization (account) GUID to list users from
+   * @param options - Search, sorting, and pagination options
+   * @returns All users when no pagination options are given, one page otherwise, as {@link PlatformUserGetResponse} items
+   *
+   * @example Basic usage
+   * ```typescript
+   * import { UiPath } from '@uipath/uipath-typescript/core';
+   * import { Users } from '@uipath/uipath-typescript/platform';
```
Raina451, line 50: question: why /platform? and why not /users?
i know earlier for usersettings etc we added under /platform, but what all other entities are going to come under platform?

Raina451: if somebody just want user service, they would rather import only user related method /user path, platform name does not sound intutive to me.

Raina451: @vnaren23 @ninja-shreyash thoughts?

Raina451 (changes requested):

src/models/platform/users.models.ts
vnaren23: Makes sense
> if somebody just want user service, they would rather import only user related method /user path

tests/integration/shared/platform/users.integration.test.ts
```diff
@@ -0,0 +1,175 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { getServices, getTestConfig, setupUnifiedTests, InitMode } from '../../config/unified-setup';
+import { Users } from '../../../../src/services/platform/users';
+import {
+  PlatformUserSortField,
+  PlatformUserSortOrder,
+  PlatformUserType,
+  PlatformUserCategory,
+} from '../../../../src/models/platform';
+import type { PlatformUserGetResponse } from '../../../../src/models/platform';
```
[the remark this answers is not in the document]
Sarath1018: Done in 213fe3e02 — collapsed onto a single import from `src/services/platform/users` (plus the type import), matching the unit test.

docs/authentication.md
```diff
@@ -199,6 +199,20 @@ const tasks = new Tasks(sdk);
...
+Some platform services (for example Users) address the organization by its GUID rather than by `orgName`. The SDK resolves that GUID for you — from the access token when it is a JWT, or by looking it up from `orgName` for personal access tokens — and caches it for the lifetime of the `UiPath` instance. Set `organizationId` to skip the lookup or to override the result:
+
+```typescript
+const sdk = new UiPath({
+  baseUrl: 'https://api.uipath.com',
+  orgName: 'your-organization',
+  tenantName: 'your-tenant',
+  secret: 'your-secret',
+  organizationId: '<organizationId>' // optional
+});
```
Raina451: i dont think this is needed to be added in public facing docs?

src/utils/runtime/constants.ts
```diff
@@ -9,6 +9,8 @@ export enum UiPathMetaTags {
   CLIENT_ID = 'uipath:client-id',
   SCOPE = 'uipath:scope',
   ORG_NAME = 'uipath:org-name',
+  /** Organization GUID — injected by deployments alongside the org name. */
```
Raina451: remove comment.

src/core/organization/organization-id-resolver.ts
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * Starts an OAuth authorization request with `acr_values=tenantName:<orgName>` and reads
+ * the redirect it produces instead of following it. No token is sent and no login is
+ * completed — Identity resolves the name while preparing the login page.
+ *
+ * @param baseUrl - Normalized SDK base URL (no trailing slash)
+ * @param orgName - Organization logical name
+ * @returns The organization GUID
+ */
+export async function lookupOrganizationIdByName(baseUrl: string, orgName: string): Promise<string> {
```
Raina451: nit: this is only for Pat tokens right? can you use naming getOrganizationIdByNameForPat or something?

src/core/organization/organization-id-resolver.ts
```diff
@@ -0,0 +1,240 @@
...
+  let url = buildAuthorizeUrl(baseUrl, orgName);
+
+  for (let hop = 0; hop < MAX_REDIRECT_HOPS; hop++) {
+    const response = await fetch(url, { redirect: 'manual', headers: tracingHeaders() });
+    const location = response.headers.get('location');
```
Raina451, line 168: Why raw fetch here instead of fetchWithRetry (ref: fetch-with-retry.ts) ?

src/services/base.ts
```diff
@@ -80,6 +82,17 @@ export class BaseService {
...
+  }
+
+  /**
+   * Resolves the organization (partition) GUID for APIs that key on it rather than the
+   * organization name. Resolved once per SDK instance and shared across services.
+   *
+   * @returns Promise resolving to the organization GUID
+   */
+  protected resolveOrganizationId(): Promise<string> {
+    return this.#organizationIdResolver.resolve();
```
Raina451: putting resolveOrganizationId in baseservice means every service gets a method that silently call Identity. can it only live on services where such thing is needed? eg: user service

src/models/platform/platform.models.ts
```diff
@@ -11,10 +11,24 @@ import type {
...
+ * Prerequisites: Initialize the SDK first - see [Getting Started](/uipath-typescript/getting-started/#import-initialize)
+ *
+ * ```typescript
+ * import { Platform, PlatformSettingKey } from '@uipath/uipath-typescript/platform';
+ *
+ * const platform = new Platform(sdk);
+ * const settings = await platform.getUserSettings(
+ *   [PlatformSettingKey.UserTheme],
+ *   '<userId>',
+ *   { organizationId: '<organizationId>' }
```
Raina451, line 29: why orgid?

src/models/platform/users.types.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * A single failure reported by a user update.
+ */
+export interface PlatformUserUpdateError {
+  /** Machine-readable error code. */
+  code: string;
+  /** Human-readable description of the failure. */
+  description: string;
+}
```
Raina451: is separate error interface needed? cant this fall into standardised errors sdk have?

src/models/platform/users.types.ts
```diff
@@ -0,0 +1,154 @@
...
+  /** Display name shown in the UI. */
+  displayName: string | null;
+  /** When the account was created. */
+  createdTime: string;
+  /** When the account was last modified. */
+  lastModifiedTime: string | null;
+  /** When the user last signed in. */
+  lastLoginTime: string | null;
+  /** GUIDs of the groups the user belongs to — empty when the user is in none. */
+  groupIds: string[];
```
Raina451, line 90: how to know the group name? because that what consumers would be intersted in. will there be a separate api for this?

src/models/platform/users.models.ts
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * Returns the user's profile, activity timestamps, and group memberships
+   * (`groupIds`).
+   *
+   * @param userId - GUID of the user
+   * @returns The user, as a {@link PlatformUserGetResponse}
+   *
+   * @example
+   * ```typescript
+   * const user = await users.getById('<userId>');
```
Raina451: add example on how to get user id.

src/models/platform/users.models.ts
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * @param userId - GUID of the user to update
+   * @param update - The fields to change
+   * @returns The update result, as a {@link PlatformUserUpdateResponse}
+   *
+   * @example Add a user to a group
+   * ```typescript
+   * const result = await users.updateById('<userId>', {
+   *   groupIdsToAdd: ['<groupId>'],
+   * });
```
Raina451, line 124: same here

.github/workflows/coverage.yml
```diff
@@ -122,7 +124,9 @@ jobs:
           FUNCTIONS_TEST_FOLDER_ID=${{ steps.config.outputs.functions_test_folder_id }}
           FUNCTIONS_TEST_FUNCTION_NAME=${{ steps.config.outputs.functions_test_function_name }}
           IDENTITY_TEST_USER_ID=${{ steps.config.outputs.identity_test_user_id }}
+          IDENTITY_MUTABLE_TEST_USER_ID=${{ steps.config.outputs.identity_mutable_test_user_id }}
           UIPATH_ORGANIZATION_ID=${{ steps.config.outputs.organization_id }}
+          UIPATH_TENANT_ID=${{ steps.config.outputs.tenant_id }}
```
Raina451: why is tenant id needed now and was not needed before?

src/models/platform/users.models.ts
Sarath1018: Done — `getAll()` no longer takes an organization id; the service resolves it internally (see the resolver commit).

docs/authentication.md
Sarath1018: Removed the section in 8993fc268. Whether the `organizationId` config field itself should stay public is the open question with Naren on the Slack thread — happy to drop it too if that's the call.

src/utils/runtime/constants.ts
Sarath1018: Removed in 8993fc268.

src/core/organization/organization-id-resolver.ts
Sarath1018: Renamed to `lookupOrganizationIdForPat` and the JSDoc now says why this is the PAT path (opaque token, no org claim) — 8993fc268.

src/core/organization/organization-id-resolver.ts
Sarath1018: No good reason — switched to `fetchWithRetry` with the SDK's default retry policy (2 retries on 408/429/5xx and network errors), plus a unit test covering a 503 followed by the redirect. 8993fc268.

src/services/base.ts
Sarath1018: Agreed. Moved out of `BaseService`: the Users service now holds the resolver itself (fetched from `SDKInternalsRegistry` in its constructor, still one shared resolver per SDK instance). Groups/Directory/Roles will do the same. 8993fc268.

src/models/platform/platform.models.ts
Sarath1018: There is a separate PR from naren to remove orgid from platform settings, will handle this there

src/models/platform/users.types.ts
Sarath1018: Good call — dropped `PlatformUserUpdateResponse`/`PlatformUserUpdateError`. `updateById` (and the bound `user.update()`) now returns `Promise<void>` and throws the SDK's `ValidationError` when Identity answers `succeeded: false`, with the reasons in the message (e.g. `Email is already taken. (DuplicateEmail)`). 8993fc268.

src/models/platform/users.types.ts
Sarath1018: Yes — the Groups service in #664 (`@uipath/uipath-typescript/groups`) returns id + name via `groups.getAll()`/`getById()`; `groupIds` here is the join key. Identity's user endpoints only return the ids, so a lookup through Groups is the way to get names.

src/models/platform/users.models.ts
Sarath1018: Added an example that finds the user by email via `users.getAll({ searchTerm })` and then calls `getById(items[0].id)` — 8993fc268.

src/models/platform/users.models.ts
Sarath1018: Will add in upcoming groups PR

.github/workflows/coverage.yml
Sarath1018: It isn't needed by this PR — I wired it while adding the mutable-user secret because the notification suite on main reads a tenant id, but that's unrelated (and it reads the `_DEV` name anyway). Removed in 8993fc268; I'll fix the notification wiring separately.

Raina451 (approved):

PR #669, opened 2026-08-13
feat(functions): acquire a Studio Web license before invoking [APPS-37123]

amrit-agarwal-1:
Coded Apps do not require a user license, yet they need to run work on Orchestrator. functions.invoke() now calls POST /api/StudioWeb/AcquireLicense first, so Orchestrator acquires the caller's own license or falls back to the free Attended Studio Web one.

Only the success matters. Nothing from the response is sent to the function -- the call is a precondition that provisions the personal robot the invocation requires, which is why a service principal fails with "Cannot find a personal robot configured". The only value read is the license expiry, used as the cache TTL.

Measured on alpha across four license tiers (none / basic / plus / pro): the acquisition is a flat 268-443 ms regardless of tier. Uncached it takes a ~890 ms invoke to ~1160 ms; cached it costs nothing.

- TTL comes from the license token's own exp claim, not a hardcoded guess.
- The cache is process-wide, keyed by tenant + user. Consumers routinely build a service per request or per render, and an instance field would start empty each time -- paying the round trip on every invocation, which is what caching is meant to avoid. Tenant is in the key because the cache outlives any one service instance.
- Concurrent invocations collapse to one acquisition: the pending promise is cached, not just the resolved value.
- A failed acquisition is never cached, so the next call retries.
- Bounded at 500 entries so a long-lived server does not accumulate one per user it has ever served.
- In memory only: coded apps in an org share one origin, so persisting would share the cache across every app, and a license token is credential-shaped.
- Licensing runs in parallel with the name lookup, which needs no license, so only the slower of the two is paid.
- Fail-open: a licensing failure warns and invokes anyway.

refreshLicense forces a fresh acquisition and is the only new public option.

### DIFF
--- a/src/models/orchestrator/functions.internal-types.ts
+++ b/src/models/orchestrator/functions.internal-types.ts
@@ -64,10 +64,12 @@ export interface RawStudioWebLicenseResponse {
   /** Last update of the licensed user session (ISO 8601). */
   lastUpdated: string;
   /**
-   * Unsigned JWT carrying the granted licence units and its own validity
-   * window. Its `exp` claim drives the SDK's cache TTL.
+   * Unsigned JWT carrying the granted license units and its own validity
+   * window; its `exp` claim is what the SDK reuses the license against. Null
+   * when the platform issues no token, in which case the SDK holds the license
+   * only for the fallback lifetime.
    */
-  licenseToken: string;
+  licenseToken: string | null;
 }
 
 /** Claims the SDK reads from the unsigned license token. Never verified. */
--- a/src/models/orchestrator/functions.internal-types.ts
+++ b/src/models/orchestrator/functions.internal-types.ts
@@ -3,6 +3,7 @@
  * transformation. Not exported through the public barrel.
  */
 
+
 /**
  * Raw HTTP trigger row from `GET /odata/HttpTriggers` after
  * `pascalToCamelCaseKeys()`. Only the fields consumed by the service are
@@ -44,3 +45,84 @@ export interface RawFolderResponse {
   /** Folder key (GUID) — the `t/{key}` segment of the function invoke URL. */
   Key: string;
 }
+
+/**
+ * Raw response from `POST /api/StudioWeb/AcquireLicense`. The endpoint already
+ * answers in camelCase, so no key transform is applied.
+ */
+export interface RawStudioWebLicenseResponse {
+  /** Robot type the license was granted for, e.g. `StudioX`. */
+  robotType: string;
+  /** Every robot type the acquired license covers. */
+  robotTypes: string[];
+  /** Whether the license came from an external provider. */
+  externalLicense: boolean;
+  /** Whether the caller ended up licensed. */
+  isLicensed: boolean;
+  /** Start of the licensed user session (ISO 8601). Refreshed on every call. */
+  started: string;
+  /** Last update of the licensed user session (ISO 8601). */
+  lastUpdated: string;
+  /**
+   * Unsigned JWT carrying the granted licence units and its own validity
+   * window. Its `exp` claim drives the SDK's cache TTL.
+   */
+  licenseToken: string;
+}
+
+/** Claims the SDK reads from the unsigned license token. Never verified. */
+export interface StudioWebLicenseTokenClaims {
+  /** Expiry, seconds since the Unix epoch. Drives the cache lifetime. */
+  exp?: number;
+  /** Not-before, seconds since the Unix epoch. */
+  nbf?: number;
+  /** User's base license tier, e.g. `BASICNU`. Absent for an unlicensed user. */
+  ubl?: string;
+  /** Licensed units granted, e.g. `['APPS', 'STDW', 'AGENT']`. */
+  lu?: string[];
+  /** Validity of the token, e.g. `VALID`. */
+  status?: string;
+}
+
+/** A license acquired for the calling user. */
+export interface StudioWebLicense {
+  /** Robot type the license was granted for, e.g. `StudioX`. */
+  robotType: string;
+  /** Every robot type the acquired license covers. */
+  robotTypes: string[];
+  /** Whether the caller ended up licensed. */
+  isLicensed: boolean;
+  /** Start of the licensed user session (ISO 8601). Refreshed on every acquisition. */
+  startedTime: string;
+  /** When the license stops being valid (ISO 8601), read from the token. */
+  expiresTime?: string;
+  /** The user's base license tier, e.g. `BASICNU`. Absent for an unlicensed user. */
+  licenseTier?: string;
+  /** Licensed units the token grants. */
+  licensedUnits?: string[];
+}
+
+/**
+ * Options for acquiring a license directly.
+ *
+ * @internal
+ */
+export interface FunctionAcquireLicenseOptions {
+  /**
+   * Acquires a fresh license instead of returning the one already held.
+   * Defaults to `false`.
+   */
+  refresh?: boolean;
+}
+
+/** A resolved or in-flight license acquisition held in the service's cache. */
+export interface LicenseCacheEntry {
+  /**
+   * The acquisition. Stored while still pending so concurrent invokes share one
+   * round trip.
+   */
+  acquisition: Promise<StudioWebLicense>;
+  /** Epoch milliseconds after which the entry is stale. */
+  expiresAtMs: number;
+}
+
--- a/src/models/orchestrator/functions.types.ts
+++ b/src/models/orchestrator/functions.types.ts
@@ -63,6 +63,16 @@ export type FunctionGetAllOptions = RequestOptions & PaginationOptions & FolderS
 export interface FunctionInvokeOptions extends FolderScopedOptions {
   /** Key (GUID) of the job this invocation belongs to, so the run is attributed to it. */
   jobKey?: string;
+  /**
+   * Acquires a fresh license instead of reusing the one already held for this
+   * user. Defaults to `false`.
+   *
+   * A license is cached for as long as the platform says it is valid — currently
+   * two hours, read from the license itself rather than assumed — so a burst of
+   * invocations costs one acquisition. Set this when the user's licensing may
+   * have changed and the invocation must reflect it before that lapses.
+   */
+  refreshLicense?: boolean;
 }
 
 /**
--- a/src/models/orchestrator/functions.types.ts
+++ b/src/models/orchestrator/functions.types.ts
@@ -69,7 +69,9 @@ export interface FunctionInvokeOptions extends FolderScopedOptions {
    *
    * A license is reused for as long as it says it is valid, read from the
    * license rather than assumed, so a burst of invocations costs one
-   * acquisition. A license that states no validity window is held only briefly.
+   * acquisition. A license that states no validity window — as the free tier's
+   * does not — is reused for five minutes and then re-acquired.
+   *
    * Set this when the user's licensing may have changed and the invocation must
    * reflect it before that lapses.
    */
--- a/src/models/orchestrator/functions.types.ts
+++ b/src/models/orchestrator/functions.types.ts
@@ -67,10 +67,11 @@ export interface FunctionInvokeOptions extends FolderScopedOptions {
    * Acquires a fresh license instead of reusing the one already held for this
    * user. Defaults to `false`.
    *
-   * A license is cached for as long as the platform says it is valid — currently
-   * two hours, read from the license itself rather than assumed — so a burst of
-   * invocations costs one acquisition. Set this when the user's licensing may
-   * have changed and the invocation must reflect it before that lapses.
+   * A license is reused for as long as it says it is valid, read from the
+   * license rather than assumed, so a burst of invocations costs one
+   * acquisition. A license that states no validity window is held only briefly.
+   * Set this when the user's licensing may have changed and the invocation must
+   * reflect it before that lapses.
    */
   refreshLicense?: boolean;
 }
--- a/src/services/orchestrator/functions/functions.ts
+++ b/src/services/orchestrator/functions/functions.ts
@@ -43,12 +43,16 @@ import type { IUiPath } from '../../../core/types';
 const MAX_SUGGESTED_NAMES = 20;
 
 /**
- * How long a license is reused when it states no expiry of its own — not every
- * grant carries a token to read one from. Short, so a license the SDK cannot
- * reason about is re-checked soon; long enough that a burst still costs one
- * acquisition.
+ * How long a license is reused when it states no expiry of its own — the free
+ * grant carries no token to read one from.
+ *
+ * Deliberately short. Where a license states a window the SDK follows it; where
+ * the platform states none, re-checking soon keeps the SDK from honouring a
+ * licensing change later than it happens. Re-checking costs almost nothing: the
+ * acquisition runs in parallel with the name lookup, so it hides inside a leg
+ * the invocation already pays for.
  */
-const DEFAULT_LICENSE_TTL_MS = 5 * 60 * 1000;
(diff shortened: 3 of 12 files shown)


src/models/orchestrator/functions.types.ts
```diff
@@ -63,8 +63,17 @@ export type FunctionGetAllOptions = RequestOptions & PaginationOptions & FolderS
 export interface FunctionInvokeOptions extends FolderScopedOptions {
   /** Key (GUID) of the job this invocation belongs to, so the run is attributed to it. */
   jobKey?: string;
+  /**
+   * Acquires a fresh license instead of reusing the one already held for this
+   * user. Defaults to `false`. Set it when the user's licensing may have changed
+   * and the invocation must reflect it immediately.
```
vnaren23: Whats the lifetime of our cache? Lets mention the same on the comments

src/services/orchestrator/functions/functions.ts
```diff
@@ -235,6 +376,29 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
 
+
+/**
+ * Epoch milliseconds at which a license should be renewed, read from the
+ * unsigned license token's `exp`. Returns `undefined` when the token cannot be
+ * read, which costs only the fallback TTL.
+ */
+export function licenseExpiryMs(licenseToken: string): number | undefined {
+  try {
+    const payload = licenseToken?.split('.')[1];
```
vnaren23: Move token decoding to a helper method so that it can be reused.

src/services/base.ts
```diff
@@ -79,7 +80,7 @@ export class BaseService {
   constructor(instance: IUiPath, headers?: Record<string, string>) {
     const { config, context, tokenManager, folderKey } = SDKInternalsRegistry.get(instance);
     this.#apiClient = new ApiClient(config, context, tokenManager, headers ? { headers } : {});
-    this.config = { folderKey };
+    this.config = { folderKey, orgName: config.orgName, tenantName: config.tenantName };
```
vnaren23: why is this needed? cc: @Raina451

tests/utils/mocks/functions.ts
```diff
@@ -1,10 +1,51 @@
 import { RawFunctionGetResponse, FunctionHttpMethod } from '../../../src/models/orchestrator/functions.types';
+import { RawStudioWebLicenseResponse, StudioWebLicenseTokenClaims } from '../../../src/models/orchestrator/functions.internal-types';
```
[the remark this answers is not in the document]
amrit-agarwal-1: fixed

src/services/orchestrator/functions/functions.ts
```diff
@@ -235,6 +376,29 @@ export class FunctionService extends FolderScopedService implements FunctionServ
   }
 }
 
+
+/**
+ * Epoch milliseconds at which a license should be renewed, read from the
+ * unsigned license token's `exp`. Returns `undefined` when the token cannot be
+ * read, which costs only the fallback TTL.
+ */
+export function licenseExpiryMs(licenseToken: string): number | undefined {
```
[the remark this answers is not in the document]
amrit-agarwal-1: fixed

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: Refactored

src/services/base.ts
amrit-agarwal-1: It was not needed indeed, reverted. The functions service reads org/tenant from SDKInternalsRegistry in its own constructor.

src/models/orchestrator/functions.types.ts
amrit-agarwal-1: Its 2 hours- added to the JSDoc. It's read from the token's own exp rather than hardcoded, so it follows the platform if that ever changes.

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +165,102 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+    } catch (error) {
+      // Never cache a rejection — the next call must retry, not replay it.
+      if (cache.get(cacheKey) === entry) cache.delete(cacheKey);
+      throw error;
+    }
+  }
+
+  /** Performs the acquisition. Takes no body and is not folder-scoped. */
+  private async requestLicense(): Promise<StudioWebLicense> {
+    const response = await this.post<RawStudioWebLicenseResponse>(STUDIO_WEB_LICENSE_ENDPOINTS.ACQUIRE);
```
ninja-shreyash, line 260: what would happen if org exhausted its free licenses?

src/models/orchestrator/functions.types.ts
```diff
@@ -63,6 +63,16 @@ export type FunctionGetAllOptions = RequestOptions & PaginationOptions & FolderS
 export interface FunctionInvokeOptions extends FolderScopedOptions {
   /** Key (GUID) of the job this invocation belongs to, so the run is attributed to it. */
   jobKey?: string;
+  /**
+   * Acquires a fresh license instead of reusing the one already held for this
+   * user. Defaults to `false`.
+   *
+   * A license is cached for as long as the platform says it is valid — currently
```
ninja-shreyash: Doc claims 2 hrs. But for default license - the time is only 5 mins?

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +165,102 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+      // Optimistic until the real expiry arrives, so a pending entry is never
+      // read as stale by a concurrent caller.
+      expiresAtMs: now + DEFAULT_LICENSE_TTL_MS,
+    };
+    evictIfFull(cache);
+    cache.set(cacheKey, entry);
+
+    try {
+      const license = await entry.acquisition;
+      entry.expiresAtMs = licenseExpiryMs(license.expiresTime) ?? now + DEFAULT_LICENSE_TTL_MS;
```
ninja-shreyash: does active license also guarentee that robot is active? because we will cache the license but the robot could timeout if left idle

src/services/orchestrator/functions/functions.ts
```diff
@@ -27,17 +36,91 @@ import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '.
...
 import { PaginationType } from '../../../utils/pagination/internal-types';
 import { track } from '../../../core/telemetry';
+import { SDKInternalsRegistry } from '../../../core/internals';
+import type { IUiPath } from '../../../core/types';
 
 /** Cap on the function names listed when a name lookup misses. */
 const MAX_SUGGESTED_NAMES = 20;
 
+/** Fallback TTL when a license token carries no readable `exp`. */
+const DEFAULT_LICENSE_TTL_MS = 5 * 60 * 1000;
```
ninja-shreyash: Question: why 5 mins for fallback?

src/models/orchestrator/functions.internal-types.ts
```diff
@@ -44,3 +45,84 @@ export interface RawFolderResponse {
...
+  isLicensed: boolean;
+  /** Start of the licensed user session (ISO 8601). Refreshed on every call. */
+  started: string;
+  /** Last update of the licensed user session (ISO 8601). */
+  lastUpdated: string;
+  /**
+   * Unsigned JWT carrying the granted licence units and its own validity
+   * window. Its `exp` claim drives the SDK's cache TTL.
+   */
+  licenseToken: string;
```
ninja-shreyash: Isn't it null for free license?
```suggestion
  licenseToken: string | null;
```

ninja-shreyash (commented):
Do we need to update oauth-scopes.md since `invoke()` is calling one extra endpoint?

tests/unit/services/orchestrator/functions.test.ts
```diff
@@ -494,4 +525,349 @@ describe('FunctionService Unit Tests', () => {
...
+      mockApiClient.post.mockResolvedValueOnce(createMockRawStudioWebLicense());
+
+      const license = await functionService.acquireLicense();
+
+      expect(license.isLicensed).toBe(true);
+      expect(license.licenseTier).toBe(FUNCTION_LICENSE_TEST_CONSTANTS.LICENSE_TIER);
+      expect(license.startedTime).toBe(FUNCTION_LICENSE_TEST_CONSTANTS.STARTED);
+      // Only the acquisition — nothing was invoked
+      expect(mockApiClient.post).toHaveBeenCalledTimes(1);
+    });
```
Sarath1018, line 857: Assert the transformations in test cases, check in other places as well

```
expect((license as any).started).toBeUndefined();
expect((license as any).ubl).toBeUndefined();
expect((license as any).lu).toBeUndefined();
```

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +165,102 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+      if (cached && cached.expiresAtMs > now) return cached.acquisition;
+    }
+
+    const entry: LicenseCacheEntry = {
+      acquisition: this.requestLicense(),
+      // Optimistic until the real expiry arrives, so a pending entry is never
+      // read as stale by a concurrent caller.
+      expiresAtMs: now + DEFAULT_LICENSE_TTL_MS,
+    };
+    evictIfFull(cache);
```
Sarath1018:
```suggestion
    if (!cache.has(cacheKey)) evictIfFull(cache);
```

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +165,102 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+    if (licensing.status === 'rejected') throw licensing.reason;
+    if (lookup.status === 'rejected') throw lookup.reason;
+
+    const fn = lookup.value;
     const folderKey = await this.resolveInvokeFolderKey(fn, folderOptions);
     return this.invokeFunction<TOutput>(fn, folderKey, input ?? {}, jobKey);
   }
 
+
+
```
Sarath1018: nit: one blank line should be fine

Sarath1018:
invoke() now also calls `POST /api/StudioWeb/AcquireLicense`  can you confirm OR.Default covers that endpoint also

src/services/orchestrator/functions/functions.ts
```diff
@@ -27,17 +36,91 @@ import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '.
...
+  return (store[LICENSE_CACHE_KEY] ??= new Map<string, LicenseCacheEntry>());
+}
+
+/**
+ * Empties the license cache. The cache is process-wide, so tests need this to
+ * avoid inheriting licenses from each other.
+ *
+ * @internal
+ */
+export function clearLicenseCache(): void {
```
anil-uipath: Is this unused? I don't see any callers. can we remove?

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +165,102 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+      // Awaiting a pending entry is the point: concurrent invokes share one
+      // round trip rather than each issuing their own.
+      if (cached && cached.expiresAtMs > now) return cached.acquisition;
+    }
+
+    const entry: LicenseCacheEntry = {
+      acquisition: this.requestLicense(),
+      // Optimistic until the real expiry arrives, so a pending entry is never
+      // read as stale by a concurrent caller.
+      expiresAtMs: now + DEFAULT_LICENSE_TTL_MS,
```
anil-uipath: can we avoid storing the dummy value for expiresAtMs if possible. it appears to be rewritten later when the actual expiresAt comes.

Looks like `acquisition` is already cached as a promise, and L236 is awaiting for it.
pls explore using the promise itself for the concurrent case in L231. I think that will simplify the try-catch logic.

```
type LicenseCacheEntry = Promise<{ license: StudioWebLicense; expiresAtMs: number }>;

const acquisition = this.requestLicense().then((license) => ({
...
```

amrit-agarwal-1: Tried that, evictIfFull reads expiresAtMs synchronously to drop expired entries before live ones, so it can't move inside the promise. It also stops a never-settling request from capturing callers forever. Comment was misleading though, fixed that.

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: fixed

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: fixed

tests/unit/services/orchestrator/functions.test.ts
amrit-agarwal-1: Added. Used 'started' in license instead of `as any`: no `any`, and stricter.

src/models/orchestrator/functions.internal-types.ts
amrit-agarwal-1: fixed

src/models/orchestrator/functions.types.ts
amrit-agarwal-1: 2h is licensed-only. Free tier returns no token, so nothing to read → 5 min. Dropped the hardcoded 2h, renamed it FALLBACK_LICENSE_TTL_MS, and corrected the docs

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: This only kicks in when the license states no expiry (free tier returns no token), so there's nothing to derive it from. Bounded either side: long enough to collapse a burst into one acquire, short enough that we're not honouring a stale licensing change for hours.

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: I haven't seen a case yet where license is active but robot is inactive. Have asked Sorin for confirmation

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: If the API fails in that case, then we fail similarly. Its difficult to repro this case, have asked Sorin for confirmation.

amrit-agarwal-1:
> Do we need to update oauth-scopes.md since `invoke()` is calling one extra endpoint?

Tested it, the endpoint works with the scopes invoke() already needs, nothing extra. Added a line to the doc mentioning the call.

src/services/orchestrator/functions/functions.ts
```diff
@@ -82,20 +174,105 @@ export class FunctionService extends FolderScopedService implements FunctionServ
...
+    const [licensing, lookup] = await Promise.allSettled([
+      this.resolveLicense(refreshLicense),
+      this.findByName(func.name, folderOptions),
+    ]);
+
+    // Licensing is a precondition, not an optimisation: acquiring is what
+    // provisions the personal robot the invocation runs on. Without it the
+    // invocation fails anyway, further in, as "no personal robot configured" —
+    // so surface the licensing failure instead of that.
+    if (licensing.status === 'rejected') throw licensing.reason;
```
anil-uipath, line 206: For 4xx codes, if promise gets rejected it is fine.  if it succeeds it is not. pls test if not already. ex: 4xx - no folder permission (or) license not allocated for user

ninja-shreyash (approved):

src/services/orchestrator/functions/functions.ts
amrit-agarwal-1: 4xx already rejects - tested with a typed 4xx, and we saw a real 403 propagate and fail the invoke in CI too. License-not-allocated doesn't 4xx though - tested, it returns 200 with the free Attended Studio Web grant. Folder permission doesn't apply to this call since it's not folder-scoped.

Sarath1018 (approved):

PR #670, opened 2026-08-13
test(integration): authenticate Integration tests with Minter using user token

amrit-agarwal-1:
Six suites - Agents, Agent Memory, Agent Traces, Governance, Notifications, Subscriptions , were permanently describe.skip-ed: their APIs reject PAT and client-credentials tokens with 401 regardless of scopes, and the required scopes aren't available to external applications.

CI now mints a user access token with Minter (Portal team's headless-login image, pulled from ACR with a scoped token) and writes it to UIPATH_USER_TOKEN. Those six suites run under it.

Suites are declared with describeIntegration(name, requirement, modes, body), which expands them over init modes × credentials and names each cell [initMode][authMode], so a failure says which credential failed. The requirement is stated once and drives the credential, the host and the collection-time skip guard — declaring the guard separately let the two disagree.

'both' — the default — runs a suite once per configured credential. 23 suites run under PAT and the user token: the PAT exercises the external-application scope model most SDK consumers use, the user token exercises the API surface generally. Neither subsumes the other.

User-token cells use MINTER_BASE_URL (the platform host). The default host is a CORS proxy whose path whitelist must name every service a suite touches.

Also in here:

getActiveAuth() gives the four raw-fetch call sites the credential and host of the cell actually running; config.userToken ?? config.secret was correct only while a suite ran under one credential.
Folder-filter tests use INTEGRATION_TEST_FOLDER_KEY instead of fabricated GUIDs — folderKeys is authorized before it's applied, so an inaccessible folder returns 403, an assumption these tests documented but had never run to verify.
Fixtures are created and owned by the suites that read them.
Five tests are skipped with the blocker named: a notification-service $filter stall, topic groups absent from the tenant, and three agent-activity data gaps.

Local runs are unchanged — with only UIPATH_SECRET, 'both' suites run PAT-only and user suites skip. Minter stays optional.

Verified against alpha procodeapps/integrationtest.


tests/integration/shared/agents/agents.integration.test.ts
```diff
@@ -14,24 +15,31 @@ import { AGENT_TEST_CONSTANTS } from '../../../utils/constants';
...
 const modes: InitMode[] = ['v1'];
 
-// skip: insightsrtm_ endpoints reject PAT tokens entirely (401 regardless of scopes) and require OAuth.
-// Skipped at the outer level so the live-auth setup (setupUnifiedTests + beforeAll) does not run.
-describe.skip.each(modes)('Agents - Integration Tests [%s]', (mode) => {
-  setupUnifiedTests(mode);
-
+// insightsrtm_ rejects PAT tokens entirely (401 regardless of scopes), so this
+// suite authenticates with a user token and skips when one is not configured.
+describeIntegration('Agents - Integration Tests', 'user', modes, () => {
```
Raina451, line 20: do we want to have minter based tests for all services eventually?

would it makes sense to have 2 folders under tests/integration , one for pat based, one for minter based. This would be easily readable segregation and will give us better hand in extending in future.

for the apis which dont support pat, we can add them in minter based only.

amrit-agarwal-1: Most suites aren't in one camp - 20 of 26 now run under both credentials, so folders would mean either duplicating those files or forcing each into one permanently. The requirement is already one word on the suite (describeIntegration(..., 'both' | 'user', ...)), which greps just as well.

Raina451 (approved):

Sarath1018:
we are running vo mode with PAT then user token and again v1 mode with PAT and user all serially. This will double our integration tests run time. I dont see a way to run them parallel without redesigning the entire integration tests. I would suggest to increase the runner cpu from 2 to 4 and increase workercount to 4 which will increase the concurrence between different test suits and improve the time

tests/integration/shared/maestro/processes.integration.test.ts
Sarath1018: can we unskip insights tests?

tests/integration/config/test-config.ts
```diff
@@ -208,6 +240,54 @@ export function loadIntegrationConfig(): IntegrationConfig {
...
+ * cells reach services missing from it.
+ */
+export function resolveBaseUrl(config: IntegrationConfig, authMode: AuthMode): string {
+  return authMode === 'user' ? config.minterBaseUrl ?? config.baseUrl : config.baseUrl;
+}
+
+/** Whether any configured credential can satisfy the requirement. */
+export function canAuthenticate(requirement: AuthRequirement): boolean {
+  return resolveAuthModes(requirement).length > 0;
+}
```
Sarath1018: Where are we using this?

tests/integration/config/unified-setup.ts
```diff
@@ -175,20 +202,30 @@ function createV1Services(config: IntegrationConfig): TestServices {
...
+ * Initialize services in the specified init mode and auth mode.
+ *
+ * The cached instance is keyed on both — a suite running under one credential
+ * must never be handed the SDK built for the other.
  */
-export async function initializeServices(mode: InitMode): Promise<TestServices> {
-  if (servicesInstance && currentMode === mode) {
+export async function initializeServices(
+  mode: InitMode,
+  authMode: AuthMode = 'pat'
```
Sarath1018: Why defaulting to pat? looks like it will be set always. check if this can be removed

.github/workflows/coverage.yml
```diff
@@ -129,6 +138,84 @@ jobs:
...
+          TOKEN=$(python3 -c "import json;print(json.load(open('$RUNNER_TEMP/minter/tokens.json'))['accessToken'])")
+          if [ -z "$TOKEN" ]; then
+            echo "Minter produced no accessToken" >&2
+            exit 1
+          fi
+          # Mask before the value can reach any later log line, then write it
+          # straight into the env file. Deliberately NOT routed through
+          # $GITHUB_OUTPUT, which persists values in job metadata.
+          echo "::add-mask::$TOKEN"
+          echo "UIPATH_USER_TOKEN=$TOKEN" >> tests/.env.integration
```
Sarath1018, line 237: I think we reuse the infrastructure so better to clean this up

tests/.env.integration.example
```diff
@@ -18,6 +25,37 @@ UIPATH_TENANT_NAME=your-tenant-name
...
+#     npm run generate -- -u <email> -p <password> -n <org> -t <tenant> -e <env> -v basic -o /out/tokens.json
+# then copy the `accessToken` field from out/tokens.json.
+#
+# Note: the token is short-lived and is never refreshed by the SDK, so re-mint it
+# if a run starts failing with 401s. Docs: https://uipath.atlassian.net/wiki/spaces/CLD/pages/87134404744
+UIPATH_USER_TOKEN=
+
+# Tenant GUID. Required by the Notifications and Subscriptions suites, which pass
+# it as a header — those throw without it. Admin > Tenant > Tenant details.
+UIPATH_TENANT_ID_DEV=
```
Sarath1018: No need to add UIPATH prefix and _DEV suffix. for remaining examples as well.

tests/integration/shared/maestro/processes.integration.test.ts
amrit-agarwal-1: done

tests/integration/config/test-config.ts
amrit-agarwal-1: it was redundant, removed it now

tests/integration/config/unified-setup.ts
amrit-agarwal-1: Dropped default and made the param required

.github/workflows/coverage.yml
amrit-agarwal-1: Added rm -f tests/.env.integration with if: always() after the integration run, so the minted token doesn't survive on a reused runner.

tests/.env.integration.example
amrit-agarwal-1: removed

.github/workflows/coverage.yml
```diff
@@ -217,9 +217,10 @@ jobs:
           echo "Minted a user access token; user-token suites will run, and either-credential suites will run under both."
 
       - name: Run Integration Tests
-        # Network-bound suite: almost all of the wall clock is spent awaiting HTTP,
-        # so workers are oversubscribed relative to the runner's 2 vCPUs on purpose.
-        run: npm run test:integration:coverage -- --run --maxWorkers=4
+        # Network-bound suite, but the tenant is the bottleneck, not the runner.
+        # 4 workers bought 13% wall clock (1292s -> 1120s) and cost a wave of 504s,
+        # 503s and request timeouts that tripled the failing files. Stay at 2.
+        run: npm run test:integration:coverage -- --run --maxWorkers=2
```
Sarath1018: There will be no improvement by increasing maxWorkers without increasing cpu. check if we can increase cpu to 4 from 2. Also keep the comment in 1 line or remove it all together

Sarath1018 (approved):

PR #675, opened 2026-08-18
feat(core): zero-config authentication from the execution context

anil-uipath:
Closes [APPS-35692](https://uipath.atlassian.net/browse/APPS-35692).

## Problem

`new UiPath()` resolved configuration only from browser meta tags, so it always threw outside the browser — and the error told the caller to configure a bundler plugin, advice that cannot apply on a server:

> UiPath SDK configuration not found. Ensure @uipath/coded-apps plugin is set up in your bundler…

A coded function receives its platform coordinates as the handler's `ctx`, which the SDK could not reach. Authors therefore mapped it by hand, naming fields the SDK does not document as interchangeable — an org **id** passed as `orgName`, a tenant **id** as `tenantName`, a workload token as `secret` — and repeated that glue in every function.

## What changed

```ts
handler: async (_input, ctx) => {
  const sdk = new UiPath(ctx);      // coded function: pass the handler context
}

const sdk = new UiPath();           // script or test: reads the environment contract
```

- **`new UiPath(ctx)`** — the constructor also accepts a coded function's context and maps it onto the existing config fields. `CodedFunctionContext` mirrors the Functions SDK's type *structurally*, so neither package depends on the other. Configuration and a context are disjoint shapes, so the constructor tells them apart without the caller saying which was passed. A context whose `platform` is null falls through to the remaining sources.
- **`loadFromEnvironment()`** reads `UIPATH_BASE_URL` / `UIPATH_ORG_NAME` / `UIPATH_TENANT_NAME` / `UIPATH_ACCESS_TOKEN` — the names this repository already uses for the same values, so an existing `.env` works unchanged. Both org and tenant accept an id as readily as a name. Reads work under Node **and** Deno, are a no-op in browsers so meta tags remain the browser source, and treat empty values as absent.
- **No new config options.** The context and the environment both resolve to `orgName` / `tenantName` / `secret`, which already accept these values: org and tenant go into the same URL position either way, and `secret` is documented as taking a PAT or a bearer token.
- **Merge order:** constructor > meta tags > environment. Each layer is compacted before merging so a sparse layer cannot blank out the layer beneath it.
- **Missing-configuration errors** now match where the SDK is running, and a service built from an unconfigured instance reports that rather than `Invalid SDK instance`.

## Security

- The resolved token stays in `SDKInternalsRegistry` and is never placed on the public `config` object — [a test asserts this](tests/unit/core/uipath-zero-config.test.ts).
- Organization and tenant come only from the environment, an explicit config, or the runtime-supplied context — never from caller-controlled request data. This preserves the reason the `_baseUrl` / `_orgId` / `_tenantId` input-field pattern was deprecated on the Functions side.

## Compatibility

Fully backward compatible, and the public config shape is unchanged. `{ baseUrl, orgName, tenantName, secret }` and the OAuth/PKCE flow behave as before; browser meta-tag resolution is untouched and still takes precedence over the environment.

## Testing

- `npm run test:unit` — **2372 passing** (86 files), including new coverage for context mapping, the environment contract, merge precedence, and the unconfigured-instance diagnostic
- `npm run typecheck`, `npm run lint`, `npm run build` — all clean
- Exercised against a live tenant through all three paths — a coded function's `ctx`, the environment contract, and an explicit config — including from a function deployed to a runner

## Follow-ups (not in this PR)

- `ctx.platform.folderKey` is mirrored on the context type but not consumed; wiring it would let folder-scoped services resolve their folder from the invocation.
- The token manager treats `secret` as non-expiring, which does not hold for a short-lived bearer token.
- Version bump for release, per the separate-PR convention.

[APPS-35692]: https://uipath.atlassian.net/browse/APPS-35692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -85,13 +85,82 @@ To create a confidential app-scoped External App:
     The client-credentials flow does **not** support refresh tokens. To refresh, request a new token directly using the External App's client-credentials.
 
 
+## Coded Functions
+
+A coded function receives its platform coordinates as the handler's `ctx`. Pass
+that straight to the SDK — it maps the context itself, so the handler names no
+credential fields and never touches the workload token:
+
+```typescript
+import { defineFunction } from '@uipath/coded-functions-js-sdk';
+import { UiPath } from '@uipath/uipath-typescript/core';
+import { Entities } from '@uipath/uipath-typescript/entities';
+
+export default defineFunction({
+  name: 'list-entities',
+  handler: async (_input, ctx) => {
+    const sdk = new UiPath(ctx);
+
+    const entities = await new Entities(sdk).getAll();
+    return { entityCount: entities.length };
+  },
+});
+```
+
+Construct the SDK **inside the handler**, once per invocation. A single instance
+hoisted to module scope would keep serving the org, tenant and token of whichever
+invocation created it.
+
+## Server-side and scripts (environment contract)
+
+Outside the browser — a script, a test, a CI job — the SDK configures itself
+from environment variables, so `new UiPath()` needs no arguments:
+
+```typescript
+import { UiPath } from '@uipath/uipath-typescript/core';
+
+const sdk = new UiPath();
+```
+
+```bash
+UIPATH_BASE_URL=https://cloud.uipath.com
+UIPATH_ORG_NAME=<organization>          # an id works here as well as a name
+UIPATH_TENANT_NAME=<tenant>
+UIPATH_ACCESS_TOKEN=<accessToken>       # PAT or bearer token
+```
+
+These are the names this SDK already uses elsewhere for the same values, so one
+`.env` serves a script and a test run alike.
+
+The variables are read from `process.env` or `Deno.env`, whichever the runtime
+provides.
+
+To state the configuration explicitly instead, pass it to the constructor — this
+takes precedence over both the environment and meta tags:
+
+```typescript
+const sdk = new UiPath({
+  baseUrl: 'https://cloud.uipath.com',
+  orgName: '<organizationId>',   // an id works here as well as a name
+  tenantName: '<tenantId>',
+  secret: '<accessToken>'        // any bearer token
+});
+```
+
+The access token is consumed internally and is never exposed on `sdk.config`.
+
+!!! info "Precedence"
+    Constructor arguments win over meta tags, which win over the environment.
+    Meta tags apply in the browser only; the environment contract applies
+    outside it.
+
 ## SDK Initialization - The initialize() Method
 
 ### When to Use initialize()
 
 The `initialize()` method completes the authentication process for the SDK:
 
-- **Secret Authentication**: Auto-initializes when creating the SDK instance - **no need to call initialize()**
+- **Secret Authentication**: Auto-initializes when creating the SDK instance - **no need to call initialize()**. This covers `secret`, `accessToken`, a coded function's `ctx`, and the environment contract.
 - **OAuth Authentication**: **MUST call** `await sdk.initialize()` before using any SDK services
 
 ### Example: Secret Authentication (Auto-initialized)
--- a/src/core/config/environment.ts
+++ b/src/core/config/environment.ts
@@ -1,0 +1,73 @@
+import { PartialUiPathConfig } from './sdk-config';
+import { isBrowser } from '../../utils/platform';
+
+/**
+ * Shape of the globals that expose environment variables. Read defensively so
+ * the SDK works under Node and Deno, and stays inert in browser bundles where
+ * neither exists.
+ */
+interface EnvCapableGlobal {
+  process?: { env?: Record<string, string | undefined> };
+  Deno?: { env?: { get?(name: string): string | undefined } };
+}
+
+/**
+ * Reads a single environment variable, returning undefined for missing or
+ * empty values so a blank variable never satisfies a required field.
+ */
+export function readEnv(name: string): string | undefined {
+  const globals = globalThis as EnvCapableGlobal;
+
+  // Both reads are guarded: Deno throws without `--allow-env`, and its
+  // `process.env` shim proxies the same permission-gated `Deno.env`, so reading
+  // either one unguarded would escape the SDK constructor.
+  try {
+    const fromNode = globals.process?.env?.[name];
+    if (fromNode) return fromNode;
+  } catch {
+    // No env permission; fall through to the Deno accessor.
+  }
+
+  try {
+    const fromDeno = globals.Deno?.env?.get?.(name);
+    if (fromDeno) return fromDeno;
+  } catch {
+    // No env permission; treat the variable as absent.
+  }
+
+  return undefined;
+}
+
+/**
+ * The environment contract, matching the names this repository already uses for
+ * the same values in its integration configuration and documentation, so one
+ * `.env` serves both.
+ */
+export const UiPathEnvVars = {
+  BASE_URL: 'UIPATH_BASE_URL',
+  ORG_NAME: 'UIPATH_ORG_NAME',
+  TENANT_NAME: 'UIPATH_TENANT_NAME',
+  ACCESS_TOKEN: 'UIPATH_ACCESS_TOKEN',
+} as const;
+
+/**
+ * Load configuration from the execution-context environment contract.
+ *
+ * Returns null in browsers (where meta tags are the configuration source) and
+ * whenever no contract variable is present, so callers can fall through to
+ * other sources instead of acting on a half-populated config.
+ */
+export function loadFromEnvironment(): PartialUiPathConfig | null {
+  if (isBrowser) return null;
+
+  // Org and tenant accept an id as readily as a name — both go into the same URL
+  // position — and the token goes to `secret`, used verbatim as the bearer value.
+  const config: PartialUiPathConfig = {
+    baseUrl: readEnv(UiPathEnvVars.BASE_URL),
+    orgName: readEnv(UiPathEnvVars.ORG_NAME),
+    tenantName: readEnv(UiPathEnvVars.TENANT_NAME),
+    secret: readEnv(UiPathEnvVars.ACCESS_TOKEN),
+  };
+
+  return Object.values(config).some(Boolean) ? config : null;
+}
--- a/src/core/uipath.ts
+++ b/src/core/uipath.ts
@@ -3,10 +3,12 @@ import { ExecutionContext } from './context/execution';
 import { AuthService } from './auth/service';
 import { TokenInfo, LogoutOptions } from './auth/types';
 import { UiPathSDKConfig, PartialUiPathConfig, BaseConfig, hasOAuthConfig, hasSecretConfig } from './config/sdk-config';
-import { validateConfig, normalizeBaseUrl, isCompleteConfig } from './config/config-utils';
+import { validateConfig, normalizeBaseUrl, isCompleteConfig, compactConfig, missingConfigMessage } from './config/config-utils';
 import { telemetryClient, trackEvent } from './telemetry';
 import { SDKInternalsRegistry } from './internals';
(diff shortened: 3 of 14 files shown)


anil-uipath:
Addressed the review comments. Two applied, two not — reasoning below.

### Applied

**`tests/unit/core/uipath-zero-config.test.ts` — `const AuthService: any`**
Fixed with the suggested `Object.assign` form, so the static property is typed by intersection rather than `any`. `grep -c ": any"` on the file is now 0.

**`tests/unit/core/config/environment-browser.test.ts` — `CONTRACT_VARS` missing three names**
Correct, and the same list was duplicated across three suites, so I extracted it into `tests/utils/env-contract.ts` (`UIPATH_CONTRACT_VARS` + a `clearContractEnv()` snapshot/restore helper) and switched all three files to it. That fixes the missing `UIPATH_ORGANIZATION_NAME` / `UIPATH_ORG_NAME` / `UIPATH_SECRET` and removes the drift that caused it.

Verified the suite is now genuinely env-independent — identical results with those variables exported:

```
UIPATH_ORGANIZATION_NAME=amb UIPATH_ORG_NAME=amb UIPATH_SECRET=amb \
UIPATH_URL=https://x.uipath.com UIPATH_ACCESS_TOKEN=tok npm run test:unit
→ 84 files, 2361 tests passed   (same as a clean environment)
```

### Not applied

**`src/core/config/config-utils.ts:70` — `Object.fromEntries` resolving to `any`**

I don't think this one holds. `Object.fromEntries` resolves to its **typed** overload here, not the `any`-returning one — a mutable `[string, T][]` is assignable to `readonly [PropertyKey, T][]`, so the first overload matches.

Verified with an is-`any` probe:

```ts
const r = Object.fromEntries(Object.entries(config).filter(([, v]) => v !== undefined));
type IsAny<T> = 0 extends (1 & T) ? true : false;
const check: IsAny<typeof r> = true;   // error TS2322: Type 'true' is not assignable to type 'false'
```

The probe fails to compile, which means `r` is not `any`. Forcing the inferred type into an error message confirms what it actually is:

```
error TS2322: Type '{ [k: string]: string; }' is not assignable to type 'number'
```

So there's no implicit `any` to launder, and adding `as PartialUiPathConfig` would be the unnecessary cast `agent_docs/conventions.md` calls out: *"NEVER add unnecessary type casts on already-typed values."* Happy to add it if you'd still prefer the explicit annotation for readability.

**`src/core/uipath.ts:11` — class JSDoc still describes two config sources**

Already fixed in 4b6281a2a, which landed after the commit this review ran against. Current state:

```
 * Configuration is resolved from three sources, in precedence order:
 * 1. Config passed to the constructor — explicit, wins over everything
 * 2. Meta tags injected by the @uipath/coded-apps-dev plugin — browser only
 * 3. The execution-context environment contract — outside the browser
```

`initialize()`'s JSDoc was updated too, and a zero-config `@example` added.

---

typecheck clean · lint 0 errors · 2361 tests across 84 files · build exit 0

src/core/config/environment.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * The documented execution-context contract. A UiPath runner populates these;
+ * a developer running a plain script can set them by hand.
+ *
+ * Organization and tenant accept either an id or a logical name — the platform
+ * addresses both in the same URL position.
+ */
+export const UiPathEnvVars = {
```
amrit-agarwal-1, line 46: Do we need to define folder key also here? How will folder-scoped resources be handled with this approach?

src/core/config/environment.ts
```diff
@@ -0,0 +1,75 @@
...
+  BASE_URL: ['UIPATH_URL', 'UIPATH_BASE_URL'],
+  ORG: ['UIPATH_ORGANIZATION_ID', 'UIPATH_ORG_ID', 'UIPATH_ORGANIZATION_NAME', 'UIPATH_ORG_NAME'],
+  TENANT: ['UIPATH_TENANT_ID', 'UIPATH_TENANT_NAME'],
+  ACCESS_TOKEN: ['UIPATH_ACCESS_TOKEN', 'UIPATH_SECRET'],
```
amrit-agarwal-1: we are mapping access token to uipath secret, but TokenManager treats secret tokens as non-expiring. This is fine for PAT, but for our case it may fail for long-running invocations?

src/core/config/environment.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Reads a single environment variable, returning undefined for missing or
+ * empty values so a blank variable never satisfies a required field.
+ */
+export function readEnv(name: string): string | undefined {
+  const globals = globalThis as EnvCapableGlobal;
+
+  const fromNode = globals.process?.env?.[name];
```
amrit-agarwal-1: Guard looks one line too low- `process.env` is read outside the try, and Deno exposes both
globals with `process.env` as a proxy over Deno.env.get. If that's permission-gated too,
`deno run` without --allow-env throws here and escapes the constructor, which is what the
catch was for.

docs/authentication.md
```diff
@@ -188,13 +238,13 @@ useEffect(() => {
...
 
-Create `.env` file:
+Create `.env` file — the SDK reads these itself:
 ```bash
 # .env
-UIPATH_BASE_URL=https://api.uipath.com
-UIPATH_ORG_NAME=your-organization-name
-UIPATH_TENANT_NAME=your-tenant-name
-UIPATH_SECRET=your-pat-token
+UIPATH_URL=https://cloud.uipath.com
```
amrit-agarwal-1: This says cloud.uipath.com, the baseUrl examples further down still say api.uipath.com. Which is right?

src/core/config/environment.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * The documented execution-context contract. A UiPath runner populates these;
+ * a developer running a plain script can set them by hand.
+ *
+ * Organization and tenant accept either an id or a logical name — the platform
+ * addresses both in the same URL position.
+ */
+export const UiPathEnvVars = {
+  BASE_URL: ['UIPATH_URL', 'UIPATH_BASE_URL'],
+  ORG: ['UIPATH_ORGANIZATION_ID', 'UIPATH_ORG_ID', 'UIPATH_ORGANIZATION_NAME', 'UIPATH_ORG_NAME'],
```
amrit-agarwal-1: Preferring the ID spelling means a runner setting both gives us the org GUID in every URL.
Has that been exercised against the live API across services, or mainly tenant-by-Guid?

src/core/uipath.ts
```diff
@@ -144,21 +172,80 @@ export class UiPath implements IUiPath {
...
+    if (layers.length === 0) return undefined;
+
+    const merged = layers.reduce<PartialUiPathConfig>((acc, layer) => ({ ...acc, ...layer }), {});
+
+    // Secret and OAuth are mutually exclusive, so a plain merge would let an
+    // ambient token from the environment collide with an explicitly configured
+    // OAuth client and leave the config satisfying neither. The
+    // highest-precedence layer that supplies a complete auth method wins, and
+    // the competing method's fields are dropped.
+    for (let i = layers.length - 1; i >= 0; i--) {
```
deepeshrai-tech: Worth a one-line comment right here `sdk-config.ts` — something like `// per-layer check: does this layer *supply* a field of this auth method, not is it independently complete`.

src/core/config/sdk-config.ts
```diff
@@ -18,9 +18,19 @@ export type UiPathSDKConfig = BaseConfig & (
   | ({ secret?: never } & OAuthFields)
 );
 
+// Identifier/token aliases accepted by the constructor.
+// `orgId`/`tenantId` address org and tenant by id rather than logical name;
+// `accessToken` supplies an already-issued bearer token. Each is normalized
+// onto its canonical field (orgName/tenantName/secret) before validation.
+export interface ConfigAliases {
+  orgId: string;
```
deepeshrai-tech: `orgName` can now hold a UUID via `orgId` or `UIPATH_ORGANIZATION_ID`, since the platform addresses both at the same URL position. Someone reading `sdk.config.orgName` and seeing a GUID will be confused. A one-line JSDoc saying "accepts either a logical name or an id — the platform addresses both at the same URL position" would be good.

src/core/uipath.ts
```diff
@@ -144,21 +172,80 @@ export class UiPath implements IUiPath {
...
     const metaConfig = loadFromMetaTags();
     this.#metaFolderKey = metaConfig?.folderKey;
     this.#metaOrgId = metaConfig?.orgName;
     this.#metaTenantId = metaConfig?.tenantName;
 
-    // Merge with any partial config from constructor (constructor overrides meta tags)
-    const merged = { ...metaConfig, ...this.#partialConfig };
+    // Merge with any partial config from constructor (constructor overrides
+    // meta tags, which override the environment contract)
+    const merged = UiPath.#mergeConfigSources(metaConfig, this.#partialConfig);
```
deepeshrai-tech, line 243: nit — this re-runs the merge (which calls `loadFromEnvironment` and `loadFromMetaTags` again) each time `#loadConfig` is invoked, one line here — `// re-run to pick up meta tags injected after construction` — would make it clear.

docs/authentication.md
```diff
@@ -40,6 +40,56 @@ const sdk = new UiPath();
 await sdk.initialize();
 ```
 
+## Zero-config (execution context)
+
+Outside the browser the SDK configures itself from the execution-context
```
deepeshrai-tech: the environment-contract path covers Node and Deno specifically, but not edge runtimes (Cloudflare Workers, Vercel Edge, etc.) i am guessing its out of scope or we should add a liner for these runtimes not supported?

src/core/config/environment.ts
anil-uipath: folder is not required when initializing sdk. 
Even today in coded functions, they only include folder prop when requesting a resource at later time.

src/core/config/environment.ts
anil-uipath: This will be same flow as secret-based authentication, where it is developer's responsibility to refresh it, in case of expiry using `sdk.updateToken`.
and, long-running invocation may not be applicable for coded functions. the max duration is 25 sec, as per their documentation.

I have simplified the above logic to directly wire to current constructor params `secret: env.ACCESS_TOKEN`, and remove all but 4 env variables.

docs/authentication.md
anil-uipath: Good catch, changing to api.

src/core/config/sdk-config.ts
anil-uipath: removed orgId, and using orgName itself.
Confirmed [here](https://uipath-product.slack.com/archives/C0B8A8CCKM1/p1787552003040189) that orgName is customer facing, but same param also supports orgId also, so not documenting specifically

docs/authentication.md
anil-uipath: Changed the section to Coded function, so it supports whichever runtime the coded function supports

src/core/uipath.ts
anil-uipath: Good catch, added a comment. I framed it around *why* the re-read is deliberate, so nobody later tries to optimise it by caching the constructor's merge:

```
// Re-reads every source rather than reusing the constructor's merge: meta
// tags can be injected after construction, and recovering that is why this
// runs at all.
```

Also dropped the old `// Load from meta tags` line above it, which only restated the code.

src/core/config/environment.ts
anil-uipath: Addressed by refactoring to only keep UIPATH_ORG_NAME to stay consistent with customer facing naming

amrit-agarwal-1 (approved):

deepeshrai-tech (approved):

PR #680, opened 2026-08-19
feat(overrides): resolve resource overrides through an ambient channel

deepeshrai-tech:
## Summary

Adds runtime resolution of admin-configured resource overrides for `getByName` lookups, delivered through two sources with a shared shape:

- **Ambient channel** — a host runtime installs a source at `globalThis[Symbol.for('uipath.resourceOverwrites.v1')] = () => dict`. Read on every lookup, so per-request state on the host side (e.g. AsyncLocalStorage in a Node runtime, tenant-scoped tables in a pooled process) shows through correctly.
- **Page fallback** — when no host has installed itself, reads `<script type="application/json" id="uipath-overrides">` from the DOM. This is the existing coded-app FE deployment shape (script tag injected at deploy time by `syncResourceOverwritesToCdn`).

Wired centrally into [`FolderScopedService.getByNameLookup`](src/services/folder-scoped.ts) so every folder-scoped service (Assets, Buckets, Processes) inherits the swap without per-service changes.

## Key contract

Table shape: `Record<string, {name?: string, folderPath?: string, ...}>`
Keys: `type.name` (unscoped) or `type.name.folderPath` (scoped) — publisher's exact literal, case-sensitive.
Types onboarded: `Asset`, `Bucket`, `Process` (SDK label → publisher prefix mapped case-sensitively via `BINDING_TYPES`).

Lookup order:
1. Scoped `type.name.folderPath` if caller supplied a `folderPath`
2. Unscoped `type.name` fallback — for solution-inline bindings the admin scoped without a folder
3. Nothing → caller's original identity holds

## Files

- `src/utils/ambient/channel.ts` — 21-line `probeChannel` primitive (generic, reusable)
- `src/utils/overrides/resolve-override.ts` — resolver with scoped→unscoped fallback
- `src/utils/overrides/page-overrides.ts` — DOM `<script>` reader as a channel source
- `src/utils/overrides/overrides.types.ts` — shared types
- `src/utils/runtime/constants.ts` — new `UiPathElementIds.RESOURCE_OVERRIDES` enum
- `src/services/folder-scoped.ts` — call `resolveOverride` in `getByNameLookup`
- Unit tests for all of the above (+ folder-scoped integration test asserting the wire-in)

## Test plan

- [x] `npm run test:unit` — 2421 passed (1 pre-existing failure on `tests/unit/scripts/check-samples.test.ts` due to missing `node-html-parser` dep on main, unaffected by this change)
- [x] `npm run build` — all bundles + `.d.ts` emitted, exit 0
- [ ] Integration validation with a coded-functions host that installs a channel source

## Not in scope (follow-ups)

- Expanding `BINDING_TYPES` to Queue / Entity — separate PR
- Onboarding `Function` as an override-able resource type (requires OR-side `WellKnownBindingsProperties` + jamjam merged dict update)
- `AsyncLocalStorage` helper for host runtimes that want per-request scoping — hosts own that concern today

Co-Authored-By: Luca Rachiteanu <luca.rachiteanu@uipath.com>


src/utils/ambient/channel.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/** Binds a prober for one channel: the installed source when there is one, else `fallback`. */
+export function probeChannel<T>(key: symbol, fallback?: ChannelSource<T>): ChannelSource<T> {
+  return () => {
+    const source = currentSource<T>(key) ?? fallback;
+    if (!source) return undefined;
+    try {
+      return source();
+    } catch {
+      return undefined;
```
anil-uipath: How to identify/trace the failures. should we have telemetry?

lucarachiteanu (approved):

src/services/folder-scoped.ts
```diff
@@ -96,10 +97,19 @@ export class FolderScopedService extends BaseService {
     const validatedName = validateName(resourceType, name);
     const { folderId, folderKey, folderPath, ...queryOptions } = options;
 
+    // Always consulted, whatever folder context the caller supplied: an override redirects the
+    // design-time name an admin configured, and that holds however the call names its folder.
+    // `folderPath` also scopes the lookup, for publishers that key entries by design-time folder.
+    const override = resolveOverride(resourceType, validatedName, folderPath);
+    const resolvedName = override?.name ?? validatedName;
+    // The override's folder describes where its target lives, so it replaces the caller's path.
+    // `folderId`/`folderKey` still ride along; the server applies folderPath > folderKey > folderId.
```
Raina451: please add concise comments. and remove comments if not needed.

src/services/folder-scoped.ts
```diff
@@ -96,10 +97,19 @@ export class FolderScopedService extends BaseService {
...
+    const resolvedFolderPath = override?.folderPath ?? folderPath;
```
Raina451: If the caller passed folderId and the override supplies folderPath, we send both headers and lean on the server picking path first ?

src/utils/ambient/channel.ts
Raina451, line 1: lets not add a new folder `ambient/channel` in utils. use some exisitng folder and files for this.
also channel name isnt intutive, please check if we can add these in existing files

src/utils/overrides/resolve-override.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Matching is case-sensitive, so the publisher's exact literal is written here rather than derived
+ * from the label at runtime. A label absent from this map is a resource the publisher does not
+ * describe, and is never redirected.
+ */
+const BINDING_TYPES: Record<string, string> = {
+  Asset: 'asset',
+  Bucket: 'bucket',
+  Process: 'process',
+};
```
Raina451: why we need hardcoded list? can we have a better way here?

Raina451: The `asset.` part is already inside the key that comes from Orchestrator ( asset.xyz.Shared ), 2 things:

1. if bindings have:
{
  taggedAsset: { key: "asset.CustomerConfig.Shared", name: "CustomerConfig" }
};

// users pass bindings object, not just the name
assets.getByName(bindings.taggedAsset)

// sdk do:
const entry = overridesTable[binding.key];


or resourceType.toLowerCase can be used, please check once

Raina451 (changes requested):

src/services/folder-scoped.ts
deepeshrai-tech: When the caller passes folderId (or folderKey) and the override supplies a folderPath, both headers ship: `X-UIPATH-FolderPath-Encoded` from the override, and `X-UIPATH-OrganizationUnitId` (or folder-key header) from the caller. Orchestrator resolves in priority order `folderPath > folderKey > folderId`, so the override wins as intended. Documented in the comment on the affected line

src/utils/ambient/channel.ts
deepeshrai-tech: inlined the primitive into `resolve-override.ts` as a private `readOverrides()` helper, deleted the `utils/ambient/` folder.

src/utils/overrides/resolve-override.ts
deepeshrai-tech: replaced `BINDING_TYPES` with `resourceType.toLowerCase()`. Publisher writes lowercase prefixes (Orchestrator `ResourceTypeRaw`); SDK labels are PascalCase 

the resolver logic here works whether callers pass raw strings today or generated bindings.taggedAsset.name later.

src/services/folder-scoped.ts
deepeshrai-tech: Trimmed to single-line comments

tests/unit/utils/overrides/resolve-override.test.ts
```diff
@@ -0,0 +1,125 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import { resolveOverride } from '../../../../src/utils/overrides/resolve-override';
+import type { ResourceOverrides } from '../../../../src/utils/overrides/overrides.types';
+
+const KEY = Symbol.for('uipath.resourceOverwrites.v1');
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in d70a41e78 — imported `OVERRIDE_TEST_CONSTANTS.CHANNEL_KEY` and derive the symbol from it.

tests/unit/utils/overrides/page-overrides.test.ts
```diff
@@ -0,0 +1,105 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+
+const ELEMENT_ID = 'uipath-overrides';
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in d70a41e78 — imported `UiPathElementIds.RESOURCE_OVERRIDES` and reference it directly.

src/utils/overrides/resolve-override.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+export function resolveOverride(
+  resourceType: string,
+  name: string,
+  folderPath?: string,
+): ResourceOverride | undefined {
+  const table = readOverrides();
+  if (!table) return undefined;
```
Raina451, line 30: all these errors return undefined, can we log atleast?

Raina451 (approved):
minor comment, rest lgtm

src/utils/overrides/resolve-override.ts
deepeshrai-tech: Addressed in eea78a2e5 — added `console.warn` at both silent-catch points: the ambient-channel source throw in `resolve-override.ts`, and the DOM JSON parse in `page-overrides.ts`. Behavior stays "degrade to no overrides" (still safer than failing every `getByName`), but operators now get a diagnostic breadcrumb in browser console / Node logs.

PR #681, opened 2026-08-21
refactor(queues): fold WithMethods variants into getAll and getById [PLT-104203]

shivendra6720:
## Summary

Removed the `getAllWithMethods()` / `getByIdWithMethods()` method names (shipped only in 1.6.2):

- **`getAll`** — options now extend `FolderScopedOptions` (`folderId` / `folderKey` / `folderPath`; header-based), and it returns queues **with the operational methods attached** (`QueueGetWithMethodsResponse`). No overload — the released options already carried `folderId`, so an overload could only be selected by *requiring* `folderKey`/`folderPath`. 
- **`getById`** — overloaded: new `getById(id, { folderId | folderKey | folderPath })` returns the queue with methods; the released positional `getById(id, folderId)` is **retained but deprecated** and keeps returning plain data.
- Docs updated (oauth-scopes, pagination, error-handling); `release-metadata.json` untouched — the next version bump records the two removals automatically.


tests/unit/services/orchestrator/queues.test.ts
```diff
@@ -163,11 +163,11 @@ describe('QueueService Unit Tests', () => {
     });
   });
 
-  describe('getByIdWithMethods', () => {
+  describe('getById — options-object overload', () => {
```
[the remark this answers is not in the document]
swati354: @shivendra6720 check if valid

src/models/orchestrator/queues.models.ts
```diff
@@ -146,7 +102,25 @@ export interface QueueServiceModel {
...
+   * @param id - Queue ID
+   * @param folderId - Required folder ID
+   * @param options - Optional query options
+   * @returns Promise resolving to a {@link QueueGetResponse} — the queue definition
+   * @example
+   * ```typescript
+   * const queue = await queues.getById(<queueId>, <folderId>);
+   * ```
+   */
+  getById(id: number, folderId: number, options?: QueueGetByIdOptions): Promise<QueueGetResponse>;
```
swati354, line 123: for the other folder scope change PRs we kept it overloaded since we didn't want any breaking changes. Here since there anyways is a breaking change, would it make sense to just keep the new getById and remove this one? cc @deepeshrai-tech

shivendra6720: getAllWithMethods/getByIdWithMethods were public for exactly one release (since release 1.6.2 about a week ago). 
The positional getById(id, folderId) has shipped in every release since 1.0.0, so dropping it would hard-break all existing queue consumers with a compile error on upgrade and also following the same pattern as followed for the getFileMetaData for buckets service.

swati354 (approved):

deepeshrai-tech (approved):

PR #683, opened 2026-08-22
feat(refs): add resolveRefToId helper for the resource-resolution framework

deepeshrai-tech:
## Summary

Phase 1 of the [Resource Resolution Framework](https://uipath.atlassian.net/wiki/spaces/CLD/pages/90924515367/) — introduces the shared `resolveRefToId` helper that turns a `ResourceRef<TId>` into the canonical id the API expects.

The `ResourceRef<TId>` type already exists at [src/models/common/types.ts:57](src/models/common/types.ts#L57) and is used today by `Queues.startTransaction` (via `QueueRef`). This PR adds the missing resolver so subsequent service migrations share one call site.

## Design

```ts
export async function resolveRefToId<TId>(
  ref: ResourceRef<TId> | undefined,
  resolveByName: (name: string) => Promise<TId>,
  callerLabel: string,
): Promise<TId>
```

- **`{id}` branch**: returns id directly, does not invoke the lookup
- **`{name}` branch**: invokes the service-supplied `resolveByName` closure (which owns folder scoping and query shape)
- **Missing / neither branch**: throws `ValidationError` with the caller label so failures point at the offending method

The closure ownership is deliberate — folder-scoped services (Assets, Buckets, Queues, Processes) plumb folder options through their own closure; non-folder-scoped services (Data Fabric later) supply a simpler one. Framework stays orthogonal.

## No service changes in this PR

Per the phased migration cadence in the framework doc, this PR lands **the primitive alone**. Service signature changes ship in follow-ups:
- PR #2: Assets — `updateValueById` → `updateValue(ref, ...)`
- PR #3: Buckets — `uploadFile` / `getReadUri` / `deleteFile` / `getFileMetaData` / `getFiles` take `ref`
- PR #4: Queues — standardize `insertItemByName`, `getAllItems`, `completeTransaction` on ref
- PR #5: Processes — `start` takes `ProcessRef`-shaped input
- PR #6: Maestro — Cases, ProcessInstances (needs server-side name resolution for PIMs)

Data Fabric services (Entities, ChoiceSets) deferred — DF source not in this workspace to verify name-lookup support.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — 0 / 0
- [x] `npm run test:unit` — 2421 passed (1 pre-existing failure on `check-samples.test.ts` unrelated)
- [x] `npm run build` — exit 0
[...]

PR #684, opened 2026-08-22
feat(assets): add ref-based updateValue; deprecate updateValueById

deepeshrai-tech:
## Summary

Phase 2 of the [Resource Resolution Framework](https://uipath.atlassian.net/wiki/spaces/CLD/pages/90924515367/). **Stacked on #683** — merge that first.

Introduces `AssetService.updateValue(ref, newValue, options?)` where `ref: AssetRef = { id } | { name }`. Consolidates the operational value-update path behind a single ref-aware entry point per the framework.

## Diff shape

- **New**: `updateValue(ref, newValue, options?)` — CRUD-preserving; when \`ref\` is \`{ name }\`, the folder scoping in \`options\` drives the name-to-id lookup via \`getByNameLookup\` (protected base method, no extra telemetry event)
- **Deprecated**: \`updateValueById(id, newValue, options?)\` — kept for one minor version, delegates to a private helper. Same wire behaviour, doubled entry point until we cut in the next major
- **Types**: \`AssetRef = ResourceRef<number>\`, \`AssetUpdateValueOptions extends FolderScopedOptions\`
- **DRY**: both public methods call \`updateValueByResolvedId(id, newValue, headers)\` — one place owns the GET-then-PUT wire shape, no double \`@track\` per convention

## Test plan

- [x] \`npm run typecheck\` — clean
- [x] \`npm run lint\` — 0 / 0
- [x] \`npm run test:unit\` — 2426 passed (5 new for \`updateValue\`, all existing \`updateValueById\` tests still green — the alias works via the shared helper)
- [x] \`npm run build\` — exit 0

New tests cover:
- \`updateValue({ id })\` — id branch, no lookup fired
- \`updateValue({ name })\` — OData \$filter lookup with folder header, then id-based fetch and PUT
- \`updateValue({})\` — \`ValidationError\` before the API is touched
- \`updateValue(ref, undefined)\` — \`ValidationError\` before ref resolution
- folderId + folderKey propagation on the PUT

## Not in scope

- \`getByName\` already exists on \`AssetService\`; no addition here
- \`Credential\` / \`Secret\` value types stay unsupported (same as \`updateValueById\`)
- Deprecated method removal is a follow-up for the next major

### DIFF
--- a/docs/oauth-scopes.md
+++ b/docs/oauth-scopes.md
@@ -9,7 +9,9 @@ This page lists the specific OAuth scopes required in external app for each SDK 
 | `getAll()` | `OR.Assets` or `OR.Assets.Read` |
 | `getById()` | `OR.Assets` or `OR.Assets.Read` |
 | `getByName()` | `OR.Assets` or `OR.Assets.Read` |
-| `updateValueById()` | `OR.Assets` or `OR.Assets.Write` |
+| `getByKey()` | `OR.Assets` or `OR.Assets.Read` |
+| `updateValue()` | `OR.Assets` or `OR.Assets.Read`, `OR.Assets` or `OR.Assets.Write` |
+| `updateValueById()` | `OR.Assets` or `OR.Assets.Read`, `OR.Assets` or `OR.Assets.Write` |
 
 ## Jobs
 
--- a/src/models/orchestrator/assets.models.ts
+++ b/src/models/orchestrator/assets.models.ts
@@ -1,4 +1,4 @@
-import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions, AssetGetByNameOptions, AssetNewValue, AssetUpdateValueByIdOptions } from './assets.types';
+import { AssetGetAllOptions, AssetGetResponse, AssetGetByIdOptions, AssetGetByKeyOptions, AssetGetByNameOptions, AssetNewValue, AssetRef, AssetUpdateValueByIdOptions, AssetUpdateValueOptions } from './assets.types';
 import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '../../utils/pagination';
 
 /**
@@ -92,10 +92,35 @@ export interface AssetServiceModel {
   getByName(name: string, options?: AssetGetByNameOptions): Promise<AssetGetResponse>;
 
   /**
-   * Updates the value of an existing asset by ID.
+   * Retrieves a single asset by key (GUID).
+   *
+   * @param key - Asset key (GUID)
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) and optional query parameters (`expand`, `select`)
+   * @returns Promise resolving to a single {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // By folder ID
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderId: 123 });
+   *
+   * // By folder key (GUID)
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderKey: '<folderKey>' });
+   *
+   * // By folder path
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderPath: 'Shared/Finance' });
+   *
+   * // With expand
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderPath: 'Shared/Finance', expand: 'keyValueList' });
+   * ```
+   */
+  getByKey(key: string, options?: AssetGetByKeyOptions): Promise<AssetGetResponse>;
+
+  /**
+   * Updates the value of an existing asset, identified by `assetRef` (`{ id }`, `{ name }`, or `{ key }` (GUID)).
    *
    * Fetches the asset internally to determine its type, then updates only the value while
-   * preserving the asset's name, scope, and description.
+   * preserving the asset's name, scope, and description. Folder scoping in `options` drives
+   * both the ref's name/key lookup and the update itself — the same folder options apply to
+   * both calls.
    *
    * **Supported value types:** `Text`, `Integer`, and `Bool` only. Other types
    * (`Credential`, `Secret`) throw a `ValidationError`.
@@ -105,22 +130,34 @@ export interface AssetServiceModel {
    * - `Integer` → `number` (integer)
    * - `Bool` → `boolean`
    *
-   * @param id - Asset ID
-   * @param newValue - New value to apply (string for `Text`, number for `Integer`, boolean for `Bool`)
+   * @param assetRef - Asset ref (`{ id }`, `{ name }`, or `{ key }` (GUID))
+   * @param newValue - New value to apply
    * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`)
    * @returns Promise resolving when the asset has been updated
    *
    * @example
    * ```typescript
-   * // Update a Text asset by folder ID
-   * await assets.updateValueById(<assetId>, 'new-value', { folderId: <folderId> });
+   * // Update by id
+   * await assets.updateValue({ id: <assetId> }, 'new-value', { folderId: <folderId> });
    *
-   * // Update an Integer asset by folder key (GUID)
-   * await assets.updateValueById(<assetId>, 42, { folderKey: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' });
+   * // Update by name (folder options drive the name lookup)
+   * await assets.updateValue({ name: 'ApiKey' }, 42, { folderPath: 'Shared/Finance' });
    *
-   * // Update a Bool asset by folder path
-   * await assets.updateValueById(<assetId>, true, { folderPath: 'Shared/Finance' });
+   * // Update by GUID key
+   * await assets.updateValue({ key: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' }, true, { folderPath: 'Shared/Finance' });
    * ```
+   */
+  updateValue(assetRef: AssetRef, newValue: AssetNewValue, options?: AssetUpdateValueOptions): Promise<void>;
+
+  /**
+   * Updates the value of an existing asset by ID.
+   *
+   * @deprecated Use {@link AssetServiceModel.updateValue} with `{ id }`, `{ name }`, or `{ key }` instead. This
+   * method will be removed in the next major version.
+   *
+   * @param id - Asset ID
+   * @param newValue - New value to apply
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`)
    */
   updateValueById(id: number, newValue: AssetNewValue, options?: AssetUpdateValueByIdOptions): Promise<void>;
 }
--- a/src/models/orchestrator/assets.types.ts
+++ b/src/models/orchestrator/assets.types.ts
@@ -1,4 +1,4 @@
-import { BaseOptions, FolderScopedOptions, RequestOptions } from '../common/types';
+import { BaseOptions, FolderScopedOptions, FolderScopingOnly, RequestOptions, ResourceRef } from '../common/types';
 import { PaginationOptions } from '../../utils/pagination';
 
 /**
@@ -71,7 +71,13 @@ export interface AssetGetByIdOptions extends BaseOptions {}
 export interface AssetGetByNameOptions extends FolderScopedOptions {}
 
 /**
- * New value accepted by {@link AssetServiceModel.updateValueById}.
+ * Options for getting a single asset by key (GUID)
+ */
+export interface AssetGetByKeyOptions extends FolderScopedOptions {}
+
+/**
+ * New value for {@link AssetServiceModel.updateValue} (and the deprecated
+ * {@link AssetServiceModel.updateValueById}).
  *
  * The runtime type must match the asset's `valueType`:
  * - `Text` → `string`
@@ -82,5 +88,22 @@ export type AssetNewValue = string | number | boolean;
 
 /**
  * Options for updating an asset value by ID
+ * @deprecated Use {@link AssetUpdateValueOptions} with the ref-based `updateValue` method instead.
  */
 export interface AssetUpdateValueByIdOptions extends FolderScopedOptions {}
+
+/**
+ * Options for {@link AssetServiceModel.updateValue}. Folder scoping applies to both
+ * the ref's name/key lookup and the update — the same folder options drive both calls.
+ *
+ * Only folder-scoping fields are accepted; `expand` / `select` are intentionally absent because
+ * the internal lookup needs the full asset shape (name, valueType, valueScope, description) to
+ * round-trip the update body without a redundant follow-up GET.
+ */
+export interface AssetUpdateValueOptions extends FolderScopingOnly {}
+
+/**
+ * Selects an asset by exactly one identifier — `{ id }`, `{ name }`, or `{ key }` (GUID).
+ * See {@link ResourceRef}.
+ */
+export type AssetRef = ResourceRef<number>;
--- a/src/services/orchestrator/assets/assets.ts
+++ b/src/services/orchestrator/assets/assets.ts
@@ -1,6 +1,7 @@
 import { FolderScopedService } from '../../folder-scoped';
-import { AssetGetResponse, AssetGetAllOptions, AssetGetByIdOptions, AssetGetByNameOptions, AssetNewValue, AssetUpdateValueByIdOptions, AssetValueScope, AssetValueType } from '../../../models/orchestrator/assets.types';
+import { AssetGetResponse, AssetGetAllOptions, AssetGetByIdOptions, AssetGetByKeyOptions, AssetGetByNameOptions, AssetNewValue, AssetRef, AssetUpdateValueByIdOptions, AssetUpdateValueOptions, AssetValueScope, AssetValueType } from '../../../models/orchestrator/assets.types';
 import { AssetServiceModel } from '../../../models/orchestrator/assets.models';
+import { resolveRefToId } from '../../../utils/validation/resolve-ref';
 import { addPrefixToKeys, pascalToCamelCaseKeys, transformData, transformOptions } from '../../../utils/transform';
 import { createHeaders } from '../../../utils/http/headers';
 import { FOLDER_ID } from '../../../utils/constants/headers';
@@ -76,7 +77,7 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
 
   @track('Assets.GetByName')
   async getByName(name: string, options: AssetGetByNameOptions = {}): Promise<AssetGetResponse> {
-    return this.getByNameLookup<AssetGetResponse, AssetGetResponse>(
+    const { result } = await this.getByNameLookup<AssetGetResponse, AssetGetResponse>(
       'Asset',
       ASSET_ENDPOINTS.GET_BY_FOLDER,
       name,
@@ -84,6 +85,83 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
       (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
       AssetMap,
     );
+    return result;
+  }
+
+  @track('Assets.GetByKey')
+  async getByKey(key: string, options: AssetGetByKeyOptions = {}): Promise<AssetGetResponse> {
+    const { result } = await this.getByKeyLookup<Record<string, unknown>, AssetGetResponse>(
+      'Asset',
+      ASSET_ENDPOINTS.GET_BY_FOLDER,
+      key,
+      options,
+      (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
+      AssetMap,
+    );
+    return result;
+  }
+
+  @track('Assets.UpdateValue')
+  async updateValue(assetRef: AssetRef, newValue: AssetNewValue, options?: AssetUpdateValueOptions): Promise<void> {
+    if (newValue === null || newValue === undefined) {
+      throw new ValidationError({ message: 'newValue is required for updateValue' });
+    }
+
+    // Name/key lookups already return the full asset — stash it here so `updateValueByResolvedId`
+    // can skip the follow-up `getById` and go straight to the PUT.
+    let preFetched: Pick<AssetGetResponse, 'name' | 'valueScope' | 'valueType' | 'description'> | undefined;
+
+    const { id, effectiveFolder } = await resolveRefToId<number>(
+      assetRef,
+      {
+        byName: async (name) => {
+          const { result: asset, effectiveFolder: folder } = await this.getByNameLookup<Record<string, unknown>, AssetGetResponse>(
+            'Asset',
+            ASSET_ENDPOINTS.GET_BY_FOLDER,
+            name,
+            { folderId: options?.folderId, folderKey: options?.folderKey, folderPath: options?.folderPath },
+            (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
+            AssetMap,
+            'Assets.updateValue',
+          );
+          preFetched = asset;
+          return { id: asset.id, ...folder };
+        },
+        byKey: async (key) => {
+          const { result: asset, effectiveFolder: folder } = await this.getByKeyLookup<Record<string, unknown>, AssetGetResponse>(
+            'Asset',
+            ASSET_ENDPOINTS.GET_BY_FOLDER,
+            key,
+            { folderId: options?.folderId, folderKey: options?.folderKey, folderPath: options?.folderPath },
+            (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
+            AssetMap,
+            'Assets.updateValue',
+          );
+          preFetched = asset;
+          return { id: asset.id, ...folder };
+        },
+      },
+      'Assets.updateValue',
+    );
+
+    // `resolveRefToId` treats `id: 0` as a real value (generic over `TId`); assets have positive
+    // numeric ids only, so reject `0`/negatives here rather than letting them hit the API.
+    if (id <= 0) {
+      throw new ValidationError({ message: 'Assets.updateValue: assetRef.id must be a positive number.' });
+    }
+
+    // Prefer the folder the lookup confirmed against — for {name} refs an override redirect may
+    // have changed folderPath. Falls back to caller-supplied options only for the {id} branch,
+    // which leaves effectiveFolder empty (no lookup ran).
+    const headers = resolveFolderHeaders({
+      folderId: effectiveFolder.folderId ?? options?.folderId,
+      folderKey: effectiveFolder.folderKey ?? options?.folderKey,
+      folderPath: effectiveFolder.folderPath ?? options?.folderPath,
+      resourceType: 'Assets.updateValue',
+      fallbackFolderKey: this.config.folderKey,
+    });
+
+    await this.updateValueByResolvedId(id, newValue, headers, preFetched);
   }
 
   @track('Assets.UpdateValueById')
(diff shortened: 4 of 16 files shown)


src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +87,43 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+      'Assets.updateValue',
+    );
+
+    await this.updateValueByResolvedId(id, newValue, headers);
+  }
+
+  /**
+   * @deprecated Use {@link AssetService.updateValue} with `{ id }` or `{ name }` instead. This
+   * method will be removed in the next major version.
+   */
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8941bfac9 — removed the `@deprecated` JSDoc block from the class method. `AssetServiceModel.updateValueById` on the interface still carries the `@deprecated` tag, so IDE tooltips and TypeDoc pick it up through `implements`.

tests/unit/services/orchestrator/assets.test.ts
```diff
@@ -655,4 +655,102 @@ describe('AssetService Unit Tests', () => {
...
+        'v',
+        { folderId: TEST_CONSTANTS.FOLDER_ID, folderKey: 'k' },
+      );
+
+      const [, , putOpts] = mockApiClient.put.mock.calls[0];
+      expect(putOpts?.headers?.[FOLDER_ID]).toBe(TEST_CONSTANTS.FOLDER_ID.toString());
+      expect(putOpts?.headers?.[FOLDER_KEY]).toBe('k');
+    });
+  });
 });
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8941bfac9 — added a test in the `updateValue (ref-based)` block covering Credential + Secret rejection (both variants throw `ValidationError` before the PUT fires).

src/models/orchestrator/assets.models.ts
```diff
@@ -105,22 +123,34 @@ export interface AssetServiceModel {
    * - `Integer` → `number` (integer)
    * - `Bool` → `boolean`
    *
-   * @param id - Asset ID
-   * @param newValue - New value to apply (string for `Text`, number for `Integer`, boolean for `Bool`)
+   * @param ref - Asset ref (`{ id }` or `{ name }`)
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8941bfac9 — opening sentence and `@param ref` now include `{ key }`, and the `@example` block has a new "Update by GUID key" line.

src/models/orchestrator/assets.models.ts
```diff
@@ -92,10 +92,28 @@ export interface AssetServiceModel {
   getByName(name: string, options?: AssetGetByNameOptions): Promise<AssetGetResponse>;
 
   /**
-   * Updates the value of an existing asset by ID.
+   * Retrieves a single asset by key (GUID).
+   *
+   * @param key - Asset key (GUID)
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) and optional query parameters (`expand`, `select`)
+   * @returns Promise resolving to a single asset
+   * {@link AssetGetResponse}
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8941bfac9 — `@returns` on `getByKey` is now `Promise resolving to a single {@link AssetGetResponse}` inline.

tests/unit/services/orchestrator/assets.test.ts
```diff
@@ -463,6 +463,47 @@ describe('AssetService Unit Tests', () => {
...
+
+      expect(result.id).toBe(ASSET_TEST_CONSTANTS.ASSET_ID);
+      expect(result.name).toBe(ASSET_TEST_CONSTANTS.ASSET_NAME);
+
+      const [endpoint, opts] = mockApiClient.get.mock.calls[0];
+      expect(endpoint).toBe(ASSET_ENDPOINTS.GET_BY_FOLDER);
+      expect(opts?.params?.$filter).toBe(`Key eq ${TEST_KEY}`);
+      expect(opts?.params?.$top).toBe('1');
+      expect(opts?.headers?.[FOLDER_ID]).toBe(TEST_CONSTANTS.FOLDER_ID.toString());
+    });
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 5bc6a3f95 — added transform-completeness assertions to the `getByKey` success test: `result.createdTime` / `result.lastModifiedTime` present, `CreationTime` / `LastModificationTime` absent. Mirrors the existing `getByName` reference pattern.

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +96,58 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+    // confirmed by the successful lookup itself.
+    const headers = resolveFolderHeaders({
+      folderId: options?.folderId,
+      folderKey: options?.folderKey,
+      folderPath: options?.folderPath,
+      resourceType: 'Assets.updateValue',
+      fallbackFolderKey: this.config.folderKey,
+    });
+
+    await this.updateValueByResolvedId(id, newValue, headers);
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 05748c915 — `resolveValueField`\s default-case message now reads `Asset ${id} has valueType ${valueType}; only Text, Integer, and Bool are supported`, no longer names either public method. Correct for both callers of the shared helper.

src/models/orchestrator/assets.models.ts
```diff
@@ -92,10 +92,27 @@ export interface AssetServiceModel {
...
+   * @param key - Asset key (GUID)
+   * @param options - Folder scoping (`folderId` / `folderKey` / `folderPath`) and optional query parameters (`expand`, `select`)
+   * @returns Promise resolving to a single {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // By folder path
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderPath: 'Shared/Finance' });
+   * ```
+   */
+  getByKey(key: string, options?: AssetGetByKeyOptions): Promise<AssetGetResponse>;
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 1727a593c — added both `getByKey()` and `updateValue()` to the Assets section of `docs/oauth-scopes.md` with the read/write scopes matching the existing pattern.

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +96,58 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
     );
   }
 
+  @track('Assets.GetByKey')
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 1727a593c — added integration test blocks for both methods:
- `getByKey` — success case (retrieve by GUID key using folderKey, includes transform-completeness assertions) + NotFoundError for missing key
- `updateValue (ref-based)` — three success cases covering `{id}`, `{name}`, and `{key}` branches; each fetches an existing Text asset, updates via the new method, verifies the change with `getById`, and restores. Mirrors the existing `updateValueById` reference pattern.

src/models/orchestrator/assets.types.ts
```diff
@@ -82,5 +87,20 @@ export type AssetNewValue = string | number | boolean;
...
  */
 export interface AssetUpdateValueByIdOptions extends FolderScopedOptions {}
+
+/**
+ * Options for {@link AssetServiceModel.updateValue}. Folder scoping applies to the caller-supplied
+ * ref's name lookup; when a name lookup runs, the update itself targets the folder the resolved
+ * asset actually lives in (from the lookup response) rather than the caller-supplied folder,
+ * so runtime overrides that redirect across folders remain consistent.
+ */
+export interface AssetUpdateValueOptions extends FolderScopedOptions {}
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8ddf59584 — the `AssetUpdateValueOptions` JSDoc (and the same misleading claim on `AssetServiceModel.updateValue`) now describes the actual behaviour: caller's folder options drive both the ref lookup and the update. Override redirects are handled inside `getByNameLookup` → `resolveOverride`; the lookup's success confirms the target is in the redirected folder.

src/models/orchestrator/assets.models.ts
```diff
@@ -105,22 +122,34 @@ export interface AssetServiceModel {
...
+   * await assets.updateValue({ key: '5f6dadf1-3677-49dc-8aca-c2999dd4b3ba' }, true, { folderPath: 'Shared/Finance' });
    * ```
    */
+  updateValue(ref: AssetRef, newValue: AssetNewValue, options?: AssetUpdateValueOptions): Promise<void>;
+
+  /**
+   * Updates the value of an existing asset by ID.
+   *
+   * @deprecated Use {@link AssetServiceModel.updateValue} with `{ id }` or `{ name }` instead. This
+   * method will be removed in the next major version.
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8ddf59584 — `@deprecated` on `updateValueById` now lists all three variants: `{ id }`, `{ name }`, or `{ key }`.

tests/unit/services/orchestrator/assets.test.ts
```diff
@@ -655,4 +703,152 @@ describe('AssetService Unit Tests', () => {
...
+      mockApiClient.get.mockResolvedValue(mockExistingAsset({ ValueType: AssetValueType.Secret }));
+
+      await expect(
+        assetService.updateValue({ id: ASSET_TEST_CONSTANTS.ASSET_ID }, 'x', {
+          folderId: TEST_CONSTANTS.FOLDER_ID,
+        }),
+      ).rejects.toBeInstanceOf(ValidationError);
+
+      expect(mockApiClient.put).not.toHaveBeenCalled();
+    });
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 8ddf59584 — added both tests to the `updateValue (ref-based)` block:
- `preserves name, scope, type, and description when updating via {id} ref` — locks in that the PUT body still carries all preserved fields even if a future refactor bypasses `updateValueByResolvedId`.
- `throws ValidationError when no folder context is provided with {id} ref` — matches the sibling coverage on `updateValueById`.

src/services/orchestrator/assets/assets.ts
```diff
@@ -131,6 +206,50 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+      folderKey,
+      folderPath,
+      resourceType: callerLabel,
+      fallbackFolderKey: this.config.folderKey,
+    });
+
+    const apiFieldOptions = transformOptions(queryOptions, AssetMap);
+    const apiOptions = {
+      ...addPrefixToKeys(apiFieldOptions, ODATA_PREFIX, Object.keys(apiFieldOptions)),
+      '$filter': `Key eq ${trimmedKey.replace(SINGLE_QUOTE_RE, "''")}`,
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 368d059a0 — dropped the `.replace(SINGLE_QUOTE_RE, "")` call from the `$filter` and removed the now-unused `SINGLE_QUOTE_RE` constant. `GUID_REGEX` guarantees the key is hex+hyphens, so the escape was defending against something impossible.

src/models/orchestrator/assets.types.ts
```diff
@@ -70,6 +70,11 @@ export interface AssetGetByIdOptions extends BaseOptions {}
  */
 export interface AssetGetByNameOptions extends FolderScopedOptions {}
 
+/**
+ * Options for getting a single asset by key (GUID)
+ */
+export interface AssetGetByKeyOptions extends FolderScopedOptions {}
+
 /**
  * New value accepted by {@link AssetServiceModel.updateValueById}.
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 368d059a0 — `AssetNewValue`\s JSDoc now reads `New value for {@link AssetServiceModel.updateValue} (and the deprecated {@link AssetServiceModel.updateValueById}).` — accurate for both callers.

src/models/orchestrator/assets.types.ts
```diff
@@ -82,5 +88,21 @@ export type AssetNewValue = string | number | boolean;
...
 export interface AssetUpdateValueByIdOptions extends FolderScopedOptions {}
+
+/**
+ * Options for {@link AssetServiceModel.updateValue}. Folder scoping applies to both the ref's
+ * name/key lookup and the update — the same folder options drive both. When runtime overrides
+ * redirect a name across folders, the redirect happens inside the lookup path
+ * (`getByNameLookup` → `resolveOverride`), and the successful lookup confirms the target's
+ * presence in the redirected folder.
+ */
+export interface AssetUpdateValueOptions extends FolderScopedOptions {}
```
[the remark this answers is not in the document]
deepeshrai-tech: Already addressed — after your manual JSDoc trim (squashed into 7e20f874f), `AssetUpdateValueOptions` reads: `Options for {@link AssetServiceModel.updateValue}. Folder scoping applies to both the ref's name/key lookup and the update — the same folder options drive both calls.` No internal method names, matches your suggestion exactly.

tests/integration/shared/orchestrator/assets.integration.test.ts
```diff
@@ -195,6 +195,143 @@ describe.each(modes)('Orchestrator Assets - Integration Tests [%s]', (mode) => {
     });
   });
 
+  describe('getByKey', () => {
+    it('should retrieve an asset by its GUID key using folderKey', async () => {
+      const { assets } = getServices();
+      const config = getTestConfig();
+
+      expect(config.folderKey, 'INTEGRATION_TEST_FOLDER_KEY must be configured for getByKey').toBeDefined();
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in 7e20f874f — both occurrences in the `getByKey` block now use `if (!config.folderKey) { throw new Error(...) }`, matching the sibling `updateValue` tests below. Also preemptively swept for hardcoded values: replaced two hardcoded `5f6dadf1-...` GUIDs with `ASSET_TEST_CONSTANTS.ASSET_KEY`.

src/models/orchestrator/assets.models.ts
```diff
@@ -92,10 +92,26 @@ export interface AssetServiceModel {
...
+   * @returns Promise resolving to a single {@link AssetGetResponse}
+   * @example
+   * ```typescript
+   * // By folder path
+   * await assets.getByKey('5f6dadf1-3677-49dc-8aca-c2999dd4b3ba', { folderPath: 'Shared/Finance' });
+   * ```
+   */
+  getByKey(key: string, options?: AssetGetByKeyOptions): Promise<AssetGetResponse>;
+
+  /**
```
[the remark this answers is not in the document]
deepeshrai-tech: Addressed in e07233e09 — expanded `getByKey`\s `@example` block to match the `getByName` reference: folderId, folderKey, folderPath, plus a `expand` example. Bare minimum call first, variants after.

deepeshrai-tech:
Addressed the two remaining findings in 36869d491:

**1. `updateValue (ref-based)` — extract shared `getAll` to `beforeAll`.** All three `it` blocks in the describe now share one `getAll` call + target selection from a `beforeAll`. Each test only runs its own `updateValue` + verification + restore. Saves 2 `getAll` API calls per suite run.

**2. `getByKey` — consolidate service availability guard.** Both `it` blocks used to repeat `if (!config.folderKey) throw new Error(...)`. Now consolidated in a `beforeAll` that asserts folder config + fetches the target asset once, exposing `folderKey` and `existing` as `let` variables the tests reuse.

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +93,58 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+    if (newValue === null || newValue === undefined) {
+      throw new ValidationError({ message: 'newValue is required for updateValue' });
+    }
+
+    // Resolve the ref via the shared framework helper. Name-lookup goes through the protected
+    // `getByNameLookup` (which applies runtime resource overrides); key-lookup goes through the
+    // private `findAssetByKey` helper. Both return `{ id }` — Assets' OData response does not
+    // carry the folder id, so `effectiveFolder.folderId` stays undefined and the update falls
+    // back to the caller's folder options.
+    const { id } = await resolveRefToId<number>(
```
swati354: nit : https://github.com/UiPath/uipath-typescript/blob/4f331f200693e2c586df26cb99dd3fff19072983/src/utils/validation/resolve-ref.ts#L71 
`updateValue({ id: 0 })` hits the API

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +93,58 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+        byName: async (name) => {
+          const asset = await this.getByNameLookup<AssetGetResponse, AssetGetResponse>(
+            'Asset',
+            ASSET_ENDPOINTS.GET_BY_FOLDER,
+            name,
+            options ?? {},
+            (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
+            AssetMap,
+          );
+          return { id: asset.id };
```
swati354: `byName/byKey` resolvers already hold the full asset (Name, ValueScope, ValueType, Description), but only `asset.id` is threaded through, why?
for `byName` and `byKey` we are making 2 get calls(first `getByNameLookup/ findAssetByKey` then `getById` in `updateValueByResolvedId`), even though both return the same result?

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +93,58 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+    // Assets' name/key lookups do not return the folder id, so the caller's folder options
+    // remain authoritative for the update. Runtime override redirects that cross folders are
+    // confirmed by the successful lookup itself.
+    const headers = resolveFolderHeaders({
+      folderId: options?.folderId,
+      folderKey: options?.folderKey,
+      folderPath: options?.folderPath,
+      resourceType: 'Assets.updateValue',
+      fallbackFolderKey: this.config.folderKey,
+    });
```
swati354, line 162: we apply `resolveOverride` inside `getByNameLookup`, but in the PUT call later we use the user's original folder headers. What if a binding redirects the name lookup to a different folder, the PUT will hit the wrong folder?
won't the successful lookup only confirm that the asset exists in the overridden folder?

src/services/orchestrator/assets/assets.ts
```diff
@@ -13,7 +14,13 @@ import { PaginatedResponse, NonPaginatedResponse, HasPaginationOptions } from '.
...
 import { PaginationType } from '../../../utils/pagination/internal-types';
 import { track } from '../../../core/telemetry';
-import { ValidationError } from '../../../core/errors';
+import { CollectionResponse } from '../../../models/common/types';
+import { NotFoundError, ValidationError } from '../../../core/errors';
+
+/**
+ * Matches a canonical GUID; used to reject non-GUID keys on `getByKey` before hitting the API.
+ */
+const GUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
swati354: this and `findAssetByKey` structure are already present in [queues.ts ](https://github.com/UiPath/uipath-typescript/blob/4f331f200693e2c586df26cb99dd3fff19072983/src/services/orchestrator/queues/queues.ts#L181)
should we have a `getByKeyLookup`, similar to how we have a centralised `getByNameLookup`?
(can take this in a separate PR as well)

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: added an explicit check

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: The byName/byKey resolvers now stash the fetched asset in a closure variable, and `updateValueByResolvedId` accepts an optional `preFetched arg. When supplied (name/key branches), the follow-up `getById` is skipped and we go straight to the PUT.

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: done changes , extracted so that on method call we always have resolved values

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: okay

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: Actually addressed in 3844d2dd4 — did the fix Deepesh suggested: apply `resolveOverride` upfront in `updateValue` for `{name}` refs, so the resolved name AND folderPath flow into both the lookup and the follow-up PUT header.

```ts
if (ref && "name" in ref && ref.name) {
  const override = resolveOverride("Asset", ref.name, options?.folderPath);
  if (override?.name) refForLookup = { name: override.name };
  if (override?.folderPath) resolvedFolderPath = override.folderPath;
}
// resolvedFolderPath now drives both the getByNameLookup call and the PUT header below.
```

No changes to `AssetGetResponse` or `getByNameLookup` — the resolver is idempotent so the (redundant) internal call inside `getByNameLookup` no-ops on the already-resolved name.

New unit test asserts: with a cross-folder override installed (`Shared/Apps → Prod/Live`), both the lookup GET and the update PUT ship `X-UIPATH-FolderPath-Encoded` scoped to `Prod/Live`. No more mismatch.

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +94,77 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+      refForLookup,
+      {
+        byName: async (name) => {
+          const asset = await this.getByNameLookup<AssetGetResponse, AssetGetResponse>(
+            'Asset',
+            ASSET_ENDPOINTS.GET_BY_FOLDER,
+            name,
+            { ...options, folderPath: resolvedFolderPath },
+            (raw) => transformData(pascalToCamelCaseKeys(raw), AssetMap),
+            AssetMap,
```
swati354, line 124: `byName` and `byKey` are passing options (unlike `byId` which simple returns id). These options extend `BaseOptions` and so also contain `select`. If somebody then passed something like 
`await assets.updateValue({ name: 'ApiKey' }, 'v', { folderPath: 'X', select: 'id,name' });`

The lookup returns only id and name, `preFetched.valueType` is undefined, `resolveValueField` hits the default: branch and throws `"Asset x has valueType undefined; only Text, Integer, and Bool are supported".`

src/services/orchestrator/assets/assets.ts
```diff
@@ -86,6 +94,77 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+      const override = resolveOverride('Asset', ref.name, options?.folderPath);
+      if (override?.name) refForLookup = { name: override.name };
+      if (override?.folderPath) resolvedFolderPath = override.folderPath;
+    }
+
+    // Name/key lookups already return the full asset — stash it here so `updateValueByResolvedId`
+    // can skip the follow-up `getById` and go straight to the PUT.
+    let preFetched: Pick<AssetGetResponse, 'name' | 'valueScope' | 'valueType' | 'description'> | undefined;
+
+    const { id } = await resolveRefToId<number>(
```
swati354: `resolveRefToId` returns `{ id, effectiveFolder }` so name/key lookups can carry the resolved folder up to the follow-up mutation. Here, `effectiveFolder` is dropped on both branches, both return only `{ id }`.
And so the upper block duplicates work that `getByNameLookup` already does, the only reason it's here is because resolveRefToId's `effectiveFolder` return value is being dropped.
We could extend `EffectiveFolder` to `{ folderId?, folderPath? }` and have `getByNameLookup` return `{ result, effectiveFolder }`
Could be a bigger change since `getByNameLookup` has many callers but I think otherwise every service which will implement ref based methods will have to do duplicate this.

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: Fixed

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: fixed

swati354 (approved):

src/services/orchestrator/assets/assets.ts
```diff
@@ -76,14 +83,78 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+
+    // Name/key lookups already return the full asset — stash it here so `updateValueByResolvedId`
+    // can skip the follow-up `getById` and go straight to the PUT.
+    let preFetched: Pick<AssetGetResponse, 'name' | 'valueScope' | 'valueType' | 'description'> | undefined;
+
+    const { id, effectiveFolder } = await resolveRefToId<number>(
+      assetRef,
+      {
+        byName: async (name) => {
+          const { result: asset, effectiveFolder: folder } = await this.getByNameLookup<AssetGetResponse, AssetGetResponse>(
```
ninja-shreyash: this will flag wrong error message when folder options are not provided 

"Asset.getByName requires folder context"

src/services/orchestrator/assets/assets.ts
```diff
@@ -112,24 +228,56 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+      ...addPrefixToKeys(apiFieldOptions, ODATA_PREFIX, Object.keys(apiFieldOptions)),
+      '$filter': `Key eq ${trimmedKey}`,
+      '$top': '1',
     };
 
-    await this.put(
-      ASSET_ENDPOINTS.GET_BY_ID(id),
-      body,
-      { headers },
+    const response = await this.get<CollectionResponse<AssetGetResponse>>(
```
ninja-shreyash: should it accept the raw response as type instead of AssetGetResponse (AssetGetResponse is post conversion right?)

src/services/orchestrator/assets/assets.ts
```diff
@@ -112,24 +228,56 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+    };
+  }
 
-    const valueField = resolveValueField(id, existing.ValueType, newValue);
+  /**
+   * Looks up a single asset by its GUID key on the folder-scoped OData collection. Shared by
+   * public `getByKey` and the byKey branch of `updateValue` — no `@track` here so calling from
+   * within another `@track`-decorated method does not double-fire telemetry.
+   */
+  private async findAssetByKey(
```
ninja-shreyash: can we define a common helper getByKeyLookup just like getByNameLookup and use it?

src/services/orchestrator/assets/assets.ts
ninja-shreyash: same regex is used in queues and a few other places. consider extracting and reusing this constant.

src/models/orchestrator/assets.types.ts
```diff
@@ -82,5 +88,29 @@ export type AssetNewValue = string | number | boolean;
...
+
+/**
+ * Options for {@link AssetServiceModel.updateValue}. Folder scoping applies to both
+ * the ref's name/key lookup and the update — the same folder options drive both calls.
+ *
+ * Only folder-scoping fields are accepted; `expand` / `select` are intentionally absent because
+ * the internal lookup needs the full asset shape (name, valueType, valueScope, description) to
+ * round-trip the update body without a redundant follow-up GET.
+ */
+export interface AssetUpdateValueOptions {
```
ninja-shreyash: Should we omit 'expand' | 'select' on folderScopedOptions itself?

docs/oauth-scopes.md
```diff
@@ -9,6 +9,8 @@ This page lists the specific OAuth scopes required in external app for each SDK
 | `getAll()` | `OR.Assets` or `OR.Assets.Read` |
 | `getById()` | `OR.Assets` or `OR.Assets.Read` |
 | `getByName()` | `OR.Assets` or `OR.Assets.Read` |
+| `getByKey()` | `OR.Assets` or `OR.Assets.Read` |
+| `updateValue()` | `OR.Assets` or `OR.Assets.Write` |
```
ninja-shreyash: just write won't be enough because it makes a get call first

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: fixed

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: fixed

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: Done

src/services/orchestrator/assets/assets.ts
deepeshrai-tech: done

src/models/orchestrator/assets.types.ts
deepeshrai-tech: fixed by adding a interface just for folderScopingOnly

docs/oauth-scopes.md
deepeshrai-tech: fixed

src/services/orchestrator/assets/assets.ts
```diff
@@ -76,14 +77,91 @@ export class AssetService extends FolderScopedService implements AssetServiceMod
...
+
+    // `resolveRefToId` treats `id: 0` as a real value (generic over `TId`); assets have positive
+    // numeric ids only, so reject `0`/negatives here rather than letting them hit the API.
+    if (id <= 0) {
+      throw new ValidationError({ message: 'Assets.updateValue: assetRef.id must be a positive number.' });
+    }
+
+    // Prefer the folder the lookup actually confirmed against (may be an override redirect for a
+    // {name} ref). Falls back to caller-supplied options for the {id}/{key} branches, which do not
+    // populate `effectiveFolder`.
```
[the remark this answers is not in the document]
deepeshrai-tech: Right — after the getByKeyLookup extraction, the byKey branch also populates `effectiveFolder` (just echoes the caller's options back). Comment updated verbatim to your suggestion in dcaa18527.

ninja-shreyash (approved):

PR #695, opened 2026-08-28
feat: add httpRequest and wait runtime helpers [APPS-35694]

anil-uipath:
## Summary

Adds two runtime helpers to the SDK root export:

- **`httpRequest(url, init?)`** — a `fetch` convenience for calling any URL, with configurable retry, backoff, and per-attempt timeout.
- **`wait(durationMs)`** — pause for a duration.

Both are standalone functions, not services: they take no `UiPath` instance, send no UiPath authentication, and add no UiPath headers.

## Motivation

Code that runs as a coded function can only reach this SDK. Calling a third-party API from there previously meant hand-rolling `fetch` plus a retry loop in every project. `httpRequest` makes that a supported, tested part of the SDK, with the robustness controls — retry count, backoff shape, delay caps, timeouts — expressed as options rather than reimplemented each time.

The design is intentionally general rather than shaped around any single caller: no product-specific limits, no product-specific field names, and the same options object is intended to work on SDK service methods in a follow-up.

## Behaviour

Two contracts differ from the SDK's service methods, both because this helper proxies arbitrary third-party calls:

**A status the server returned is never an exception.** A 404 or 500 resolves with `ok: false`, so callers can read the response body — which on an error is usually where the vendor's diagnostic lives. Only a request that never produced a response throws, as a `NetworkError`. (An explicit `responseType: 'json'` with an unparseable body raises a `ServerError`.)

**No authentication or tracing headers are attached.** The target is a third-party host, so forwarding `traceparent` would leak internal correlation IDs outside UiPath.

```ts
import { httpRequest, wait } from '@uipath/uipath-typescript';

const response = await httpRequest('https://api.example.com/v1/orders', {
  method: 'POST',
  headers: { 'x-api-key': '<apiKey>' },
  body: { sku: 'ABC-123' },
  timeoutMs: 10_000,
  retry: { maxRetries: 3, initialDelayMs: 1000, backoffStrategy: 'exponential' },
});

if (response.ok) {
  const order = response.data as { id: string };
  console.log(order.id);
} else {
  console.log('failed with', response.status);
}

await wait(1000);
```

The response is a snapshot — `status`, `statusText`, `ok`, `headers`, `data`, `url` — so the body is already read. `responseType` selects `json` / `text` / `blob` / `arraybuffer` / `stream`; without it, a JSON content type is parsed and anything else is returned as text.

### `data` is `unknown`

The body is whatever the server sent: the expected shape on a success, an error payload on a 4xx or 5xx, and nothing at all on a 204. A generic `data: T` would be accurate on only one of those paths, and the error path is the one this helper exists to make readable. Callers assert or validate the shape after checking `ok`.

### Retry

`retry` accepts `maxRetries`, `initialDelayMs`, `backoffStrategy` (`'constant' | 'linear' | 'exponential'`), `backoffFactor`, `backoffMaxDelayMs`, `retryableStatusCodes`, `retryMethods`, `retryNetworkErrors`, `respectRetryAfter`, and `maxRetryAfterMs`.

Defaults follow established HTTP client behaviour:

- Retries the **idempotent methods of RFC 9110** — `GET`, `HEAD`, `PUT`, `DELETE`, `OPTIONS`. `POST` and `PATCH` are excluded unless opted into `retryMethods`, since a replayed `POST` can create the same resource twice.
- Retries transient statuses (`408`, `429`, `500`, `502`, `503`, `504`) and transport failures; `retryNetworkErrors: false` limits it to statuses only.
- Honours a `Retry-After` header **as sent**, rather than capping it with the backoff limit. `maxRetryAfterMs` bounds it separately and is unbounded by default.
- Never retries a `ReadableStream` body — it is consumed by the first attempt and cannot be replayed — or a request cancelled through `signal`.

`timeoutMs` sits on the request rather than inside `retry`, because it bounds a single attempt whether or not retrying is enabled. Each retry starts a fresh timeout.

## Effect on existing SDK calls

None. The retry engine is shared with `ApiClient` through `RequestSpec`, so retrying can be extended to service methods later, but its defaults there set `maxRetries: 0`. Every existing call keeps its single-attempt behaviour, and a test asserts exactly one `fetch` on a 500 when no options are supplied.

Service methods keep their own response types and continue to throw on non-2xx.

`RequestSpec` also gains a working `timeoutMs`, replacing a `timeoutOptions` field that was declared but never read — so service requests can carry a deadline for the first time.

## Backward compatibility

`RetryOptions`, `TimeoutOptions`, `RequestSpec.retryOptions` and `RequestSpec.timeoutOptions` are all publicly exported and shipped in 1.6.2, so they continue to work:

- `retryDelay` and `useExponentialBackoff` remain on `RetryOptions` as `@deprecated`, mapping onto `initialDelayMs` and `backoffStrategy` respectively. Each applies only when its replacement is absent.
- `retryOptions` and `timeoutOptions` remain on `RequestSpec` as `@deprecated`, honoured only when `retry` / `timeoutMs` are not supplied.
- `abortOnTimeout` gets no mapping and is documented as having no effect: with `fetch`, a timeout is an abort.

This keeps the change a patch release under the release policy, which requires a deprecation cycle before removal.

## Testing

`typecheck`, `lint`, unit tests (2558), `build`, and `docs:validate` all pass.

- `tests/unit/utils/http/resilience.test.ts` — backoff per strategy, delay caps, `Retry-After` parsing, defaults resolution, deprecated-field mapping and precedence
- `tests/unit/utils/http/http-request.test.ts` — non-2xx resolution, retry eligibility per method, stream bodies, network-error toggle, timeouts, cancellation, body parsing, query parameters
- `tests/unit/utils/wait.test.ts`
- `tests/unit/core/http/api-client.test.ts` — the retry-off-by-default contract, per-request opt-in, per-request timeout, deprecated-field fallbacks
- `tests/integration/shared/http/http-request.integration.test.ts` — live endpoints, including an unauthorised response arriving as data rather than an exception

## Notes

- No `docs/oauth-scopes.md` entry — the helper is unauthenticated.
- No `mkdocs.yml` nav entry or subpath export — it is not a service.
- `jitter` is not included. Most reference implementations offer it but default it off; it can be added later as an optional field without a breaking change.


docs/error-handling.md
```diff
@@ -251,4 +251,114 @@ try {
...
+## The `httpRequest` Helper
+
+`httpRequest` is a general-purpose HTTP utility for calling any URL. It carries no UiPath
+authentication, so it follows a different error contract from the SDK's service methods.
+
+**A status the server returned is never an exception.** A 404 or a 500 comes back as a resolved
+response with `ok: false` — branch on the status rather than catching:
+
+```typescript
+import { httpRequest } from '@uipath/uipath-typescript';
```
Raina451: I believe this should be imported from core module. 
"@uipath/uipath-typescript/core" instead the whole package.
cc: @vnaren23 thoughts?

docs/error-handling.md
```diff
@@ -251,4 +251,114 @@ try {
     console.log('Debug info:', JSON.stringify(debugInfo, null, 2));
   }
 }
-```
\ No newline at end of file
+```
+## The `httpRequest` Helper
+
```
Raina451: why do we need to add these helpers in error handling doc?
these would end up here: https://uipath.github.io/uipath-typescript/error-handling/

src/models/common/http.types.ts
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * How the delay grows between attempts, given an `initialDelayMs` of `d` and a
+   * `backoffFactor` of `f`:
+   *
+   * - `constant` — `d, d, d, …`
+   * - `linear` — `d, 2d, 3d, …`
+   * - `exponential` — `d, d×f, d×f², …`
+   *
+   * @default 'exponential'
+   */
```
Raina451: please use concise comments

src/models/common/http.types.ts
```diff
@@ -0,0 +1,187 @@
...
+   * Upper bound for a delay taken from a `Retry-After` header, in milliseconds. Unbounded by
+   * default, so a server's explicit instruction is honoured as sent.
+   * @default Infinity
+   */
+  maxRetryAfterMs?: number;
+
+  /**
+   * Base delay between retries in milliseconds.
+   *
+   * @deprecated Use {@link RetryOptions.initialDelayMs}. Still honoured when
```
Raina451, line 72: why deprecated? nothing has been released.

src/models/common/http.types.ts
```diff
@@ -0,0 +1,187 @@
...
+  signal?: AbortSignal;
+}
+
+/**
+ * The outcome of an `httpRequest` call. A response is returned for every status the server
+ * produced — including 4xx and 5xx — so callers branch on `ok` or `status` rather than catching.
+ */
+export interface HttpResponse {
+  /** Response status code. */
+  status: number;
```
Raina451, line 128: we dont need type? https://developer.mozilla.org/en-US/docs/Web/API/Response/type

src/models/common/request-spec.ts
```diff
@@ -90,10 +80,29 @@ export interface RequestSpec {
...
   responseType?: ResponseType;
   
-  /** Request timeout options */
+  /** Timeout for a single attempt, in milliseconds. Unbounded unless supplied. */
+  timeoutMs?: number;
+
+  /**
+   * Request timeout options.
+   *
+   * @deprecated Use {@link RequestSpec.timeoutMs}. `timeoutOptions.timeout` is still honoured
```
Raina451, line 89: why depricated?

src/utils/http/http-request.ts
```diff
@@ -0,0 +1,176 @@
...
+      statusCode: response.status,
+    });
+  }
+}
+
+/**
+ * Sends an HTTP request to any URL, with optional retries and exponential backoff.
+ *
+ * This helper carries no UiPath authentication and adds no UiPath headers — it is a plain
+ * `fetch` convenience for arbitrary endpoints. Pass `retry` to control retry behavior; by
```
Raina451: lets use concise comments everywhere in code comments

src/utils/timing.ts
Raina451, line 1: i dont think we need new file for this

src/models/common/http.types.ts
anil-uipath: The fields `retryDelay` and `useExponentialBackoff` are already in public interface `request-spec.ts` from the init commit. hence deprecating it

https://github.com/UiPath/uipath-typescript/blob/f1d6839c9cfe1724ba4f86fadc2286f8c0e8d93d/src/models/common/request-spec.ts#L26-L31

src/models/common/request-spec.ts
anil-uipath: Same as the `RetryOptions` thread — `TimeoutOptions` and `RequestSpec.timeoutOptions` are both in the `1.6.2` tag and publicly exported, so removing them is a breaking change for anyone referencing the names.

`TimeoutOptions` it had 2 fields
- `abortOnTimeout` flag is unimplementable with fetch. because for fetch, abort automatically happens on timeout. so removed it
- `timeout`- have implemented this duration field now.

And since it ends up with only 1 field `timeout`, simplified it for user from the previous

```
await httpRequest(url, {
  body: {},
  headers: {},
  timeoutOptions: { 
     timeout: 1000
  }
})
```
to 
```
await httpRequest(url, {
  body: {},
  headers: {},
  timeoutMs: 1000
})
```

docs/error-handling.md
anil-uipath: Agreed — moved to `docs/helpers.md` ("Helper Methods"), with a nav entry.

docs/error-handling.md
anil-uipath: Done. `httpRequest` and `wait` are now exported from `/core` too, along with their types, so the example imports cleanly from there. They were root-only before, which is why that import would not have resolved.

src/models/common/http.types.ts
anil-uipath: Done — trimmed across `http.types.ts`, `http-request.ts`, `retry-policy.ts` and `fetch-with-retry.ts`.

src/utils/http/http-request.ts
anil-uipath: Done — trimmed across `http.types.ts`, `http-request.ts`, `retry-policy.ts` and `fetch-with-retry.ts`.

src/models/common/http.types.ts
anil-uipath: They did ship — at tag `1.6.2`, `request-spec.ts` has `RetryOptions` (26), `TimeoutOptions` (40), `RequestSpec.timeoutOptions` (94) and `RequestSpec.retryOptions` (97), all exported publicly via `models/common` → `src/index.ts`.

They were never wired to anything, so they did nothing at runtime, but the names are importable — removing them breaks anyone referencing the types. The release policy also blocks it both ways: a patch must stay backward compatible, and a minor requires a deprecation cycle these never had.

If you know there are no external consumers yet, dropping them outright is cleaner and I will do that — just say.

src/models/common/http.types.ts
anil-uipath: Left out on purpose. Three of `type`s five values are unreachable here — `opaque` needs `mode: 'no-cors'`, `opaqueredirect` needs `redirect: 'manual'` (we expose neither), and `error` only comes from a constructed `Response.error()`, never a resolved fetch. That leaves `basic`/`cors`, which is constant in Node.

Added `redirected` instead — it is the only tell when an API silently redirects to an HTML login page and returns 200.

src/utils/timing.ts
anil-uipath: Fair. Would `src/utils/timing.ts` work instead? A file named after its only export looks unjustified, but one named after a category does not, and it gives the next small time helper somewhere to land.

I would avoid folding it into `http-request.ts` — `wait` is public and is not HTTP, and `fetch-with-retry.ts` imports it, so that would create a cycle.

src/utils/wait.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Waits for a fixed duration before resolving.
+ *
+ * @param durationMs - How long to wait, in milliseconds. Values below zero resolve immediately.
```
divyaprakash-dp: nit: it won't resolve immediately right even if 0, should we mention next cycle?

anil-uipath: Good catch. since it is an edge case, i will remove it from the user facing doc. I see API workflow also did not document this part.

src/models/common/http.types.ts
```diff
@@ -0,0 +1,150 @@
...
+  /** Cancels the request, including a retry that is waiting. */
+  signal?: AbortSignal;
+}
+
+/**
+ * What `httpRequest` returns. Every status the server sent comes back here, including 4xx and
+ * 5xx, so check `ok` or `status` instead of catching.
+ */
+export interface HttpResponse {
+  /** Response status code. */
```
Raina451, line 127: Why are we re-specifying the whole response shape instead of returning a Response? ts already knows this from @types/node

anil-uipath: Returning `Response` and returning `data` are the same decision — a body can only be read once, so you get one or the other. Libraries that parse for you (axios, got, superagent) all return their own shape; ky returns a `Response` precisely because it does not parse.

We parse mainly so a 4xx body is there without the caller having to `await res.json()` on what might be an HTML error page. Using the whole response shape, means that dropping `data` and `responseType` too.  not just swapping the return type.

Raina451 (approved):

deepeshrai-tech (approved):

amrit-agarwal-1 (approved):

PR #697, opened 2026-08-31
docs: fix sidebar nav collapse behaviour and group the per-builder guides

amrit-agarwal-1:
Two changes to the docs sidebar.

### 1. Preserve collapse state across navigation

Collapsible sections re-expanded on every page load, so any section a reader collapsed sprang back open as soon as they navigated. `navigation.expand` renders every section with Material's `md-toggle--indeterminate` class, which the theme treats as expanded; the server re-emits it on each page load, discarding the reader's choice.

Removing the flag, plus a small script that remembers expand/collapse choices in `sessionStorage`, with the current page's ancestors pinned open so you can't hide where you are. State is fingerprinted so a nav-tree change drops stale entries, and all storage access is wrapped in `try/catch`.

### 2. Group the per-builder guides

The AI App Builders section lists one page per builder, so it grows every time we add one. `navigation.sections` prevents collapsing a top-level section, and removing it is a site-wide change we don't want — but it only applies at the top level, so one level down groups collapse normally (`Coded Apps → CI/CD` already does).

```yaml
- AI App Builders:
    - Getting Started: ai-app-builders/getting-started.md
    - Guides:
        - Vercel (v0): ai-app-builders/vercel.md
        - Replit: ai-app-builders/replit.md
        - Bolt: ai-app-builders/bolt.md
        - Lovable: ai-app-builders/lovable.md
```

The section stays three rows however many builders we add, and the entry points stay visible.

### Verification

Built and loaded in a browser: `Guides` renders as a collapsible node, not force-expanded, and the persistence script ships. The second change depends on the first — with `navigation.expand` still on, the new node would render permanently open.

Docs-site only — no SDK source, no public API surface.

### DIFF
--- a/docs/overrides/assets/javascripts/extra.js
+++ b/docs/overrides/assets/javascripts/extra.js
@@ -240,8 +240,10 @@ function readNavState() {
         if (!saved || !Array.isArray(saved.open)) return null;
         if (saved.sig !== navSignature()) return null; // nav tree changed
         return new Set(saved.open);
-    } catch {
-        return null; // storage blocked (private mode) or corrupt — use defaults
+    } catch (error) {
+        // Storage blocked (private mode) or corrupt — fall back to defaults.
+        console.warn(error);
+        return null;
     }
 }
 
@@ -261,8 +263,9 @@ function updateNavState(toggle) {
             NAV_STATE_KEY,
             JSON.stringify({ sig: navSignature(), open: [...open] })
         );
-    } catch {
-        // Storage unavailable or full — silently fall back to default behaviour.
+    } catch (error) {
+        // Storage unavailable or full — fall back to default behaviour.
+        console.warn(error);
     }
 }
 
--- a/docs/overrides/assets/javascripts/extra.js
+++ b/docs/overrides/assets/javascripts/extra.js
@@ -197,3 +197,130 @@ function extractMarkdownContent() {
     
     return markdown;
 }
+/* ------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


docs/overrides/assets/javascripts/extra.js
```diff
@@ -196,4 +196,131 @@ function extractMarkdownContent() {
     });
     
     return markdown;
-}
\ No newline at end of file
+}
+/* ------------------------------------------------------------------
+ * Sidebar nav collapse-state persistence
+ *
```
Raina451, line 202: isnt anything in mkdocs is available for this?
can you check this also once if it helps: https://github.com/squidfunk/mkdocs-material/discussions/2173#discussioncomment-3281832

Raina451: i see their docs itself have this: 

Image

so something should be available ootb

amrit-agarwal-1: The collapse behaviour is OOTB. Dropping `navigation.expand` gives collapsible sections where only the active path opens, same as Material's own docs. No code needed for that.

What's not OOTB is remembering a section you opened as you move between pages. squidfunk confirms it in that thread: the nav always expands the active path and collapses the rest, and that holds with instant loading too since the whole nav gets replaced. I tried it on their docs, expanded a non-active section, clicked through to another page, and it was collapsed again. That's the only thing extra.js does.

Raina451 (approved):

Sarath1018 (approved):

PR #700, opened 2026-08-31
feat(auth): add enforceSso option to sign in via the org's identity provider

Raina451:
Adds an opt-in `enforceSso` flag that sends `acr_values=tenant:<orgId>` on the authorize request, so users land on the org's IdP instead of the UiPath account screen. Off by default; requires `orgName` to be the org id.

Context: https://uipath-product.slack.com/archives/C0175DZP4PQ/p1782397128159069?thread_ts=1777038498.265349&cid=C0175DZP4PQ -> acr_values was removed there because it blocks basic-auth users in mixed-auth orgs, so this stays opt-in.

### DIFF
--- a/src/core/auth/service.ts
+++ b/src/core/auth/service.ts
@@ -7,6 +7,8 @@ import { hasOAuthConfig } from '../config/sdk-config';
 import { isBrowser } from '../../utils/platform';
 import { IDENTITY_ENDPOINTS } from '../../utils/constants/endpoints';
 
+const GUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
 export class AuthService {
   private config: Config;
   private tokenManager: TokenManager;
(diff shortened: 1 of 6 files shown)


src/core/uipath.ts
```diff
@@ -142,6 +142,7 @@ export class UiPath implements IUiPath {
       clientId: hasOAuthAuth ? config.clientId : undefined,
       redirectUri: hasOAuthAuth ? config.redirectUri : undefined,
       scope: hasOAuthAuth ? config.scope : undefined,
+      forceSso: config.forceSso,
```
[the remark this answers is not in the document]
Raina451: not needed for this

vnaren23 (commented):
Can we improve naming of `forceSso`?

Raina451:
> Can we improve naming of `forceSso`?

I couldnt think of a better clear name. few others we can consider:
- useOrgIdentityProvider
- ssoOnly 
- directSso

please suggest any other naming you can think of

src/core/config/sdk-config.ts
```diff
@@ -12,15 +12,20 @@ export interface OAuthFields {
   scope: string;
 }
 
+// Opt-in sign-in behaviour.
+export interface SsoFields {
```
vnaren23: LoginFields maybe?

vnaren23 (commented):
Rest looks good, update PR description as well

vnaren23 (approved):

src/core/auth/service.ts
```diff
@@ -7,6 +7,8 @@ import { hasOAuthConfig } from '../config/sdk-config';
 import { isBrowser } from '../../utils/platform';
 import { IDENTITY_ENDPOINTS } from '../../utils/constants/endpoints';
 
+const GUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
deepeshrai-tech, line 10: this needs to be here or if we already have guid regex anywhere in code base we can reuse it?

deepeshrai-tech:
minor comment rest LGTM

deepeshrai-tech (approved):

src/core/auth/service.ts
Raina451: there's no existing GUID regex to use

PR #703, opened 2026-08-31
fix(data-fabric): correct the MULTILINE_MAX read contract in the SDK docs and restore list coverage

aayushuipath:
## What

The SDK documented `MULTILINE_MAX` list and query reads as always returning a size marker (`"HasValue=true Length=512"`). That is one of several shapes the backend can return, so the docs promised a guarantee the platform does not make. This corrects the read contract across the whole read surface, adds the two facts a caller hits first (the tenant feature gate and the 400s on filter/sort/join), and restores the list assertions that were dropped from the `MULTILINE_MAX` integration test in #635.

Backend behavior, verified against `SelectQueryBuilder.cs` and the V1/V2/V3 controllers:

| Read | Returns |
|---|---|
| list / query, value <= 10,000 chars | the value verbatim (flag on) |
| list / query, value > 10,000 chars | first 10,000 chars + `...[Truncated]` (flag on) |
| list / query, flag off | `HasValue=true Length=N`, plus a trailing hint on the list route |
| list / query, encrypted field | `HasValue=true Encrypted=true`, never content, either flag state |
| single-record read by ID | the whole value |
| null value, join query, insert/update response | key absent from the row |

Which preview a tenant returns depends on `enable-multiline-max-truncated-content`: on for alpha, off for staging and both gov rings. `enable-multiline-text-max`, which allows the field type at all, is off on prod, agov and pgov, so the type cannot currently be created there.

## How

- `entities.types.ts`: the `MULTILINE_MAX` enum member is now the canonical contract (byte budget and tenant gate, the preview shapes and the instruction to treat a preview as opaque, the key-absent rule and the 400s). `EntityJoin` distinguishes the hard 400 on a join condition from the silently omitted key on `selectedFields`.
- `entities.models.ts`: all four public record-reading methods now carry the preview note and link to the enum (`getAllRecords`, `queryRecordsById` and the bound `EntityModel` twins), `getRecord` gains the full-value note its `getRecordById` twin already had, and both update methods carry the read-modify-write warning inline.
- `data-fabric.ts`: endpoint comment no longer says "size marker".
- Integration test: body raised from 276 to 12,000 characters so a truncating server has to shorten it, and the restored assertions pin the preview to one of the two known shapes rather than merely "different from the stored value", which an empty string also satisfied.

## Security

No authentication, authorization, tenant-isolation or secret-handling surface. Two changes reduce risk: the docs now state that encrypted `MULTILINE_MAX` never returns content on a list or query, and both update methods warn that echoing a preview back silently overwrites the stored value.

Zero executable production lines changed. Filtering the `src/` diff for any non-comment, non-blank line returns nothing.

## Tests

- `npm run build`: exit 0, warnings identical to a clean `main` baseline.
- `tsc --noEmit`: clean. `oxlint`: 0 warnings, 0 errors.
- `npm run test:unit`: 2528 passed, 93 files.
- `typedoc --options typedoc.validation.json`: 0 errors (2 pre-existing unrelated warnings about duplicate `metaData` comments).
- **Not run here:** the `MULTILINE_MAX` integration test. It needs a PAT carrying `DataFabric.Schema.Write` and it creates entities on the target tenant. It should be run once before merge, ideally on a tenant with `enable-multiline-max-truncated-content` on and one with it off, since the assertions are written to hold in both states.

## Review triage

Three adversarial review rounds by an independent reviewer given the diff and the backend source, not the author's rationale.

Round 1 raised 1 P1 and 6 P2/P3. Round 2 raised 2 P2 and 4 P3 (one a regression from the round 1 fix). Round 3 returned clean with an explicit safe-to-ship verdict.

**P1, fixed:** the first version of this change said list and query return a preview "never the full content". False. The truncated projection's inner branch is the bare column, so a value at or under 10,000 characters is returned complete. The words "never the full content" and "lazy-loaded" are gone.

**P2, all fixed:**
- "Encrypted fields always return `HasValue=true Encrypted=true`" was wrong on flag-off tenants because the list route appends a hint. Reworded to a prefix.
- The round 1 fix then over-corrected and made encrypted look as though a short value could arrive verbatim. It cannot. Encrypted is now its own sentence.
- The byte-exact `Length=512` example was accurate only for `queryRecordsById`; `getAllRecords` uses `Marker` mode and appends the hint. The marker is now described as a prefix.
- Half the public read surface had no note: the bound `EntityModel.getAllRecords`, `queryRecords` and `getRecord`. Added.
- Insert and update responses drop `MULTILINE_MAX` entirely, previously undocumented anywhere. Stated in the enum, and inline on the two update methods where the round-trip hazard actually lives. Insert methods intentionally left to the enum: there is no round-trip hazard when supplying new content.
- "Comes back empty" for join queries was imprecise. The key is removed by `ResponseNullRemover`, so consumers see `undefined`. Reworded.

**P3, fixed:** filter, sort and join-condition 400s documented; the test comment no longer restates the false invariant; test assertions tightened and the `typeof` check kept so a broken contract reports what it received; the 128 KB byte budget and the 10,000 character limit separated so the units cannot be conflated.

**P3, accepted as-is:** "sometimes with a trailing hint" attaches grammatically to the `Length=N` marker while the encrypted marker can also carry it on flag-off tenants. The only behavior this could mislead is an equality comparison against the quoted string, which the preceding sentence explicitly forbids. A further caveat would add words for no behavioral gain.

## Follow-ups, not in this PR

- `records list` / `records query` guidance in the `uipath-platform` skill and its `integration_multiline_max` eval encode the same stale marker contract, and the eval's judge rubric currently scores a correct truncation explanation as wrong.
- Join queries project `NULL` for `MULTILINE_MAX` because the original marker was a dot-aliased column pair that could not coexist with the multi-entity path's underscore-flattened aliases. The preview is now a single scalar expression, so that constraint no longer applies and the null could be replaced. Deliberately left alone: bounding join response size is arguably the more valuable property, and nobody has asked for the content.

### DIFF
--- a/src/models/data-fabric/entities.models.ts
+++ b/src/models/data-fabric/entities.models.ts
@@ -147,8 +147,9 @@ export interface EntityServiceModel {
   /**
    * Gets entity records by entity ID
    *
-   * `MULTILINE_MAX` fields are returned as a size marker (e.g. `"HasValue=true Length=512"`)
-   * instead of the full content — use {@link getRecordById} to retrieve the full value.
+   * `MULTILINE_MAX` fields are returned as a bounded preview, not necessarily the whole
+   * value (see {@link EntityFieldDataType.MULTILINE_MAX}); use {@link getRecordById} for
+   * the guaranteed full value.
    *
    * @param entityId - UUID of the entity
    * @param options - Query options. The `folderKey` property is **experimental**.
@@ -191,8 +192,9 @@ export interface EntityServiceModel {
    * Sibling of {@link getAllRecords} that addresses the entity by name — useful when you only
    * have the resource name (e.g. solution binding overrides), avoiding a lookup to resolve the ID.
    *
-   * `MULTILINE_MAX` fields are returned as a size marker (e.g. `"HasValue=true Length=512"`)
-   * instead of the full content — use {@link getRecordByName} to retrieve the full value.
+   * `MULTILINE_MAX` fields are returned as a bounded preview, not necessarily the whole
+   * value (see {@link EntityFieldDataType.MULTILINE_MAX}); use {@link getRecordByName} for
+   * the guaranteed full value.
    *
    * @param entityName - Name of the entity
    * @param options - Query options. The `folderKey` property is **experimental**.
@@ -412,6 +414,11 @@ export interface EntityServiceModel {
   /**
    * Updates a single record in an entity, identified by ref (`{ id }` or `{ name }`)
    *
+   * Omit `MULTILINE_MAX` keys unless you intend to replace their content: a record echoed back
+   * from a list or query carries only a preview, and writing that overwrites the stored value
+   * (see {@link EntityFieldDataType.MULTILINE_MAX}). Those fields are not returned in the
+   * response either.
+   *
    * Note: Data Fabric supports trigger events only on individual updates, not on updating multiple records.
    * Use this method if you need trigger events to fire for the updated record.
    *
@@ -436,6 +443,11 @@ export interface EntityServiceModel {
 
   /**
    * Updates a single record in an entity by entity ID
+   *
+   * Omit `MULTILINE_MAX` keys unless you intend to replace their content: a record echoed back
+   * from a list or query carries only a preview, and writing that overwrites the stored value
+   * (see {@link EntityFieldDataType.MULTILINE_MAX}). Those fields are not returned in the
+   * response either.
    *
    * @deprecated Use {@link updateRecord} with `{ id }` or `{ name }` instead. This method will be removed in a future major version.
    *
@@ -468,6 +480,11 @@ export interface EntityServiceModel {
   /**
    * Updates data in an entity, identified by ref (`{ id }` or `{ name }`)
    *
+   * Omit `MULTILINE_MAX` keys unless you intend to replace their content: a record echoed back
+   * from a list or query carries only a preview, and writing that overwrites the stored value
+   * (see {@link EntityFieldDataType.MULTILINE_MAX}). Those fields are not returned in the
+   * response either.
+   *
    * Note: Records updated using updateRecords will not trigger Data Fabric trigger events. Use {@link updateRecord} if you need trigger events to fire for each updated record.
    *
    * @param entityRef - Entity ref (`{ id }` (GUID) or `{ name }`)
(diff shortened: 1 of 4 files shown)


AditiGoyalUipath (approved):

src/models/data-fabric/entities.types.ts
```diff
@@ -21,10 +21,22 @@ export enum EntityFieldDataType {
   BIG_INTEGER = "BIG_INTEGER",
   MULTILINE_TEXT = "MULTILINE_TEXT",
   /**
-   * Large multi-line text (up to 128 KB). Unlike {@link MULTILINE_TEXT}, the full
-   * value is lazy-loaded: list/query operations return a size marker
-   * (e.g. `"HasValue=true Length=512"`) instead of the content; reading a single
-   * record by ID returns the full value.
+   * Large multi-line text, up to a 128 KB byte budget (about 65,536 characters). Creating the
+   * field requires the Multi-line (Max) feature to be enabled for the tenant; `create` and
+   * add-field otherwise reject it with 400.
```
ninja-shreyash: "`create` and add-field otherwise reject it with 400"
No need to expose these internal details to sdk users

src/models/data-fabric/entities.models.ts
```diff
@@ -337,6 +343,11 @@ export interface EntityServiceModel {
   /**
    * Updates data in an entity by entity ID
    *
+   * Omit `MULTILINE_MAX` keys unless you intend to replace their content: a record echoed back
```
ninja-shreyash, line 346: add this on bound methods as well

src/models/data-fabric/entities.types.ts
aayushuipath: Split this one, since two things were bundled in that clause.

Dropped the parts you flagged: "add-field" is not an SDK method (the surface is `updateById({ addFields })`) and the raw status code is not what a caller handles when the SDK wraps errors into typed ones. Also removed the matching "rejects those with 400" from the filter/sort/join sentence so the block stays at one altitude.

Kept the fact itself, deliberately: `enable-multiline-text-max` is off on prod and both gov rings, so a production caller who picks `MULTILINE_MAX` out of the SDK's own public enum gets a rejection with no explanation anywhere. That is caller-visible behavior rather than an internal detail. It now reads: the Multi-line (Max) feature must be enabled for the tenant, otherwise creating the field or adding it to an existing entity is rejected.

Worth flagging if the wider principle is that public JSDoc should not name status codes: this file already does it above my change. `EntityHavingFilter` (lines 286-287 on main) names the raw flag key `enable-having-on-query` and says "the server responds 400". Mine was the milder version, naming the user-facing feature rather than the flag key. Happy to go either way, but that reads like a repo-wide convention call rather than something this PR should settle on its own, so I left the neighbouring text alone.

src/models/data-fabric/entities.models.ts
aayushuipath: Fxed. The bound `EntityModel.updateRecord` and `updateRecords` now carry the same note as their service-level counterparts.

Same gap had been caught earlier in this PR on the read side, where I added the note to the bound `getAllRecords`, `queryRecords` and `getRecord`, and then did not carry it across to the write side.

ninja-shreyash (approved):

PR #705, opened 2026-09-01
feat(attachments): add JobAttachmentSchema input type [APPS-37347]

lucarachiteanu:
## Added

| Layer | Member | Signature |
| Type | `JobAttachmentSchema` | `interface JobAttachmentSchema { ID: string; FullName: string; MimeType: string; Metadata?: Record<string, string> }` |

Declaring a `JobAttachmentSchema` field in a coded function's input contract is what makes Studio Web / Orchestrator render a file picker for it (schema emission is the coded-functions-js side, tracked separately under HDENS-7683); this type is the value the platform then delivers. To fetch the file it points at, pass its ID to the existing `getById()`.

No service changes: `getById(id: string, options?)` is unchanged from `main`, and no new method is added (per @vnaren23's review — the `getFor()` variant from an earlier revision was removed in d7a5721).

Field names are intentionally PascalCase: they mirror the platform's `job-attachment` schema verbatim (`ID`/`FullName`/`MimeType`/`Metadata`), the same contract the Python SDK's `uipath.platform.attachments.Attachment` models. The payload is injected into the function's input JSON by the job executor and never passes through the SDK's transform layer, so the type must match the wire exactly — see the casing thread in the review.

## Endpoint Called

None — this PR adds a type only. No new endpoint, no new OAuth scope.

## Example Usage

```typescript
import { Attachments } from '@uipath/uipath-typescript/attachments';
import type { JobAttachmentSchema } from '@uipath/uipath-typescript/attachments';

const attachments = new Attachments(sdk);

// Inside a coded function whose input declares a JobAttachmentSchema field
const invoice: JobAttachmentSchema = input.invoice;
const attachment = await attachments.getById(invoice.ID);
```

## API Response vs SDK Response

Not applicable — no service method is added or changed.

## Sample SDK Response

Not applicable.

## Files

| Area | Files |
|------|-------|
| Types | `src/models/orchestrator/attachments.types.ts` (`JobAttachmentSchema`) |

Notes for reviewers:
- `JobAttachment` (the start-job link reference in `processes.types.ts`) is a different contract — a link instruction you send when starting a job — and is released API, hence the new name.
- No version bump in this PR (per repo convention, bumps ship separately).

Refs APPS-37347

### DIFF
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -1,10 +1,8 @@
 import {
   AttachmentResponse,
   AttachmentGetByIdOptions,
-  AttachmentGetForOptions,
   AttachmentCreateOptions,
   AttachmentCreateResponse,
-  JobAttachmentSchema,
 } from './attachments.types';
 
 /**
@@ -29,24 +27,6 @@ export interface AttachmentServiceModel {
    * ```
    */
   getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
-
-  /**
-   * Gets the attachment a job attachment input refers to.
-   *
-   * @param attachment - The job attachment input received from the automation
-   * @param options - Optional query parameters (expand, select)
-   * @returns Promise resolving to the attachment, including `blobFileAccess` for downloading its content — {@link AttachmentResponse}
-   * @example
-   * ```typescript
-   * import { Attachments } from '@uipath/uipath-typescript/attachments';
-   *
-   * const attachments = new Attachments(sdk);
-   *
-   * // Inside a coded function whose input declares a JobAttachmentSchema field
-   * const attachment = await attachments.getFor(input.invoice);
-   * ```
-   */
-  getFor(attachment: JobAttachmentSchema, options?: AttachmentGetForOptions): Promise<AttachmentResponse>;
 
   /**
    * Creates an attachment and uploads its content.
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -3,6 +3,7 @@ import {
   AttachmentGetByIdOptions,
   AttachmentCreateOptions,
   AttachmentCreateResponse,
+  JobAttachmentSchema,
 } from './attachments.types';
 
 /**
@@ -14,10 +15,12 @@ export interface AttachmentServiceModel {
   /**
    * Gets an attachment by ID
    *
-   * @param id - The UUID of the attachment to retrieve
+   * Also accepts a {@link JobAttachmentSchema} received as automation input, so a
+   * coded function can pass its input value straight through to fetch the file.
+   *
+   * @param id - The UUID of the attachment to retrieve, or a job attachment received as automation input
    * @param options - Optional query parameters (expand, select)
-   * @returns Promise resolving to the attachment
-   * {@link AttachmentResponse}
+   * @returns Promise resolving to the attachment, including `blobFileAccess` for downloading its content. {@link AttachmentResponse}
    * @example
    * ```typescript
    * import { Attachments } from '@uipath/uipath-typescript/attachments';
@@ -25,8 +28,13 @@ export interface AttachmentServiceModel {
    * const attachments = new Attachments(sdk);
    * const attachment = await attachments.getById('12345678-1234-1234-1234-123456789abc');
    * ```
+   * @example
+   * ```typescript
+   * // Inside a coded function whose input declares a JobAttachmentSchema field
+   * const attachment = await attachments.getById(input.invoice);
+   * ```
    */
-  getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
+  getById(id: string | JobAttachmentSchema, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
 
   /**
    * Creates an attachment and uploads its content.
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -35,10 +35,13 @@ export interface AttachmentServiceModel {
    *
    * @param attachment - The job attachment input received from the automation
    * @param options - Optional query parameters (expand, select)
-   * @returns Promise resolving to the attachment
-   * {@link AttachmentResponse}
+   * @returns Promise resolving to the attachment, including `blobFileAccess` for downloading its content — {@link AttachmentResponse}
    * @example
    * ```typescript
+   * import { Attachments } from '@uipath/uipath-typescript/attachments';
+   *
+   * const attachments = new Attachments(sdk);
+   *
    * // Inside a coded function whose input declares a JobAttachmentSchema field
    * const attachment = await attachments.getFor(input.invoice);
    * ```
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -3,7 +3,6 @@ import {
   AttachmentGetByIdOptions,
   AttachmentCreateOptions,
   AttachmentCreateResponse,
-  JobAttachmentSchema,
 } from './attachments.types';
 
 /**
@@ -15,12 +14,10 @@ export interface AttachmentServiceModel {
   /**
    * Gets an attachment by ID
    *
-   * Also accepts a {@link JobAttachmentSchema} received as automation input, so a
-   * coded function can pass its input value straight through to fetch the file.
-   *
-   * @param id - The UUID of the attachment to retrieve, or a job attachment received as automation input
+   * @param id - The UUID of the attachment to retrieve
    * @param options - Optional query parameters (expand, select)
-   * @returns Promise resolving to the attachment, including `blobFileAccess` for downloading its content. {@link AttachmentResponse}
+   * @returns Promise resolving to the attachment
+   * {@link AttachmentResponse}
    * @example
    * ```typescript
    * import { Attachments } from '@uipath/uipath-typescript/attachments';
@@ -28,13 +25,8 @@ export interface AttachmentServiceModel {
    * const attachments = new Attachments(sdk);
    * const attachment = await attachments.getById('12345678-1234-1234-1234-123456789abc');
    * ```
-   * @example
-   * ```typescript
-   * // Inside a coded function whose input declares a JobAttachmentSchema field
-   * const attachment = await attachments.getById(input.invoice);
-   * ```
    */
-  getById(id: string | JobAttachmentSchema, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
+  getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
 
   /**
    * Creates an attachment and uploads its content.
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -1,8 +1,10 @@
 import {
   AttachmentResponse,
   AttachmentGetByIdOptions,
+  AttachmentGetForOptions,
   AttachmentCreateOptions,
   AttachmentCreateResponse,
+  JobAttachmentSchema,
 } from './attachments.types';
 
 /**
@@ -27,6 +29,21 @@ export interface AttachmentServiceModel {
    * ```
    */
   getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
+
+  /**
+   * Gets the attachment a job attachment input refers to.
+   *
+   * @param attachment - A {@link JobAttachmentSchema} received as automation input
+   * @param options - Optional query parameters (expand, select)
+   * @returns Promise resolving to the attachment
+   * {@link AttachmentResponse}
+   * @example
+   * ```typescript
+   * // Inside a coded function whose input declares a JobAttachmentSchema field
+   * const attachment = await attachments.getFor(input.invoice);
+   * ```
+   */
+  getFor(attachment: JobAttachmentSchema, options?: AttachmentGetForOptions): Promise<AttachmentResponse>;
 
   /**
    * Creates an attachment and uploads its content.
--- a/src/models/orchestrator/attachments.models.ts
+++ b/src/models/orchestrator/attachments.models.ts
@@ -33,7 +33,7 @@ export interface AttachmentServiceModel {
   /**
    * Gets the attachment a job attachment input refers to.
    *
-   * @param attachment - A {@link JobAttachmentSchema} received as automation input
+   * @param attachment - The job attachment input received from the automation
    * @param options - Optional query parameters (expand, select)
    * @returns Promise resolving to the attachment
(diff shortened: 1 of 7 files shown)


src/models/orchestrator/attachments.models.ts
```diff
@@ -14,19 +15,26 @@ export interface AttachmentServiceModel {
...
    * const attachment = await attachments.getById('12345678-1234-1234-1234-123456789abc');
    * ```
+   * @example
+   * ```typescript
+   * // Inside a coded function whose input declares a JobAttachmentSchema field
+   * const attachment = await attachments.getById(input.invoice);
+   * ```
    */
-  getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
+  getById(id: string | JobAttachmentSchema, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
```
amrit-agarwal-1: Is the name JobAttachmentSchema load-bearing for the schema emission in coded-functions-js? If yes, please note that in the JSDoc so nobody renames it later and quietly kills the file picker.

src/models/orchestrator/attachments.types.ts
```diff
@@ -48,6 +48,40 @@ export interface AttachmentResponse {
...
+ * Declare a field of this type in a coded function's input contract to receive
+ * a file at runtime — Studio Web and Orchestrator render a file picker for it
+ * and pass the selected file's metadata in this shape. The type carries
+ * metadata only; fetch the content with `attachments.getById()`.
+ *
+ * Field names are intentionally PascalCase: they mirror the platform's
+ * `job-attachment` schema verbatim, so the received value matches this type
+ * with no transformation.
+ */
+export interface JobAttachmentSchema {
```
amrit-agarwal-1, line 62: The Python model for this same contract has "title": "UiPathAttachment" in its schema config. Which name does the emission side actually key off? If it's this identifier, can we say so in the JSDoc so a future rename doesn't quietly break the file picker?

src/models/orchestrator/attachments.models.ts
lucarachiteanu: No — the coded-functions-js extractor matches this type structurally (field names + types), not by name. Renaming the interface is safe; changing the field shape is what would break the picker. Added a JSDoc line stating the field shape is the contract.

src/models/orchestrator/attachments.types.ts
lucarachiteanu: Unlike Python (which keys off the pydantic title \UiPathAttachment\), the JS emission side deliberately keys off no identifier — it matches the field shape structurally (ID/FullName/MimeType as strings, no extra props). Noted in the JSDoc that the field shape is the contract.

amrit-agarwal-1 (approved):

src/models/orchestrator/attachments.models.ts
Raina451: JobAttachmentSchema should be in options, and since id is always required, keep id: string only id out of schema and move JobAttachmentSchema to options

src/models/orchestrator/attachments.types.ts
```diff
@@ -48,6 +48,42 @@ export interface AttachmentResponse {
...
+  /**
+   * UUID of the attachment
+   */
+  ID: string;
+
+  /**
+   * Name of the attachment file, including its extension
+   */
+  FullName: string;
+
```
Raina451, line 72: Use transformData to convert them to camelCase as per convention. This was standardized across all methods, including those where the server-side response uses PascalCase.

src/models/orchestrator/attachments.types.ts
```diff
@@ -48,6 +48,42 @@ export interface AttachmentResponse {
...
+ *
+ * Declare a field of this type in a coded function's input contract to receive
+ * a file at runtime — Studio Web and Orchestrator render a file picker for it
+ * and pass the selected file's metadata in this shape. The type carries
+ * metadata only; fetch the content with `attachments.getById()`.
+ *
+ * Field names are intentionally PascalCase: they mirror the platform's
+ * `job-attachment` schema verbatim, so the received value matches this type
+ * with no transformation. The field shape is the contract — platform tooling
+ * recognizes attachment inputs structurally by these fields, so their names
```
Raina451: use concise comments as these eventually lands in public facing docs.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -28,8 +29,9 @@ import { BucketMap } from '../../../models/orchestrator/buckets.constants';
 
 export class AttachmentService extends BaseService implements AttachmentServiceModel {
   @track('Attachments.GetById')
-  async getById(id: string, options: AttachmentGetByIdOptions = {}): Promise<AttachmentResponse> {
-    if (!id) {
+  async getById(id: string | JobAttachmentSchema, options: AttachmentGetByIdOptions = {}): Promise<AttachmentResponse> {
```
anil-uipath: with getById as name, accepting both types doesn't seem right. 
should we explore adding a separate method
```
getFor(attachment: JobAttachmentSchema)
```

lucarachiteanu: Seems like the more intuitive alternative. Will do.

src/models/orchestrator/attachments.models.ts
lucarachiteanu: Done — reverted the widening in dc882dd. `getById(id: string, options?)` is unchanged from `main`; callers pass `input.invoice.ID`. The PR is now type-only.

src/models/orchestrator/attachments.types.ts
lucarachiteanu: This one I'd like to keep PascalCase — it isn't an SDK response, so there's nothing to transform.

The value is injected into a coded function's input JSON by the job executor and handed straight to the handler; it never enters the SDK. After the revert above, `getById` takes only a string, so the SDK truly never touches an instance of this type. There is no pipeline stage where `pascalToCamelCaseKeys` could run (and per `agent_docs/conventions.md:93`, `transformData` is for semantic renames only — "Never use for case conversion").

Declaring it camelCase would make the type describe something that doesn't exist at runtime, and it fails loudly: the coded-functions runtime validates the injected payload with ajv against the authored contract, so `{ID, FullName, MimeType}` against a `required: ["id"]` schema is a validation error, and `input.invoice.id` would be `undefined` in the handler even if the fields were optional. Making camelCase work needs a paired value/schema rename layer on both the input and output boundaries in the runtime, plus divergence from the Python SDK, whose `uipath.platform.attachments.Attachment` is PascalCase for the same reason.

Closest precedents in the conventions for casing-as-contract: the Data Fabric exception (`:119`) and user-defined nested payloads (`:121`) — both leave keys exactly as the wire sends them. The JSDoc now states this in one line so nobody "fixes" it later.

src/models/orchestrator/attachments.types.ts
lucarachiteanu: Trimmed in dc882dd — the doc block is down to three short paragraphs (what it is, how to use it, why the casing is fixed), and the extra `getById` JSDoc went away with the signature revert.

src/services/orchestrator/attachments/attachments.ts
lucarachiteanu: Added in 6f9b5e7 — `getFor(attachment: JobAttachmentSchema, options?)`, forwarding to the shared read path. `getById` keeps `id: string` (also what @Raina451 asked for above).

The shared fetch/transform moved into a private `fetchById()` so both public methods keep their own `@track` without double-counting telemetry (`agent_docs/conventions.md:24` delegation anti-pattern). +2 unit tests, and a `getFor()` row in `docs/oauth-scopes.md` — same scope, same endpoint.

src/models/orchestrator/attachments.types.ts
```diff
@@ -48,11 +48,49 @@ export interface AttachmentResponse {
   blobFileAccess: BucketGetUriResponse;
 }
 
+/**
+ * A job attachment as delivered to an automation's input by the platform.
+ *
+ * Declare a field of this type in a coded function's input contract to get a file
+ * picker in Studio Web and Orchestrator. Carries metadata only — fetch the content
+ * with `attachments.getById()`.
```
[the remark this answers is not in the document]
lucarachiteanu: Applied in 15974a5 — the doc now points to `attachments.getFor()`.

src/models/orchestrator/attachments.models.ts
```diff
@@ -28,6 +30,21 @@ export interface AttachmentServiceModel {
    */
   getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
 
+  /**
+   * Gets the attachment a job attachment input refers to.
+   *
+   * @param attachment - A {@link JobAttachmentSchema} received as automation input
```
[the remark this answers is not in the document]
lucarachiteanu: Applied in 15974a5.

src/models/orchestrator/attachments.models.ts
lucarachiteanu: Fixed: JSDoc updated.

src/services/orchestrator/attachments/attachments.ts
lucarachiteanu: Fixed: getFor added.

src/models/orchestrator/attachments.types.ts
lucarachiteanu: Fixed: JSDoc updated.

src/models/orchestrator/attachments.models.ts
lucarachiteanu: Fixed: reverted, getFor added.

src/models/orchestrator/attachments.types.ts
lucarachiteanu: Fixed: JSDoc trimmed.

src/models/orchestrator/attachments.models.ts
```diff
@@ -28,6 +30,21 @@ export interface AttachmentServiceModel {
    */
   getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
 
+  /**
+   * Gets the attachment a job attachment input refers to.
+   *
+   * @param attachment - The job attachment input received from the automation
+   * @param options - Optional query parameters (expand, select)
+   * @returns Promise resolving to the attachment
+   * {@link AttachmentResponse}
```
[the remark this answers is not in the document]
lucarachiteanu: Fixed in 386ec1d: link inlined.

src/services/orchestrator/attachments/attachments.ts
```diff
@@ -33,6 +35,29 @@ export class AttachmentService extends BaseService implements AttachmentServiceM
       throw new ValidationError({ message: 'id is required for getById' });
     }
 
+    return this.fetchById(id, options);
+  }
+
+  @track('Attachments.GetFor')
```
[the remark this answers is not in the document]
lucarachiteanu: Added in 386ec1d — `getFor` describe block in `attachments.integration.test.ts` building a `JobAttachmentSchema` from `config.orchestratorAttachmentId` (basic retrieval + `select` forwarding). Also moved the unit tests into their own `getFor` block with test constants instead of literals, and added API-error + select-forwarding cases there.

Note: no integration env on my machine (`ORCHESTRATOR_ATTACHMENT_ID` unset), so these ran through typecheck/lint only — CI is the first live run.

tests/unit/services/orchestrator/attachments.test.ts
```diff
@@ -157,6 +157,76 @@ describe('AttachmentService Unit Tests', () => {
...
+          params: expect.objectContaining({
+            '$select': ATTACHMENT_TEST_CONSTANTS.ODATA_SELECT_FIELDS,
+          }),
+        })
+      );
+    });
+
+    it('should throw ValidationError when the attachment ID is empty', async () => {
+      await expect(attachmentService.getFor({ ...jobAttachment, ID: '' })).rejects.toThrow(
+        ATTACHMENT_TEST_CONSTANTS.ERROR_ATTACHMENT_ID_REQUIRED
```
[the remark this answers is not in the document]
lucarachiteanu: Fixed in 7fe4404; dropped the now-unused message constant.

tests/unit/services/orchestrator/attachments.test.ts
```diff
@@ -157,6 +157,74 @@ describe('AttachmentService Unit Tests', () => {
...
+      await expect(attachmentService.getFor({ ...jobAttachment, ID: '' })).rejects.toBeInstanceOf(ValidationError);
+
+      expect(mockApiClient.get).not.toHaveBeenCalled();
+    });
+
+    it('should handle API errors', async () => {
+      const error = createMockError(ATTACHMENT_TEST_CONSTANTS.ERROR_ATTACHMENT_NOT_FOUND);
+      mockApiClient.get.mockRejectedValue(error);
+
+      await expect(attachmentService.getFor(jobAttachment)).rejects.toThrow(
```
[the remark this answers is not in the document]
lucarachiteanu: Fixed in ffd175a.

anil-uipath (approved):

vnaren23 (commented):
Instead of adding a new method lets add the functionality into existing itself.

```
getById(id: string, options?: AttachmentGetByIdOptions): Promise<AttachmentResponse>;
```

This is the existing method, if the options can take fullName, mimeType, metadata  as separate properties in AttachmentGetByIdOptions , your use case should suffice right? So caller will just need to do

```
getById(<id>, { fullName: <name>, mimeType: <type>, metadata: <metadata> });
```

lucarachiteanu:
@vnaren23 done in d7a5721 — `getFor()` is removed, along with `AttachmentGetForOptions`, the private `fetchById()` split, its tests and the `docs/oauth-scopes.md` row. The PR is back to a single type: `JobAttachmentSchema`. Callers fetch the file with the existing method: `attachments.getById(input.invoice.ID)`.

I did not add `fullName` / `mimeType` / `metadata` to `AttachmentGetByIdOptions`. `getById` forwards every option key to the API as an OData query param (`transformOptions` → `addPrefixToKeys` with `$`), so those fields would either go out as `$fullName=...` or have to be special-cased and dropped, and either way the request for `Attachments({id})` never uses them — the caller already has the name and MIME type in hand, and the response returns them again. If you'd still like them on the options type as documentation-only fields, say so and I'll add them.

lucarachiteanu:
CI note on d7a5721: `coverage / coverage` (and `test-and-build`, which only echoes its result) is red from live-API integration failures unrelated to this PR — Data Fabric `entities-records` / `entities-schema` / `attachment`, Maestro `process-instances` / `processes`, and Agents `feedback` (500 "Bulk insert failed due to a database error", a 403, and a `Paused` vs `Running` state assertion). The Orchestrator `attachments` suite passed 8/8. A rerun failed the same way, and the docs-only branch `docs/faq-scope-changes` (one `.md` file changed) fails the same two jobs today, so this is the test tenant, not the diff. This PR now touches a single type file with no service code, so no integration test exercises it beyond compiling.

Raina451 (approved):

PR #706, opened 2026-09-01
docs(coded-apps): document redirectUri override for multi-hostname apps

swati354:
Adds a tip under the `"Initialize the UiPath SDK"` section of the coded-apps getting-started guide, showing users how to override `redirectUri` at runtime when their app is reachable at multiple hostnames.

Screenshot 2026-09-01 at 7 59 11 PM

### DIFF
--- a/docs/coded-apps/getting-started.md
+++ b/docs/coded-apps/getting-started.md
@@ -260,6 +260,21 @@ const sdk = new UiPath()
 await sdk.initialize()
 ```
 
+!!! tip "Tip: Overriding the redirect URI"
+    By default, the SDK reads `redirectUri` from the `<meta name="uipath:redirect-uri">` tag. You can also compute it at runtime, which is useful when the same app is reachable at multiple URLs (for example, orgs whose name contains `_` are also reachable at the hyphen form, or apps served from both a self-hosted URL and `uipath.host`):
+
+    ```typescript
+    import { UiPath } from '@uipath/uipath-typescript/core'
+    import { getAppBase } from '@uipath/uipath-typescript'
+
+    const sdk = new UiPath({
+      redirectUri: window.location.origin + getAppBase(),
+    })
+    await sdk.initialize()
+    ```
+
+    `getAppBase()` returns the app root injected by the platform, so the URI stays stable regardless of which route the user is on.
+
 ---
 
 ## Deploy


vnaren23 (approved):

docs/coded-apps/getting-started.md
```diff
@@ -260,6 +260,21 @@ const sdk = new UiPath()
 await sdk.initialize()
 ```
 
+!!! tip "Tip: Overriding the redirect URI"
+    By default, the SDK reads `redirectUri` from the `<meta name="uipath:redirect-uri">` tag, which has a fixed hostname. If your app is reachable at multiple hostnames, you can compute `redirectUri` at runtime instead:
```
Raina451: > If your app is reachable at multiple hostnames

Every different hostname (e.g. org1.uipath.host and org2.uipath.host) will have a different value for the uipath:redirect-uri meta tag. So, reading it from the meta tag should always be enough. Even if someone edits the routing name, the meta tag gets updated accordingly.

Fetching it at runtime is just another option. I can't think of a case where the meta tag would fail 🤔, apart from the underscore case, which is very specific to a few orgs.

Maybe we can phrase it more like: "You can also get the redirect URI at runtime like this..." and mention that this can be useful for underscore orgs if they want the app to be reachable through both hostnames, or when they want app to be reachable at self hosted app url as well as uipath app url.

Saying "multiple hostnames" could be confusing because different hostnames usually imply different orgs, which is probably how most people will interpret it.

wdyt?

swati354: hmm, how about this then

> By default, the SDK reads `redirectUri` from the `<meta name="uipath:redirect-uri">` tag. You can also compute it at runtime, which is useful when the same app is reachable at multiple URLs (for example, orgs whose name contains `_` are also reachable at the hyphen form, or apps served from both a self-hosted URL and `uipath.host`):

Raina451: sounds good.

docs/coded-apps/getting-started.md
```diff
@@ -260,6 +260,21 @@ const sdk = new UiPath()
...
+    import { UiPath } from '@uipath/uipath-typescript/core'
+    import { getAppBase } from '@uipath/uipath-typescript'
+
+    const sdk = new UiPath({
+      redirectUri: window.location.origin + getAppBase(),
+    })
+    await sdk.initialize()
+    ```
+
+    `getAppBase()` returns the app root injected by the platform, so the URI stays stable regardless of which route the user is on.
```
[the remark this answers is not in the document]
swati354: want to keep the guideline general.

Raina451 (approved):

PR #707, opened 2026-09-02
test(data-fabric): split entities integration tests into records/query/schema files [APPS-37378]

Sarath1018:
## Why

- `entities.integration.test.ts` (1886 lines, 138 tests) ran fully serial in one file: **~26 min**, the wall-clock bottleneck of the whole integration suite.
- Vitest parallelizes per file — splitting lets the areas run on separate workers.

## What changed

| File | Modes | Covers |
|------|-------|--------|
| `entities-records.integration.test.ts` | v0 + v1 | getAll/getById, record CRUD (service + entity-level), updateRecordById, import, single-record deletes, folder-scoped CRUD |
| `entities-query.integration.test.ts` | v0 + v1 | queryRecordsById: pagination, aggregates, havingFilter, expansionLevel 0–3, LEFT/INNER joins, join negative control |
| `entities-schema.integration.test.ts` | v1 only | create, updateById, sqlType constraint defaults, RELATIONSHIP + FILE, MULTILINE_MAX, deleteById, bound schema methods |

- **Test bodies ported verbatim** — all 88 `it`/`describe` declarations byte-identical to the original (diff-verified).
- **Schema suite runs v1 only**: schema APIs shipped in the modular-SDK era (#352), and both init modes resolve to the same `EntityService` — the v0 pass doubled the slowest file while re-proving nothing. v0 init wiring stays covered by the dual-mode records/query files.
- **All schema DDL stays in one file**: two parallel schema workers caused DF 500s and stalled record calls past the 30s timeout (tenant-wide DDL contention).
- Split-required guards only (no test rewrites): query file resolves its own fallback entity id; the fallback capture and `includeFolderEntities` superset check skip transient `sdk_`-prefixed entities created by the parallel schema worker.
- **Every schema-write test now carries the 60s timeout its slower siblings already used** — the alpha tenant's entity creates baseline at 5–15s server-side and exceed the 30s default during slow phases (a passing test measured 30.4s in CI). Read paths keep the 30s default so genuine slowness there stays visible. (Transport was ruled out: 300 consecutive probe requests over both Node fetch and curl showed zero stalls — the intermittent hangs/500s are alpha server-side, evidence traceids in the run history.)
- `tests/integration/README.md` layout updated.

## Cleanup-leak fixes

- **`importRecordsById` rows were never cleaned up** — the import API returns counts, not IDs; the shared test entity had accumulated **3,448 `BulkImport_*` records**. The test now resolves its rows' IDs via `queryRecordsById` and tracks them for the existing `afterAll` cleanup.
- **Relationship targets leaked one entity per run** — `Promise.all` + silent catch swallowed the 400 when a RELATIONSHIP target's delete raced ahead of its source's; **581 `sdk_*` entities** had accumulated. Cleanup now retries first-pass failures once and `console.warn`s anything still failing.
- Tenant backlog swept as part of this work (leaked entities + records removed; fixtures untouched). Slimming the shared entity also removes a latent `pageSize: 25` flake in the join tests.

## Verification

- Live runs (PAT auth): records + query (both modes) **74 passed** ≈ 4.1 min; schema (v1) **23 passed** ≈ 4.7 min.
- **Suite wall clock ≈ 4.7 min, down from ~26 min.**
- CI (full env) ran all three files green, including folder-scoped + join tests; 18 local skips are the pre-existing `DATA_FABRIC_TEST_FOLDER_ENTITY_ID` guard, documented in `.env.integration.example`.
- `npm run typecheck`, `oxlint`, and all 2614 unit tests pass.


AditiGoyalUipath (approved):

.github/workflows/coverage.yml
```diff
@@ -130,7 +130,12 @@ jobs:
           EOF
 
       - name: Run Integration Tests
-        run: npm run test:integration:coverage -- --run
+        # Integration tests are network-bound (workers mostly await HTTP), so run
+        # more workers than vitest's CPU-based default (cores-1 = 1 on this 2-vCPU
+        # runner). 2 is the risk-adjusted sweet spot: ~2x faster than serial while
+        # halving how many in-flight tests an alpha-tenant stall window can catch
+        # (4 workers measured 3.2x but a stall window failed 10+ tests at once).
+        run: npm run test:integration:coverage -- --run --maxWorkers=2
```
Raina451: use 1-2 liner comment pls if needed.

Raina451 (approved):

PR #709, opened 2026-09-02
Add sample for du validation station in coded action app

Sandeepan-Ghosh-0312:
Adds two Coded Action App samples for the Document Understanding **Validation Station**, and brings the two existing standalone validation samples onto the same widget version and loading model.

## What's new

| Sample | Shows |
|---|---|
| [`action-app-with-document-validation`](samples/coded-action-apps/action-app-with-document-validation) | The whole app UI is the `ValidationStation` widget. Reads the task, hands `contentValidationData` to the widget, completes the action on submit. |
| [`action-app-with-document-validation-subcomponents`](samples/coded-action-apps/action-app-with-document-validation-subcomponents) | The same review flow composed from the five subcomponents — document viewer, document type, extraction fields, line-item editor, business rules — in a custom grid, linked by one shared `instanceId`. |

Both are complete Vite + React 19 + TS apps with a README, `action-schema.json`, `uipath.json.example` and a preview GIF, and both are listed in the Template Gallery.

## Also in here

**All four DU validation samples moved to `@uipath/ui-widgets-validation-station@1.1.0-rc.0`** — the two new ones plus `document-validation-app` (was `1.0.0`) and `document-validation-subcomponents-app` (was `1.0.1`). That release changes two things:

- **The widget is now loaded, not imported.** It ships as a separate bundle served from `<app base>/du-vs-wc`, registered by a `configureValidationStationWc()` call at startup. `scripts/stage-du-wc.mjs` stages it into `public/du-vs-wc` via `predev`/`prebuild` hooks; Vite serves `public/` verbatim in dev and copies it to `dist/` on build. This replaces two hand-written Vite plugins, so `vite.config.ts` now carries no bundling code of its own. `includeFonts: true` is required — the bundle's `fonts.css` is the only source of `@font-face`, so without it every icon renders blank.
- **API changes:** the `folderId` prop is gone (the folder comes from the payload), `useBucketArtifacts` → `useDuDocumentArtifacts`, and the save callbacks are renamed and re-signatured (`onSubmitComplete(result)` → `onSubmit(request, result?)`, etc.). `result` is now optional, and absent means nothing was persisted — so it's treated as a failure rather than completing a task over unsaved edits.

**Docs fixes** for two things that cost real debugging time:

- The redirect URI guidance said `https://cloud.uipath.com/<orgId>/<tenantId>/actions_`. Those are **GUIDs**, but the placeholders read as the org and tenant names in the address bar — and registering the name-based URI fails with `invalid_request` / `Invalid redirect_uri` with nothing pointing at the cause. The host was also hardcoded, so anyone on alpha or staging registered a URI their environment never sends. Applied to all five coded action app READMEs and the getting-started guide.
- Neither new README said how to run the app before deploying it, and the obvious guess is wrong: a coded action app is an iframe that receives its task over `postMessage`, so opening `localhost:5173` directly fails with *"Discarding event from invalid origin"*. There's now a **Run it locally** step covering the Action Center debug page.

**Gallery:** both new entries had `"preview": null` despite shipping GIFs, so their cards fell back to placeholder posters. Also added `document-validation-subcomponents-app`, which was the only sample missing from the gallery entirely, and moved its GIF to the conventional `screenshots/preview.gif`.

## Review notes

- **Requires `@uipath/ui-widgets-validation-station@1.1.0`**, now published as `latest`. All four samples pin it exactly — the package changed its API in a minor (1.0.1 → 1.1.0), so a `^` range is not safe.
- **`dist/` is ~84 MB per sample.** The widget bundle self-hosts its own fonts (~40 MB of woff2). Trimming them was tried and reverted — an iframe inherits no `@font-face` rules from the page around it, so the icons break.
- The two `.uis` template links and three screenshot placeholders in the new READMEs are **left as-is pending a check on who wrote them** — the files were never added to the repo.

## Verification

`tsc`, `eslint` and `npm run build` clean on all four samples against the published 1.1.0; `dist/du-vs-wc/` verified to contain `main.js`, `polyfills.js`, `styles.css`, `fonts.css` and `du-assets/`. The staging script now asserts those landed, so a partial copy fails the build instead of 404ing at runtime. End-to-end run confirmed in Action Center against a live DU validation action.

No SDK (`src/`) code is touched.


maninder-uipath (approved):

samples/document-validation-subcomponents-app/.gitignore
```diff
@@ -25,3 +25,6 @@ dist-ssr
 *.sw?
 .env
 uipath.json
```
Sarath1018, line 27: nit: not related to PR but remove uipath.json from gitignore

Sandeepan-Ghosh-0312: the coded apps samples have uipath.json in gitignore but not the coded action apps samples. Which one needs to change?

ioanajoj (approved):

samples/document-validation-subcomponents-app/.gitignore
Sarath1018: need to remove for both

Sarath1018 (approved):

PR #710, opened 2026-09-03
docs(faq): document why new scopes need a redeploy and a fresh token

amrit-agarwal-1:
Adds an FAQ entry for something that keeps catching people out: you grant a scope to the External Application, but calls still fail because the token has the old scopes.

Two reasons it doesn't just work:

1. The app reads the scopes it asks for from the `uipath:scope` meta tag, which is written at deploy time. Editing `uipath.json` does nothing until you redeploy.
2. The token is cached in `sessionStorage` and only checked for expiry, not for whether its scopes still match. So an open session keeps the old token.

The entry lists the four steps (grant, `uipath.json`, redeploy, fresh token) and shows `sdk.logout()`.

It also says a page refresh won't help. I checked this in a browser, `sessionStorage` survives a reload. Hard reload only skips the HTTP cache so it won't clear it either, though I couldn't test that one directly.

Docs only.

Side note: the scope isn't part of the cache key, so the SDK can't tell the token is stale. Adding it to the key would fix this properly and we could drop the FAQ entry. Can raise separately if you want.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -47,3 +47,35 @@ https://cloud.uipath.com/identity_/web/?errorCode=invalid_request&errorId=eyJDcm
   }
 }
 ```
+
+### Scope Changes Not Taking Effect
+
+**Scenario**: You grant a new scope to the External Application, but calls still fail with a permission error.
+
+**Cause**: Two things hold on to the old scopes.
+
+First, the app requests the scopes it had when it was last deployed. What you need to change depends on how those scopes are set:
+
+| How scopes are set | What to do after granting the scope |
+| --- | --- |
+| `uipath.json` with no `scope` field | Redeploy. Every scope granted to the client is picked up automatically at deploy time, so no code change is needed. |
+| `uipath.json` with `scope` set | Add the scope to `scope`, then rebuild, pack, publish, and deploy. |
+| `scope` passed to `UiPath()` in code | Add the scope in the code, then rebuild, pack, publish, and deploy. |
+
+Second, the token is cached in `sessionStorage`. The SDK only checks that it is well formed and not expired, never that its scopes still match, so an open session keeps using the old token.
+
+**Solution**:
+
+1. Add the scope in **Admin → External Applications**.
+2. Update and deploy the app as shown in the table above.
+3. Sign out and sign back in, or close the browser tab and reopen the app.
+
+```typescript
+// Sign out (clears the cached token)
+sdk.logout();
+
+// Sign out of UiPath as well, so the next sign in asks for credentials
+sdk.logout({ endSession: true });
+```
+
+Refreshing the page won't help. `sessionStorage` belongs to the browser tab and survives reloads, including a hard reload. Closing the tab clears it, as does **DevTools → Application → Clear site data**.


docs/FAQ.md
```diff
@@ -46,4 +46,32 @@ https://cloud.uipath.com/identity_/web/?errorCode=invalid_request&errorId=eyJDcm
...
   }
 }
-```
\ No newline at end of file
+```
+---
+
+## Scope Changes Not Taking Effect
+
+**Problem**: You grant a new scope to the External Application, but calls still fail with a permission error. The scopes in the JWT don't match what the app has been granted.
```
vnaren23: **Scenario** is better than **Problem**

docs/FAQ.md
```diff
@@ -46,4 +46,31 @@ https://cloud.uipath.com/identity_/web/?errorCode=invalid_request&errorId=eyJDcm
     "ClientId": "00000000-0000-0000-0000-000000000000"
   }
 }
-```
\ No newline at end of file
+```
+---
+
+## Scope Changes Not Taking Effect
```
Raina451: can this be added as a subsection under ## Authentication Errors itself? because it falls under that only

docs/FAQ.md
```diff
@@ -46,4 +46,31 @@ https://cloud.uipath.com/identity_/web/?errorCode=invalid_request&errorId=eyJDcm
...
+
+**Scenario**: You grant a new scope to the External Application, but calls still fail with a permission error. The scopes in the token don't match what the app has been granted.
+
+**Cause**: Two things hold on to the old scopes:
+
+1. The app reads the scopes it asks for from the `uipath:scope` meta tag in `index.html`, and that tag is written when the app is deployed. Changing `uipath.json` on its own does nothing to an app that is already deployed.
+2. The token is cached in `sessionStorage` under `uipath_sdk_user_token-<clientId>`. The SDK only checks that it is well formed and not expired, never that its scopes still match. So an open session keeps using the old token.
```
Raina451: i dont think we need to mention this detail:
> under `uipath_sdk_user_token-<clientId>`

docs/FAQ.md
```diff
@@ -46,4 +46,31 @@ https://cloud.uipath.com/identity_/web/?errorCode=invalid_request&errorId=eyJDcm
...
+```
+---
+
+## Scope Changes Not Taking Effect
+
+**Scenario**: You grant a new scope to the External Application, but calls still fail with a permission error. The scopes in the token don't match what the app has been granted.
+
+**Cause**: Two things hold on to the old scopes:
+
+1. The app reads the scopes it asks for from the `uipath:scope` meta tag in `index.html`, and that tag is written when the app is deployed. Changing `uipath.json` on its own does nothing to an app that is already deployed.
```
Raina451: there are 2 things basically :
1. if user is using uipath.json and not relying on hardcoded values in UiPath()
-    if in scope field in uipath.json they havent provided any scope -> in that case, at time of deployment, all the scopes which client is provided is being added to uipath:scope, so in this case after updating scope in external app, only redeploying is needed without change in code,
- if scope field is populated in uipath.json then code change is required with updated scope. and whole building code, packaging again , and publish, deploy needed.

3. if using hardcoded values in UiPath()
- again code change is required and re build , pack , publish, deploy is needed.

Raina451: i think we should mention all scenarios

docs/FAQ.md
amrit-agarwal-1: Added as subsection now

docs/FAQ.md
amrit-agarwal-1: removed

docs/FAQ.md
amrit-agarwal-1: Agree, I have mentioned all the scenarios now

Raina451 (approved):

vnaren23 (approved):

PR #713, opened 2026-09-04
feat(queues): apply resource overrides on operational byName paths; adopt getByKeyLookup

deepeshrai-tech:
## Summary

Two runtime-override gaps in the Queues service now closed:

- `startTransaction({ name })` now calls `resolveOverride('Queue', name, folderPath)` before building the wire body / folder headers. A cross-folder redirect steers both the transaction body (`transactionData.Name`) and the `X-UIPATH-FolderPath-Encoded` header to the override target. The `{ id }` branch is unaffected (the id → name lookup already hits the server's own id-based route).
- `insertItemByName` applies the same override resolution before posting to `ADD_ITEM`, so overrides published by the host reach the Add-Queue-Item wire body.

Cleanup: `getByKey` now delegates to the shared `FolderScopedService.getByKeyLookup` helper (added in the Assets ref-based PR #684). The service-local GUID/OData plumbing and the `NotFoundError` / `CollectionResponse` imports go away.

Behavior for the `getByName` read path is unchanged — it already went through `getByNameLookup`, which applies overrides internally.

## Test plan

- [x] `npm run lint` — 0/0
- [x] `npm run test:unit` — 2731/2731 pass (single failing file is a pre-existing `check-samples.test.ts` module-not-found)
- [x] `npm run build` — exit 0
- New tests:
  - Override redirect on `insertItemByName` — wire `Name` + `X-UIPATH-FolderPath-Encoded` both target the override
  - Override redirect on `startTransaction({ name })` — same

### DIFF
--- a/src/services/orchestrator/queues/queues.ts
+++ b/src/services/orchestrator/queues/queues.ts
@@ -28,8 +28,7 @@ import {
   transformData,
   transformOptions
 } from '../../../utils/transform';
-import { NotFoundError, ValidationError } from '../../../core/errors';
-import { CollectionResponse } from '../../../models/common/types';
+import { ValidationError } from '../../../core/errors';
 import { createHeaders } from '../../../utils/http/headers';
 import { FOLDER_ID } from '../../../utils/constants/headers';
 import { resolveFolderHeaders } from '../../../utils/folder/folder-headers';
@@ -40,7 +39,7 @@ import { PaginationHelpers } from '../../../utils/pagination/helpers';
 import { PaginationType } from '../../../utils/pagination/internal-types';
 import { QueueMap, QueueItemMap, QueueItemProcessingErrorMap } from '../../../models/orchestrator/queues.constants';
 import { track } from '../../../core/telemetry';
-import { GUID_REGEX } from '../../../utils/validation/guid';
+import { resolveOverride } from '../../../utils/overrides/resolve-override';
 
 /**
  * Transforms a raw API queue item into the SDK shape. `SpecificContent` and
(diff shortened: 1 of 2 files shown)


tests/unit/services/orchestrator/queues.test.ts
```diff
@@ -869,6 +870,34 @@ describe('QueueService Unit Tests', () => {
 
       expect(mockApiClient.get).not.toHaveBeenCalled();
     });
+
+    it('redirects both the wire name and folderPath header when a runtime override matches the queueName', async () => {
```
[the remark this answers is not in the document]
deepeshrai-tech: Fixed in 896cc5452 — moved the override test out of `describe('getByKey', ...)` and into `describe('insertItemByName', ...)` right after the folder-scoping tests. 55/55 pass, description matches the method under test.

src/services/orchestrator/queues/queues.ts
```diff
@@ -307,23 +289,31 @@ export class QueueService extends FolderScopedService implements QueueServiceMod
 
   @track('Queues.StartTransaction')
   async startTransaction(queue: QueueRef, options: QueueStartTransactionOptions = {}): Promise<QueueItem | null> {
```
shivendra6720, line 291: @deepeshrai-tech When an override redirects the folder on startTransaction, how should the follow-up completeTransaction / setTransactionProgress be scoped? Is the expectation that deployed apps never pass explicit design-time folderPaths (relying on the injected default), and if so, should that be documented — or do we need the item's folder carried across (e.g. via the returned item)?

deepeshrai-tech: Fair concern — this is a real design gap. The scope of this PR is limited to *insertion-time* override resolution (`startTransaction` and `insertItemByName` when passed a name). Downstream lifecycle calls (`completeTransaction`, `setTransactionProgress`) take a numeric `itemId` (a stable server handle) and re-derive their folder header from whatever the caller passes in `options`.

Two observations on how deployed apps actually consume this:

1. **The typical deployed-apps flow doesn't pass explicit `folderPath` on downstream calls** — they rely on the `uipath:folder-key` meta-tag fallback that `resolveFolderHeaders` reads, and the SDK's init-time config carries the effective folder. So when startTransaction's override redirects across folders, the follow-up completeTransaction/setTransactionProgress calls inherit the *runtime* folder from the same fallback — no per-call plumbing needed.

2. **When the caller *does* pass a design-time `folderPath` on `completeTransaction`**, they'll bypass the override and hit the wrong folder. That's the case worth documenting, and there's a cleaner fix: return the effective folder on the item response and thread it through the bound methods that `startTransaction` returns. That way `item.complete(outcome)` on the returned QueueItem uses the same folder as the insertion.

I'd rather do (2) as a separate PR that touches all three lifecycle methods together, so the returned QueueItem carries the effective folder consistently. If that's acceptable, I'll add a doc note on `startTransaction` here calling out the design-time-folderPath caveat and file a follow-up ticket for the bound-methods work.

PR #719, opened 2026-09-07
docs(helpers): align helper methods page with SDK doc standards

anil-uipath:
## What Changed

The [Helper Methods](https://uipath.github.io/uipath-typescript/helpers/) page had two sections and no usage examples, with its snippets embedded in prose.

It now follows the same shape as the other TypeScript SDK guide pages: sectioned by helper, with each concept carrying its own example and the interface types linked to their generated API reference pages. A minimal `httpRequest()` call opens the page, and retries and backoff — the main reason to reach for the helper — follow directly beneath it.

Docs-only. One file, no source or build changes.

## Structure

```
# Helper Methods           intro + helper table + !!! warning
## httpRequest()           basic call
### Retries and Backoff
### Backoff Strategies
### Retry-After Header
### Sending a Body and Query Parameters
### Error Handling         condition/behavior table + example
## wait()
```

## Consistency

| Before | After |
|--------|-------|
| All examples in a trailing `## Usage Examples` section; the first code on the page was a 12-line retry config | Organized by helper, each concept carrying its own example beneath it, as `error-handling.md` does per error type; a minimal call is now the first code |
| `HttpRequestInit`, `HttpResponse`, `RetryOptions` re-documented inline | Linked to their generated interface pages, as the Entities page does for `EntityCreateOptions` |
| "Error Contract", a term used nowhere else in the docs | "Error Handling", with a cross-link to that guide |
| `wait` listed as a helper but present only as the last example | Its own `## wait()` section |
| Network-error behavior split across two sections 60 lines apart | Merged into one |
| Third-party-endpoint caveat in a plain paragraph | `!!! warning` admonition, as `authentication.md` uses |
| `instanceof NetworkError` in examples | `isNetworkError()` guard, as `error-handling.md` uses |
| `####` headings, absent from the page TOC (`toc_depth: 3`) and used by no sibling page | `###` |

## Accuracy

The page's error contract was wrong in three ways, each verified against the implementation:

- **It advised checking `ok` "rather than catching"** — which reads as *never throws* — then showed an example that catches. Now stated non-exhaustively: a returned status does not by itself throw, and a few conditions do.
- **The error table omitted `ValidationError`.** `prepareBody()` runs outside the `try` wrapping `fetchWithRetry`, so a `ReadableStream` body throws `ValidationError`, not `NetworkError` (`http-request.ts:144` vs `:152`).
- **The network-retry claim was ungated.** `retriesAllowed` is `0` for a method outside `retryMethods` (`fetch-with-retry.ts:107`), so the `POST` in the example above it is not retried on a transport failure.

Smaller corrections: the first table row narrowed to 4xx/5xx, since a 200 also "returned a status" but resolves `ok: true`; "service methods, which always throw" → "throw on any failed status" (`api-client.ts:114`); "proxies arbitrary third-party calls" → "makes", since it calls `fetch` directly; the `<apiKey>` placeholder convention per `agent_docs/rules.md`; `params` now shows an array value, since the prose documents repeated query params but no example demonstrated it; two unused `const response` bindings dropped.

## Files

| Area | Files |
|------|-------|
| Docs | `docs/helpers.md` (224 → 130 lines) |

`mkdocs.yml` already lists the page. No new methods, so no `docs/oauth-scopes.md` entry is owed — `httpRequest` sends no UiPath authentication and consumes no scope, and `wait` makes no request. Not paginated, so no `docs/pagination.md` change.

### DIFF
--- a/docs/helpers.md
+++ b/docs/helpers.md
@@ -12,7 +12,7 @@ Standalone helper functions that you call directly. Unlike the SDK's services, t
 
 ## httpRequest()
 
-Takes a URL and optional [`HttpRequestInit`](/uipath-typescript/api/interfaces/HttpRequestInit), and resolves to an [`HttpResponse`](/uipath-typescript/api/interfaces/HttpResponse). A status the server returned resolves rather than throwing, even a 4xx or 5xx, so check `ok` to spot a failed request. A call that never reached the server does throw — see [Error Handling](#error-handling).
+Takes a URL and optional [`HttpRequestInit`](/uipath-typescript/api/interfaces/HttpRequestInit), and resolves to an [`HttpResponse`](/uipath-typescript/api/interfaces/HttpResponse). A status the server returned does not by itself throw, even a 4xx or 5xx, so check `ok` to spot a failed request. A few conditions do throw — see [Error Handling](#error-handling).
 
 ```typescript
 import { httpRequest } from '@uipath/uipath-typescript/core';
@@ -35,7 +35,7 @@ Pass `retry` to change any of that. See [`RetryOptions`](/uipath-typescript/api/
 ```typescript
 import { httpRequest } from '@uipath/uipath-typescript/core';
 
-const response = await httpRequest('https://api.example.com/v1/orders', {
+await httpRequest('https://api.example.com/v1/orders', {
   method: 'POST',
   body: { sku: 'ABC-123' },
   retry: {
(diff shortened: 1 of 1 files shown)


docs/helpers.md
```diff
@@ -27,92 +26,105 @@ if (response.ok) {
...
-you have:
+### Retries and Backoff
 
-```typescript
-if (response.ok) {
-  const order = response.data as { id: string };
-  console.log(order.id);
-}
-```
+By default, the idempotent methods — `GET`, `HEAD`, `PUT`, `DELETE`, and `OPTIONS`, per RFC 9110 — are retried up to twice on a transient failure: a transport error, or a `408`, `429`, `500`, `502`, `503`, or `504`. `POST` and `PATCH` are not retried, since a replayed `POST` can create the same resource twice.
```
vnaren23: hyperlink RFC 9110

vnaren23 (approved):

docs/helpers.md
anil-uipath: Linked in 0b4db13 — points at [RFC 9110 §9.2.2, Idempotent Methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2), the clause the default `retryMethods` set is derived from.

docs/helpers.md
```diff
@@ -27,89 +26,101 @@ if (response.ok) {
...
+Objects and arrays are sent as JSON. Array values in `params` are sent as repeated query parameters.
 
 ```typescript
-// one attempt, five second ceiling, no retrying involved
-const response = await httpRequest('https://api.example.com/v1/orders', { timeoutMs: 5000 });
+import { httpRequest } from '@uipath/uipath-typescript/core';
+
+const response = await httpRequest('https://api.example.com/v1/orders', {
+  method: 'POST',
+  headers: { 'x-api-key': apiKey },
```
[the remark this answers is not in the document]
anil-uipath: Already addressed in 9ee14ca — this review ran against the first commit (0b7cad0d9). Current state is `headers: { 'x-api-key': '<apiKey>' }`.

deepeshrai-tech (approved):

vnaren23 (approved):

PR #720, opened 2026-09-07
Add job key to coded action app getTask method

Sandeepan-Ghosh-0312:
## Summary

Surfaces `jobKey` — the GUID key of the job that created the task — on the `Task` object that `getTask()` resolves.

Type and documentation only. `getTask()` resolves Action Center's `LOADAPP` payload as-is, so the value already arrives on the message; this PR gives it a name on the public type so apps can read it without a cast.

## Why

A coded action app that creates Orchestrator resources needs the job behind the task to hang them off. The case in hand is file upload: `attachments.create()` takes a `jobKey`, and the link it makes is what decides who can read the file back — an attachment that belongs to no job is readable only by the account that created it, so an approver opening the action would get a 403 on the file the reviewer just uploaded.

```ts
const task = await codedActionApps.getTask();

await attachments.create(file.name, file, {
  folderId: task.folderId,
  jobKey: task.jobKey,   // the link that makes the attachment readable by anyone else
});
```

## What changed

| File | Change |
|------|--------|
| `src/types.ts` | `jobKey` added to the `Task` type |
| `src/coded-action-app.models.ts` | field added to the `getTask()` JSDoc example — the source of both the docs site and IDE hovers |
| `packages/coded-action-app/README.md` | `getTask()` field list and the `Task` type reference |
| `docs/coded-action-app-sdk/getting-started.md` | `getTask()` field list on the docs site |
| `packages/coded-action-app/package-lock.json` | drops stale `libc` arrays from optional rollup platform entries — an artifact of the npm version that regenerated the lock, no dependency or version change |

No behaviour change: no new events, no change to the existing ones, and nothing to migrate. Reading `task.jobKey` is opt-in.

## Release

The version bump is **not** in this PR — per the release workflow it ships on its own, so `jobKey` reaches consumers with the next publish of `@uipath/coded-action-app`.


Raina451 (approved):

Sandeepan-Ghosh-0312:
Will merge this once the Action Center changes have reached prod al rings

Sarath1018 (approved):

PR #724, opened 2026-09-09
feat(data-fabric): preserve raw sqlType on entity field metadata

AditiGoyalUipath:
Snapshot the raw sqlType before transformData renames+mutates it into fieldDataType so callers of getById / getByName / getAll now see the raw API wire type (e.g. NVARCHAR, INT, UNIQUEIDENTIFIER) alongside the friendly EntityFieldDataType name.


deepeshrai-tech:
@AditiGoyalUipath  is this PR needed?

PR #726, opened 2026-09-09
test(maestro): fix resumes and size lifecycle waits to case tests

Sarath1018:
Re-lands the pause/resume fix from #693 (closed unmerged) and hardens the maestro suites against fixture death and tenant instance growth.

- **Lost-resume race** (the `expected 'Paused' to be 'Running'` flake): a resume issued while the pause is still settling is accepted but can be lost — the test now re-issues the resume when the status settles on `Paused`.
- **Fixture health-check**: the seeded Running.Case instance can fault on its own (observed live); `resolveRunningInstance` re-verifies its status on every use and seeds a replacement when it is no longer Running.
- **Reuse before seeding, all fixtures**: Running.Case scavenges an existing Running instance (interrupted runs leave them Running indefinitely), the timer case reuses a Completed leftover (also zero reopen wait), matching the pattern the faulted retry fixture already uses. Terminal instances cannot be deleted via API, so creation is the only growth lever — steady-state creation drops to one instance per consuming test.
- **Wait sizing**: cancel 120s (its poll alone spans ~40–60s), reopen completion 180s (matches the retry fault wait), per CI-observed tails.

Verified green in both init modes.

### DIFF
--- a/tests/integration/shared/data-fabric/entities-records.integration.test.ts
+++ b/tests/integration/shared/data-fabric/entities-records.integration.test.ts
@@ -16,6 +16,7 @@ import {
   QueryFilterOperator,
   RawEntityGetResponse,
 } from '../../../../src/models/data-fabric/entities.types';
+import { EntityGetResponse } from '../../../../src/models/data-fabric/entities.models';
 import { DATA_FABRIC_TENANT_FOLDER_ID } from '../../../../src/utils/constants/endpoints/data-fabric';
 
 // Cache for choice set values to avoid repeated API calls within a test run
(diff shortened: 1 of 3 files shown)


amrit-agarwal-1 (approved):

Sarath1018:
Consolidated into #729. All six commits from this branch were cherry-picked there unchanged (verified file-by-file identical), alongside the join/valueList fix that this branch needed in order to pass — its join tests failed deterministically without it once the shared fixture grew past one page. Closing in favour of the single PR; branch kept in case anything needs to be recovered.

PR #731, opened 2026-09-10
test(data-fabric): stall-harden the entity-methods block

Sarath1018:
## Problem

The `Entity-level methods (via getById)` block is the recurring victim of alpha's gateway stalls: `updateRecords` and `deleteRecords` died at exactly ~30s on consecutive CI attempts. Two flow defects:

1. **Every test re-fetched the entity** via `entities.getById()` inside its own timeout budget — each write test had to fit two round-trips (metadata read + batch write) into 30s, doubling its exposure to a single stall window, and violating the repo rule that shared lookups belong in `beforeAll`.
2. **The block's write tests still ran on the default 30s timeout** — the earlier hardening pass (#723) sized the schema file's write paths and the folder-scoped block to 90s but missed this block.

## Fix

- One shared `getById` in a `beforeAll` (90s), stored in `let entity!: EntityGetResponse` — the block still exercises the getById→bound-methods path on a live response object; only the redundant re-fetches are gone.
- 90s budgets on the four write tests (`insertRecord`, `insertRecords`, `updateRecords`, `deleteRecords`); the two read tests keep the default.
- Assertions, ordering, and cleanup bookkeeping unchanged.

## Validation

- oxlint + tsc clean over the file; live run of the block passes 12/12 across both init modes (39s).

### DIFF
--- a/tests/integration/shared/data-fabric/entities-records.integration.test.ts
+++ b/tests/integration/shared/data-fabric/entities-records.integration.test.ts
@@ -16,6 +16,7 @@ import {
   QueryFilterOperator,
   RawEntityGetResponse,
 } from '../../../../src/models/data-fabric/entities.types';
+import { EntityGetResponse } from '../../../../src/models/data-fabric/entities.models';
 import { DATA_FABRIC_TENANT_FOLDER_ID } from '../../../../src/utils/constants/endpoints/data-fabric';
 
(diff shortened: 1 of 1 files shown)


Sarath1018:
Consolidated into #726 (commit cherry-picked onto sarath/fix-lost-resume-race) so the DF and maestro stall-hardening land together.

PR #732, opened 2026-09-11
feat(samples): add the coded-apps-with-functions sample collection

AlexBizon:
Adds `samples/coded-apps-with-functions/` — a collection of Coded App + Coded Functions samples, each pairing a web front end with a JS functions backend and shipping them as **one solution**, so a single `pack` / `publish` / `deploy` stands up the app, the functions, their triggers and any Orchestrator resource they need.

Each sample answers one question: **what is the function for here, and what would break without it?**

## Samples

| Sample | The question | What it shows |
|---|---|---|
| [`secrets`](samples/coded-apps-with-functions/secrets) | A credential must never reach the browser | Reads a Credential asset with the function's own robot identity, through the robot-execution endpoint — so no `AllowDirectApiAccess` flag has to be switched on. The readable set is an allowlist in code, because the robot identity is more privileged than the caller and must not fetch caller-named secrets. The page receives a username, a length and a fingerprint; never the value. |
| [`rest-backend`](samples/coded-apps-with-functions/rest-backend) | The front end expects a REST API | Three functions shaped as REST resources — a collection, an item, and a nested sub-resource. Filter vs address deciding query param vs path param, an empty collection answering `200` rather than `404`, and two distinct `404`s that name the ids or line numbers that would have worked. |
| [`calling-mode`](samples/coded-apps-with-functions/calling-mode) | How should a function be called? | Two functions with identical handlers — one declares `method` + `path`, the other declares neither. That single difference decides whether you get a URL to call or a job to start. Over ~10 KB in the HTTP trigger answers `500 errorCode 4801`; over ~512 KB out it answers `200` with an empty body. A job carries the same payload as an attachment either way. |

Three further samples are scaffolded as _planned_ rows in the collection README: `identities`, `connections`, `big-payloads`.

## Notes for reviewers

- **Shared conventions.** `lib/contract.ts` is the single source of truth for I/O types in every sample — the functions import it directly, the app imports it with `import type`, so a mismatch between the two halves is a build error rather than a runtime 400. `deploy-config.json` ships filled in, dummy secrets included, so the documented commands work as written; only the OAuth client id is the reader's to supply.
- **The `secrets` credential is a dummy** (`sample-partner-account` / `sample-secret-value-01`), checked in deliberately so the sample deploys and runs with nothing to fill in. The README documents how to point it at a real one.
- **OAuth scopes are `OR.Execution`, `OR.Folders`, `OR.Jobs`** across all three. Notably `secrets` does *not* request `OR.Assets`, even though it reads an asset — the function does that with its own robot token inside the job, so granting the app asset access would contradict the point of the sample.
- **Verified.** Each app and functions project typechecks clean and builds; each sample has been deployed and exercised against a live tenant, and the screenshots are from those runs.
- All three samples are listed in the docs Template Gallery (`docs/samples/index.md`).


deepeshrai-tech:
Hey @AlexBizon can u plz use this version [1.203.0-preview.180] of cli and codedapp-tool to scaffold codedapp , using codeapp init cmd and then add rest files, some changes have been added and this will ensure the solution gets uploaded correctly either in cloud or in local-workspace

PR #734, opened 2026-09-15
fix(config): name an empty clientId instead of "configuration not found" [APPS-37263]

ninja-shreyash:
## Why

A coded app calls `new UiPath()` and reads its config from meta tags. When `clientId` is empty the SDK throws "configuration not found. Ensure the @uipath/coded-apps plugin is set up", which is misleading because the config was found and only `clientId` is empty.

## What changed

`missingConfigMessage` now takes the resolved config. When base fields are present, there is no secret, and `clientId` is empty, it says: "clientId is empty. Set clientId in uipath.json and redeploy, or pass a client ID at deploy time." A genuinely absent config keeps the old not-found message.

Part of the deploy-time warning work: the platform emits it ([business-apps-jamjam#18088](https://github.com/UiPath/business-apps-jamjam/pull/18088)), the CLI relays it ([UiPath/cli#3850](https://github.com/UiPath/cli/pull/3850)), and this makes the runtime error specific.

## Verification

typecheck and oxlint clean.

[APPS-37263]

[APPS-37263]: https://uipath.atlassian.net/browse/APPS-37263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/config/config-utils.ts
```diff
@@ -66,8 +66,28 @@ export function normalizeBaseUrl(url: string): string {
  * Shared so the registry can reuse it when a service is constructed from a
  * UiPath instance that never resolved a configuration — otherwise the caller
  * only learns their instance is "invalid", not why.
+ *
+ * When a config was found but is unusable (the common case is a coded app whose
+ * injected config carries every field except an empty clientId), name the empty
+ * field instead of claiming nothing was found.
  */
-export function missingConfigMessage(): string {
+export function missingConfigMessage(config?: PartialUiPathConfig): string {
```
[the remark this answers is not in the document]
ninja-shreyash: Deliberately not adding a unit test here. This is a message-only change: it only alters the string thrown when config resolution already failed, and the auth/config gating (isCompleteConfig) is untouched. The branch was also since simplified to a single clientId check, so the multi-field cases above no longer apply. If SonarCloud new-code coverage flags it, we'll handle that on the gate rather than add a test for a wording change.

src/core/config/config-utils.ts
```diff
@@ -66,8 +66,28 @@ export function normalizeBaseUrl(url: string): string {
...
+  if (
+    config &&
+    hasRequiredBaseFields(config) &&
+    !hasSecretConfig(config) &&
+    !hasValidAuthConfig(config)
+  ) {
+    const oauthKeys = ['clientId', 'redirectUri', 'scope'] as const;
```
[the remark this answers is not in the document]
ninja-shreyash: stale

src/core/config/config-utils.ts
```diff
@@ -66,8 +66,21 @@ export function normalizeBaseUrl(url: string): string {
...
  * UiPath instance that never resolved a configuration — otherwise the caller
  * only learns their instance is "invalid", not why.
+ *
+ * When a config was found but only its clientId is empty (the common case is a
+ * coded app whose injected config carries every field except the client), name
+ * that instead of claiming nothing was found.
  */
-export function missingConfigMessage(): string {
+export function missingConfigMessage(config?: PartialUiPathConfig): string {
+  if (
```
Raina451, line 75: this has to be gated behind isBrowser , please check

src/core/config/config-utils.ts
```diff
@@ -66,8 +66,21 @@ export function normalizeBaseUrl(url: string): string {
...
+    config &&
+    hasRequiredBaseFields(config) &&
+    !hasSecretConfig(config) &&
+    !config.clientId
+  ) {
+    return 'UiPath SDK configuration is incomplete: clientId is empty. Set clientId in uipath.json and redeploy, or pass a client ID at deploy time.';
```
Raina451, line 81:
```suggestion
    return 'UiPath SDK configuration is incomplete: clientId is empty. Set non confidential clientId in uipath.json and redeploy, or pass a client ID at deploy time.';
```

Raina451 (approved):
please also add tests in tests/unit/core/uipath-zero-config.test.ts

src/core/config/config-utils.ts
ninja-shreyash: done

src/core/config/config-utils.ts
ninja-shreyash: done

src/core/config/config-utils.ts
```diff
@@ -66,9 +66,24 @@ export function normalizeBaseUrl(url: string): string {
...
+export function missingConfigMessage(config?: PartialUiPathConfig): string {
   if (isBrowser) {
+    // Coded-app runtime: an empty clientId next to an injected redirectUri/scope is a
+    // partial OAuth config, not a page with no auth intent — name the missing client.
+    if (
+      config &&
+      hasRequiredBaseFields(config) &&
+      !hasSecretConfig(config) &&
+      !config.clientId &&
+      (config.redirectUri || config.scope)
```
amrit-agarwal-1: check if we can handle missing redirecturi case too in error message, else looks good.

amrit-agarwal-1 (approved):

src/core/config/config-utils.ts
ninja-shreyash: done

PR #735, opened 2026-09-15
docs: publish the JS Functions docs from coded-functions-js on this site

vnaren23:
## What

The JS Functions package, CLI and docs live in [`UiPath/coded-functions-js`](https://github.com/UiPath/coded-functions-js). Authoring stays there, next to the code; this repo renders the pages as part of the SDK docs site, so readers find one site instead of two.

The pages are fetched at docs build time into a gitignored `docs/js-functions/` — the same contract the typedoc output already uses for `docs/api/` and `docs/coded-action-app-sdk/`.

## How

- **`scripts/fetch-js-functions-docs.sh`** — downloads the repo tarball over the API, drops the pages the source repo does not publish in its own nav (`README.md`, `requirements.txt`, `decisions/`), and rewrites absolute `uipath.github.io/coded-functions-js/` self-links to their new home. Relative links between pages keep working because the directory layout is preserved. The token travels in a header supplied through a `--config` file on stdin, so it appears neither in a remote URL (visible in git output) nor in `curl`'s argv (readable via `ps`).
- **`mkdocs.yml`** — a `JS Functions` nav section plus a dedicated `llmstxt` block. The block is separate so this content lands in `js-functions/llms.txt` rather than inflating `llms-full-content.txt`, which is deliberately size-capped for WebFetch ingestion (see `scripts/llms-preprocess.py`). Verified: that file stays at 275K.
- **`package.json`** — the fetch is chained onto `docs:api`, so `npm run docs:api && mkdocs build` keeps working locally and a future workflow cannot forget the step.
- **Both docs workflows** — a `Mint JS Functions Docs Token` step. The source repo is INTERNAL, so the default `GITHUB_TOKEN` cannot read it; each run mints a GitHub App installation token scoped to that single repo with Contents: read, and the action revokes it when the job ends. `docs.yml` declares the two secrets it needs on its `workflow_call` path and `publish.yml` passes exactly those, rather than `secrets: inherit` handing over the whole store.
- **`docs.yml`** — a `repository_dispatch` trigger (`js-functions-docs-updated`) so the source repo can refresh the section without waiting for the next SDK release. Docs otherwise deploy only on an SDK release or a manual dispatch.

## Failure modes, handled deliberately differently

`mkdocs-llmstxt` raises a `KeyError` during post-build for any page named in its sections that does not exist, so a partial or skipped fetch fails the entire build rather than dropping one page. That splits four ways:

| Situation | Behaviour | Why |
|---|---|---|
| No token — fork PRs (no secrets by design), a rotated App key, local builds | Placeholder pages, build green. The PR-preview minting step is skipped when the credential is absent | A PR gate should answer "did this PR break the docs build", not red-X on a credential outage |
| No token, on the **deploy** path | Hard fail | Placeholders must never reach the live site; the last good build stays up |
| A referenced page **removed/renamed** upstream | Hard fail naming the path | Drift worth surfacing loudly |
| A page **added** upstream | Warning, build continues | It renders at a URL no nav points to, but a new upstream page should not block a release |

The expected page list is derived from `mkdocs.yml` itself and resolved before anything destructive runs, so it cannot drift from the nav and an unreadable config cannot silently wipe the section.

## Verification

Real `mkdocs build` runs against the live source repo, plus the pipeline end-to-end in CI on this PR:

- **In CI:** token minted and scoped to `UiPath/coded-functions-js`, `Fetched 12 markdown pages`, `Token revoked` in post-cleanup.
- **On the published preview:** [`/js-functions/`](https://uipath.github.io/uipath-typescript/pr-preview/pr-735/js-functions/), `/js-functions/api/define-function/`, `/js-functions/cli-reference/` and `/js-functions/llms.txt` all return 200 with real content.
- **Locally, all four failure modes above** behave as tabulated, including the destructive-guard cases (missing config / no referenced pages → exit 1 with `docs/js-functions/` untouched).
- `typecheck` clean, `lint` 0 errors, 2788 unit tests pass.
- Not exercised live: the `workflow_call` secrets pass-through only runs on a release. Validated by inspection and YAML parse — dispatching `docs.yml` would publish to the live site, which is premature (see below).

## Before merge

- [x] GitHub App created and `CODED_FUNCTIONS_DOCS_APP_ID` / `CODED_FUNCTIONS_DOCS_PRIVATE_KEY` configured — confirmed working in this PR's CI run.
- [ ] **Confirm the visibility call.** `coded-functions-js` is INTERNAL and this docs site is public; merging is what makes that content public. Needs a sign-off from whoever owns that repo's visibility.
- [ ] Add the dispatch sender to `coded-functions-js/.github/workflows/docs.yml`, and retire that repo's own docs deploy (keeping `mkdocs build` as a CI gate) so there are not two live sites saying different things. Leave a redirect at `uipath.github.io/coded-functions-js`.
- [ ] Confirm that a `repository_dispatch` republishing the whole site from `main` is intended — it also publishes typedoc pages for SDK code merged but not yet released.

## Upstream follow-ups (not fixable here)

- **Four cross-page deep links land at page top.** Upstream's MkDocs uses `pymdownx` slugify, so `## Org / tenant context (\`ctx.platform\`)` becomes `#org--tenant-context-ctxplatform` (double dash); this site's default `toc` renders `#org-tenant-context-ctxplatform`. Affects two links to `platform-context.md` and two to `calling-orchestrator.md`. Fix upstream by pinning ids with `attr_list` (already enabled here), e.g. `## Org / tenant context (\`ctx.platform\`) { #org-tenant-context-ctxplatform }`. Changing this site's slugify instead would re-slug every existing page and break external deep links. MkDocs logs this at INFO, so the build gate stays green.
- **Nothing links `js-functions/llms.txt`**, unlike the sibling Coded Action App SDK section. The link has to be added upstream; if it is written as the absolute upstream URL, the script's rewrite maps it onto the right path automatically.

The fetch defaults to `main`; set `JS_FUNCTIONS_DOCS_REF` to a tag (`v0.2.0`, `v0.1.0` exist) to pin the published docs to a released version instead.


AlexBizon (approved):

scripts/fetch-js-functions-docs.sh
Raina451, line 1: should we add a preview label? as js functions are not GA yet. its better to communicate here i believe as theres no docs of js function on docs.uipath.com

.github/workflows/docs.yml
```diff
@@ -1,7 +1,19 @@
...
+    # Declared rather than inherited, so a caller hands over these two secrets
+    # and nothing else. Not required: without them the JS Functions fetch falls
+    # back to placeholder pages instead of failing the build.
+    secrets:
+      CODED_FUNCTIONS_DOCS_APP_ID:
+        required: false
+      CODED_FUNCTIONS_DOCS_PRIVATE_KEY:
+        required: false
   workflow_dispatch:
+  # Sent by UiPath/coded-functions-js when its docs/ changes on main, so the
```
Raina451, line 13: everything merged to main is released ?  or we need to fetch versions / latest tag? ref: https://github.com/UiPath/coded-functions-js/tags

.github/workflows/docs-pr-preview.yml
```diff
@@ -35,6 +37,10 @@ concurrency:
 jobs:
   build-and-preview:
     runs-on: uipath-ubuntu-latest
+    env:
+      # Surfaced as env because the `secrets` context is unavailable in `if:`.
+      # Lets the minting step below detect an absent credential.
```
Raina451: nit: its is self explainatory, we can remove code comments from this file

Raina451 (approved):
minor comments

scripts/fetch-js-functions-docs.sh
vnaren23: There is. Its waiting on this PR to go it so that same can be referenced.

.github/workflows/docs.yml
vnaren23: For now main, we'll improve later once they have a release convention.

PR #736, opened 2026-09-15
feat(entities): add Entities upsert service

Kumarvarinder87:
## Method Added

| Service | `entities.upsert()` | `upsert(entityRef: EntityRef, data: Record<string, any>, options?: EntityUpsertOptions): Promise<EntityUpsertResponse>` |
| Entity (bound) | `entity.upsert()` | `upsert(data: Record<string, any>, options?: EntityUpsertOptions): Promise<EntityUpsertResponse>` |

## Endpoint Called

| `upsert()` | POST | `datafabric_/api/v3/entities/{entityName}/upsert` | `DataFabric.Data.Write` |

- Extends `BaseService`; `folderKey` is sent as the `X-UIPATH-FolderKey` header, `expansionLevel` as a query param.
- **First `/api/v3/` endpoint in the SDK.** Kept separate from `ENTITY.UPSERT`, which is the entity *schema* upsert (`/api/Entity`) used by `create()` / `updateById()`.
- The route is **name-only** — v3 has no `entity/{entityId}/upsert` counterpart, unlike every v1 record route. An `{ id }` ref is resolved to a name via the existing private `resolveEntityName()` helper, the same approach `queryRecords()` already uses for the name-only join surface. The bound method passes `{ name }` directly, so it costs no extra lookup.
- Two write forms, selected by the server from the payload shape:
  - **Flat payload** → one record, matched on the entity's configured business key.
  - **Nested payload** (child records under a key named for their entity) → the whole tree applied as one transaction. A record carrying an `Id` updates that row; one without it is created, with the service filling in the parent foreign keys.
- The SDK does not gate on payload shape — which form an entity accepts is the server's decision.

## Example Usage

```typescript

await sdk.initialize();
const entities = new Entities(sdk);

// Single record, matched on the entity's business key
const result = await entities.upsert({ name: "Case" }, {
  caseId: "CASE-001",
  caseStatus: "Open"
});

// A report with two expenses, one of which has a line item — applied as one transaction
const tree = await entities.upsert({ name: "Report" }, {
  assignee: "assignee1",
  totalReportAmount: 25,
  Expense: [
    { vendor: "Vendor 1", totalExpense: 20, ExpenseLineItem: [{ expenseAmount: 5 }] },
    { vendor: "Vendor 2", totalExpense: 5 }
  ]
}, { folderKey: "<folderKey>" });

console.log(tree.transaction!.totalRecordsAffected); // 4
console.log(tree.transaction!.members[0].id);        // generated expense record ID
```

### Transform pipeline

**None** — the response is returned as received.

Verified against a live alpha response: the envelope is already camelCase throughout, apart from the root `Id`, so no case conversion applies. The transaction subtree in particular arrives camelCase (`entityName`, `affectedRows`, `members`, `totalRecordsAffected`), despite the generated OpenAPI clients suggesting otherwise.

Reshaping is also not possible here: which fields come back depends on the form the server applied. A single-record write echoes the record's fields; a tree write reports per-record outcomes in `transaction` and returns no field values. Picking a fixed set of fields would discard one form's payload, so the service returns `response.data` untouched and types `transaction` as optional.

### Field mapping

| API Response | SDK Response | Change | Reason |
|--------------|--------------|--------|--------|
| `Id` | `Id` | None | System record ID; matches `EntityRecord.Id` across the SDK |
| `transaction.*` | `transaction.*` | None | Already camelCase on the wire |
| record fields | record fields | None | Data Fabric field names are user-defined schema columns and are returned verbatim |

## Sample SDK Response

Both samples are real responses captured from the live integration run against `alpha` (org `entity`, tenant `IMDB`).

<details>
<summary><code>upsert()</code> — nested payload (transactional tree)</summary>

```json
{
  "Id": "416db6d6-0aad-f111-9b33-6045bda94a92",
  "children": {},
  "cascadeDeletedChildren": {},
  "deletedCount": 0,
  "transaction": {
    "totalRecordsAffected": 5,
    "entityName": "MEU_Report",
    "op": "Insert",
    "id": "416db6d6-0aad-f111-9b33-6045bda94a92",
    "affectedRows": 1,
    "noOp": false,
    "version": 1,
    "members": [
      {
        "entityName": "MEU_Expense",
        "op": "Insert",
        "id": "426db6d6-0aad-f111-9b33-6045bda94a92",
        "affectedRows": 1,
        "noOp": false,
        "version": 1,
        "members": [
          {
            "entityName": "MEU_LineItem",
            "op": "Insert",
            "id": "436db6d6-0aad-f111-9b33-6045bda94a92",
            "affectedRows": 1,
            "noOp": false,
            "version": 1,
            "members": []
          }
        ]
      }
    ]
  }
}
```

*Truncated to one of two `MEU_Expense` branches — the full response returns both, each with its own `MEU_LineItem`, totalling 5 records.*

`children`, `cascadeDeletedChildren` and `deletedCount` are composite-entity concerns that the envelope always sends and this route always leaves empty. They pass through rather than being stripped, since stripping them would require inferring which form the server applied.

</details>

<details>
<summary><code>upsert()</code> — flat payload (single record)</summary>

```json
{
  "Id": "6cf9cff4-0dad-f111-9b33-6045bda94a92",
  "children": {},
  "cascadeDeletedChildren": {},
  "deletedCount": 0
}
```

The record's own fields are echoed alongside `Id`, and `transaction` is absent. A business key only exists on case and templated entities, so native entities use the nested form; sending a flat payload for one is rejected by the server.

</details>

## Files

| Area | Files |
|------|-------|
| Endpoint | `src/utils/constants/endpoints/data-fabric.ts` |
| Types | `src/models/data-fabric/entities.types.ts` |
| Models | `src/models/data-fabric/entities.models.ts` |
| Service | `src/services/data-fabric/entities.ts` |
| Unit tests | `tests/unit/services/data-fabric/entities.test.ts` (10 tests) |
| Model tests | `tests/unit/models/data-fabric/entities.test.ts` (3 tests) |
| Integration tests | `tests/integration/shared/data-fabric/entities-records.integration.test.ts` (4 tests) |
| Test utils | `tests/utils/mocks/entities.ts`, `tests/utils/constants/entities.ts` |
| Test config | `tests/integration/config/test-config.ts`, `tests/.env.integration.example` |
| Docs | `docs/oauth-scopes.md` |

No changes needed to `package.json` / `rollup.config.js` — both Data Fabric barrels use `export *`, so the new method and types ship through the existing `@uipath/uipath-typescript/entities` subpath.

## Verification

| Check | Result |
|-------|--------|
| `npm run typecheck` | clean |
| `npm run lint` | 0 errors, 0 warnings |
| `npm run test:unit` | 2753 passed / 98 files |
| `npm run build` | exit 0 |
| `npm run docs:validate` | 0 errors |
| `npm run test:integration` (upsert) | 8/8 passed live against alpha (4 tests × v0 + v1) |

## Notes for reviewers

- **`@experimental`** on the service model, the `EntityMethods` interface, and the bound method: writing across related entities is gated behind the tenant's `multi-entity-write` feature flag, which is **off by default**. With it off, a nested payload returns a 400 naming the unsupported part.
- **Integration test fixtures**: the tree tests need two native entities where the child holds an active foreign key to the parent, set via the new `DATA_FABRIC_TEST_TREE_ENTITY_NAME` / `DATA_FABRIC_TEST_TREE_CHILD_ENTITY_NAME` keys. `beforeAll` throws with a specific message when they are missing, rather than skipping.
- **The flat form is unit-tested but not integration-tested** — it requires a case or templated entity, and the test environment has only native fixtures.
- **Server-enforced limits are documented in JSDoc, not validated SDK-side**: 500 records, 3 levels of nesting, 6 distinct entities per request. Passing `__Version__` in a record opts that row into a version check.
- **Cloudflare Workers whitelist still needed** for `datafabric_/api/v3/entities/*/upsert` — this is the SDK's first v3 Data Fabric path, so it is unlikely to match an existing pattern. `apps-dev-tools` was not available on the machine this was prepared on, so that PR has not been raised.

Refs DS-9281 (backend: `CommonEntityPlatform`)


tests/utils/mocks/entities.ts
```diff
@@ -328,6 +330,69 @@ export const createMockSingleInsertResponse = (
...
 };
 
+/**
+ * Creates a raw multi-entity transactional upsert response, in the wire shape the API returns.
+ *
+ * Shape verified against a live alpha response: the transaction tree is camelCase throughout,
+ * the root `Id` is the one PascalCase key, every node carries `noOp` and `version`, and a leaf
+ * sends `members: []` rather than omitting the key. `children`, `cascadeDeletedChildren` and
+ * `deletedCount` are always sent empty on this route and are included so the mock exercises
+ * the service dropping them.
```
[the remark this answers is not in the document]
Kumarvarinder87: Done

tests/unit/services/data-fabric/entities.test.ts
```diff
@@ -2435,6 +2438,213 @@ describe("EntityService Unit Tests", () => {
...
+
+      await expect(
+        entityService.upsert(
+          { name: ENTITY_TEST_CONSTANTS.TREE_ROOT_ENTITY_NAME },
+          ENTITY_TEST_CONSTANTS.TEST_TREE_RECORD_DATA,
+        ),
+      ).rejects.toThrow(TEST_CONSTANTS.ERROR_MESSAGE);
+    });
+
+                it("should reject with a ValidationError when the ref supplies neither id nor name", async () => {
```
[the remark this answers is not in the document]
Kumarvarinder87: Done

src/models/data-fabric/entities.models.ts
```diff
@@ -547,6 +549,66 @@ export interface EntityServiceModel {
...
+   * const result = await entities.upsert({ name: "Report" }, {
+   *   assignee: "assignee1",
+   *   totalReportAmount: 25,
+   *   Expense: [
+   *     { vendor: "Vendor 1", totalExpense: 20, ExpenseLineItem: [{ expenseAmount: 5 }] },
+   *     { vendor: "Vendor 2", totalExpense: 5 }
+   *   ]
+   * }, { folderKey: "<folderKey>" });
+   *
+   * console.log(result.transaction.totalRecordsAffected); // 4
```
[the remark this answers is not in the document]
Kumarvarinder87: Done

tests/integration/shared/data-fabric/entities-records.integration.test.ts
```diff
@@ -1321,6 +1323,187 @@ describe.each(modes)('Data Fabric Entities Records - Integration Tests [%s]', (m
...
+      const result = await entities.upsert(
+        { name: treeEntityName },
+        { ...rootData, [treeChildEntityName]: [childData] },
+      );
+      expect(result.transaction).toBeDefined();
+      const tx = result.transaction!;
+
+      expect(result).toBeDefined();
+      expect(result.Id).toBeDefined();
+      expect(result.transaction).toBeDefined();
```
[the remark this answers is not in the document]
Kumarvarinder87: Done

tests/utils/constants/entities.ts
```diff
@@ -79,6 +79,33 @@ export const ENTITY_TEST_CONSTANTS = {
...
+        totalExpense: 20,
+        ExpenseLineItem: [{ expenseAmount: 5 }],
+      },
+    ],
+  },
+  // An empty array is a cleared multi-choiceset field, not a set of child records
+  TEST_TREE_RECORD_DATA_EMPTY_CHILD_ARRAY: {
+    assignee: 'assignee1',
+    tags: [],
+  },
```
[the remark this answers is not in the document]
Kumarvarinder87: Done

src/models/data-fabric/entities.types.ts
```diff
@@ -626,6 +626,71 @@ export interface EntityUpdateResponse extends EntityOperationResponse {}
...
+  /** Name of the entity this record belongs to */
+  entityName: string;
+  /** Whether the record was created or updated */
+  op: EntityMultiEntityWriteOperation;
+  /** Record ID — generated by the service for an insert */
+  id: string;
+  /** Number of rows the write affected */
+  affectedRows: number;
+  /** Set only when an update carried no field values and merely verified the row exists */
+  noOp?: boolean;
```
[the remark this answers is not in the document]
Kumarvarinder87: done

tests/integration/config/test-config.ts
```diff
@@ -62,6 +62,11 @@ export interface IntegrationConfig {
   dataFabricTestJoinFieldName?: string;
   dataFabricTestJoinRelatedEntityName?: string;
   dataFabricTestJoinRelatedFieldName?: string;
+  // Parent/child fixture for the multi-entity transactional upsert test. Both are
+  // required (the test throws when either is missing): the child entity must hold an
+  // active foreign key to the parent, and both must be native entities.
+  dataFabricTestTreeEntityName?: string;
```
amrit-agarwal-1, line 68: dataFabricTestTreeEntityName / dataFabricTestTreeChildEntityName are plumbed through test-config.ts in three places each and never read. we should drop them or use them.

src/utils/constants/endpoints/data-fabric.ts
```diff
@@ -63,6 +63,8 @@ export const DATA_FABRIC_ENDPOINTS = {
     DELETE_RECORD_BY_NAME: (entityName: string, recordId: string) => `${DATAFABRIC_BASE}/api/EntityService/${entityName}/delete/${recordId}`,
     DELETE_BY_NAME: (entityName: string) => `${DATAFABRIC_BASE}/api/EntityService/${entityName}/delete-batch`,
     BULK_UPLOAD_BY_NAME: (entityName: string) => `${DATAFABRIC_BASE}/api/EntityService/${entityName}/bulk-upload`,
+    // v3 record upsert. Name-only — no by-id counterpart; nested child records apply as one transaction.
+    UPSERT_RECORD_BY_NAME: (entityName: string) => `${DATAFABRIC_BASE}/api/v3/entities/${entityName}/upsert`,
```
amrit-agarwal-1, line 67: I checked in apps-dev-tools. entities/[^/]+$ is anchored at the entity name, and the insert/update family sits under entities/entity/{name}/…. Needs a new entry, roughly:

new RegExp(`^\\/${DATAFABRIC_BASE}\\/api\\/v3\\/entities\\/[^\\/]+\\/upsert$`),

Otherwise browser apps are CORS-blocked on the only method this PR adds.

tests/unit/services/data-fabric/entities.test.ts
```diff
@@ -2435,6 +2438,213 @@ describe("EntityService Unit Tests", () => {
...
+
+      const result = await entityService.upsert(
+        { name: ENTITY_TEST_CONSTANTS.TREE_ROOT_ENTITY_NAME },
+        ENTITY_TEST_CONSTANTS.TEST_TREE_RECORD_DATA,
+      );
+
+      expect(result).toEqual(raw);
+    });
+
+    it("should upsert a single record from a flat payload", async () => {
```
amrit-agarwal-1, line 2521: This mock is the only coverage the flat form has. It's also the mode that needs no feature flag — so the mode that works on any tenant is unverified against the API, while the only live-tested mode needs multi-entity-write, which is off by default. The new provisioning creates a native pair, which by definition can't take a flat payload. Can a case or templated fixture be added?

src/utils/constants/endpoints/data-fabric.ts
Kumarvarinder87: https://github.com/UiPath/apps-dev-tools/pull/186

tests/integration/config/test-config.ts
Kumarvarinder87: Done

amrit-agarwal-1 (approved):

src/models/data-fabric/entities.types.ts
```diff
@@ -626,6 +626,74 @@ export interface EntityUpdateResponse extends EntityOperationResponse {}
...
+export interface EntityUpsertResponse extends EntityRecord {
+  /**
+   * What was written at each record, mirroring the nesting of the request.
+   * Present only when the payload nested child records.
+   */
+  transaction?: EntityMultiEntityWriteResponse;
+}
+
+/**
+ * Options for upserting a record
```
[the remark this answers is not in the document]
Kumarvarinder87: ChildArrayBlock isn't an open map — it's { Records: IList<IDictionary<string,object>>, HasMore: bool, … } plus a sub-resource pagination template. So declaring children honestly means modelling a composite-entity type hierarchy for a shape this method never returns, and declaring it as Record<string, unknown> would misstate the contract. Same for cascadeDeletedChildren, where the count semantics would be erased.

The convention cited doesn't quite apply. "Don't mark fields optional when the API always returns them" governs fields we declare. These aren't declared at all — they flow through EntityRecord's index signature. 

They're permanently empty for every SDK caller. They only populate for composite entities, which the SDK doesn't support — there are no composite methods. So publishing them adds three fields that are always {}, {}, 0, including a deletedCount on an upsert.

What I did instead

The legitimate concern underneath is that the passthrough was undocumented. So I documented it on EntityUpsertResponse — the envelope carries those three, they're composite-only, they arrive empty, and they're deliberately left undeclared. Anyone reading the type now knows the runtime shape without three inert fields in the public API and TypeDoc.

tests/unit/services/data-fabric/entities.test.ts
Kumarvarinder87: For flag of multi-entity-write, We have enabled it for all org in Alpha.
For composite one, we are not using case entities anywhere in sdk as of now. So skipping any test related to case or template one.

src/utils/constants/endpoints/data-fabric.ts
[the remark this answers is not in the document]
Kumarvarinder87: https://github.com/UiPath/apps-dev-tools/pull/186

tests/integration/shared/data-fabric/entities-records.integration.test.ts
```diff
@@ -151,6 +155,54 @@ async function buildDummyRecord(entityMetadata: RawEntityGetResponse): Promise<R
...
+    // A name that does not exist comes back as a 400 naming it, not a 404, so the
+    // status alone cannot distinguish "absent" from "malformed request".
+    const absent = isNotFoundError(error)
+      || (error instanceof Error && /not found/i.test(error.message));
+    if (!absent) throw error;
+  }
+
+  try {
+    const entityId = await createEntityAwaitingReady(entities, name, fields);
+    return await entities.getById(entityId);
```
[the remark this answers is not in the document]
Kumarvarinder87: What it creates is deliberately NOT deleted afterwards, unlike the transient `sdk_*` entities the schema suite deletes: this is the configured fixture named by DATA_FABRIC_TEST_TREE_ENTITY_NAME, reused by every later run, and every cell of this suite (both init modes, each credential) shares that one name — a cell dropping it on its way out would delete the entity a sibling cell is still writing to. Only the records are cleaned up (see the suite's afterAll).

src/models/data-fabric/entities.models.ts
```diff
@@ -1444,6 +1518,13 @@ function createEntityMethods(entityData: RawEntityGetResponse, service: EntitySe
       return service.updateRecords({ id: entityData.id }, data, options);
     },
 
+    // Passes `{ name }` where the other delegates pass `{ id }`: the upsert route addresses
+    // entities by name, so an id ref would make the service resolve the name we already hold.
+    async upsert(data: Record<string, any>, options?: EntityUpsertOptions): Promise<EntityUpsertResponse> {
+      if (!entityData.name) throw new Error('Entity name is undefined');
+      return service.upsert({ name: entityData.name }, data, options);
+    },
```
Sarath1018, line 1526: question: Can this hit the different entity with same name? like if I dont pass the folderkey will it upsert to a different entity in the tenant with same name

src/models/data-fabric/entities.types.ts
```diff
@@ -626,6 +626,78 @@ export interface EntityUpdateResponse extends EntityOperationResponse {}
...
+  /** Total records written across every entity in the transaction */
+  totalRecordsAffected: number;
+}
+
+/**
+ * Response from upserting a record.
+ *
+ * A single-record upsert returns the written record's fields. A transactional write across
+ * related entities reports its outcome in {@link transaction} instead, and carries no field
+ * values — the per-record IDs there are the result.
```
Sarath1018, line 676: Is the comment accurate? I cant see fields in the live API response
```
{
  "Id": "6cf9cff4-0dad-f111-9b33-6045bda94a92",
  "children": {},
  "cascadeDeletedChildren": {},
  "deletedCount": 0
}
```

docs/oauth-scopes.md
```diff
@@ -74,6 +74,7 @@ Before running the function, `invoke()` also acquires a Studio Web license for t
 | `deleteRecordById()` / `deleteRecord()` | `DataFabric.Data.Write` |
 | `updateRecordById()` / `updateRecord()` | `DataFabric.Data.Write` |
 | `updateRecordsById()` / `updateRecords()` | `DataFabric.Data.Write` |
+| `upsert()` | `DataFabric.Data.Write` |
```
Sarath1018: check if `DataFabric.Schema.Read` is also required as we unconditionally call resolveEntityName

Sarath1018 (approved):

src/models/data-fabric/entities.models.ts
Kumarvarinder87: Yes, this has been the same case with other end point like Create Record by name or Update record by name

src/models/data-fabric/entities.types.ts
Kumarvarinder87: Yes, when upsert is targeted to case entity or composite one, this is the response

docs/oauth-scopes.md
Kumarvarinder87: Added it

docs/oauth-scopes.md
```diff
@@ -74,6 +74,7 @@ Before running the function, `invoke()` also acquires a Studio Web license for t
 | `deleteRecordById()` / `deleteRecord()` | `DataFabric.Data.Write` |
 | `updateRecordById()` / `updateRecord()` | `DataFabric.Data.Write` |
 | `updateRecordsById()` / `updateRecords()` | `DataFabric.Data.Write` |
+| `upsert()` | `DataFabric.Data.Write` (also `DataFabric.Schema.Read` when called with `{ id }` — the name is resolved through the entity-metadata read) |
```
vnaren23: Do not have any explanation/documentation here. Just mention both scopes and move the text content into method description

src/models/data-fabric/entities.models.ts
```diff
@@ -547,6 +549,66 @@ export interface EntityServiceModel {
...
+   *
+   * A single-record write returns the record's fields. A tree write reports its outcome in
+   * `transaction` — one node per record, mirroring the request — and returns no field values.
+   *
+   * For a tree: at most 500 records, 3 levels of nesting, and 6 distinct entities per request.
+   * Passing `__Version__` in a record opts that row into a version check, which fails the whole
+   * transaction if the row changed in the meantime.
+   *
+   * Prefer `{ name }` over `{ id }` — the underlying route addresses entities by name, so an
+   * `{ id }` ref costs an extra lookup to resolve the name first.
```
vnaren23: Can we improve this please. This is not at all human readable.

src/models/data-fabric/entities.types.ts
```diff
@@ -626,6 +626,78 @@ export interface EntityUpdateResponse extends EntityOperationResponse {}
  */
 export interface EntityDeleteResponse extends EntityOperationResponse {}
 
+/**
+ * What a multi-entity transactional upsert did to one record.
```
vnaren23: Same here. Please have human readable descriptions

Kumarvarinder87: fixed

src/models/data-fabric/entities.models.ts
Kumarvarinder87: fixed

docs/oauth-scopes.md
Kumarvarinder87: fixed

tests/integration/shared/data-fabric/entities-records.integration.test.ts
```diff
@@ -1315,6 +1359,214 @@ describeIntegration('Data Fabric Entities Records - Integration Tests', 'both',
...
+
+      expect(tx.members).toHaveLength(1);
+      const child = tx.members[0];
+      expect(child.entityName).toBe(treeChildEntityName);
+      expect(child.op).toBe(EntityMultiEntityWriteOperation.Insert);
+      expect(child.id).toBeDefined();
+      expect(Array.isArray(child.members)).toBe(true);
+      expect(child.members).toHaveLength(0);
+
+      // Tracked here, later used to clean up test output.ß
```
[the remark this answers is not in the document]
Kumarvarinder87: fixed

src/models/data-fabric/entities.models.ts
```diff
@@ -547,6 +549,50 @@ export interface EntityServiceModel {
   updateRecordsById(id: string, data: EntityRecord[], options?: EntityUpdateRecordsOptions): Promise<EntityUpdateResponse>;
 
 
+  /**
+   * Upserts a record into an entity, and to its related child records. Two payload type:
+   *
+   * - One record: Only case an templated entities have one, for a native entity rejects this type.
```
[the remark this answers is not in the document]
Kumarvarinder87: fixed

src/models/data-fabric/entities.types.ts
```diff
@@ -626,6 +626,60 @@ export interface EntityUpdateResponse extends EntityOperationResponse {}
...
+ * Write response of a node in request tree. Results follow the same structure as the request.
+ */
+export interface EntityMultiEntityWriteResponseNode {
+  /** Name of the entity */
+  entityName: string;
+  /** Record was created or updated */
+  op: EntityMultiEntityWriteOperation;
+  /** ID of the record. */
+  id: string;
+  /** How many rows this record's write touched */
```
vnaren23: Number of rows the write affected

vnaren23 (approved):

src/models/data-fabric/entities.types.ts
Kumarvarinder87: Fixed

PR #740, opened 2026-09-16
feat(data-fabric): accept folderPath on entity, choiceset, roles options

AditiGoyalUipath:
EntityFolderScopedOptions now carries an optional folderPath alongside folderKey. Reads, record CRUD, attachments, and schema update all accept either — the DataFabric backend resolves folderPath to a folder key at auth time (PostAuthenticationHandler).

entities.create and choicesets.create still require folderKey because those endpoints authorize on entityDefinition.folderId from the body, not user.FolderKey — verified via live API that sending folderPath alone with the tenant-marker in the body silently lands the resource at tenant scope.

Header note: DataFabric uses X-UiPath-FolderPath (plain string) rather than the Orchestrator X-UIPATH-FolderPath-Encoded variant. The new buildDataFabricFolderHeaders helper is DF-specific for that reason.


src/utils/folder/folder-headers.ts
```diff
@@ -80,3 +80,23 @@ export function resolveFolderHeaders(input: ResolveFolderHeadersInput): Record<s
...
+ */
+export function buildDataFabricFolderHeaders(options?: {
+  folderKey?: string;
+  folderPath?: string;
+}): Record<string, string> {
+  const trimmedKey = options?.folderKey?.trim();
+  const trimmedPath = options?.folderPath?.trim();
+  return createHeaders({
+    [FOLDER_KEY]: trimmedKey || undefined,
+    [FOLDER_PATH]: trimmedPath || undefined,
```
deepeshrai-tech, line 100: Has this been tested with a folder path that contains non-ASCII characters? I think we allow Unicode in folder names, and undici rejects header values outside ISO-8859-1, so a plain-string X-UiPath-FolderPath carrying something like Shared/Ünicode would fail at the fetch layer before it reaches DF ,plz confirm this is only meant for ASCII paths, or that non-ASCII really does work end-to-end.

PR #525, opened 2026-06-15
feat(agents): timeline services — errors/consumption/latency (4/5) [PLT-103787]

ashishupadhyay88:
## Summary

Fourth split PR from the fat Agents PR #438 (PLT-103787), branched off `main` per the fan-out + rebase-on-merge model (follows #492, #503, #505).

Adds three agent-level **timeline** methods to the `Agents` service:

| Method | Endpoint (`INSIGHTS_RTM_BASE`) | Returns |
|--------|--------------------------------|---------|
| `getErrorsTimeline(startTime, endTime, options?)` | `POST /Agents/errors` | `AgentGetErrorsTimelineResponse[]` — error counts grouped by agent |
| `getConsumptionTimeline(startTime, endTime, options?)` | `POST /Agents/consumptionTimeline` | `AgentGetConsumptionTimelineResponse[]` — AGU (Agent Units) consumption time-series |
| `getLatencyTimeline(startTime, endTime, options?)` | `POST /Agents/latencyTimeline` | `AgentGetLatencyTimelineResponse[]` — per-percentile (P50/P95) latency |

All three are non-paginated POSTs. The API always returns a `{ data: [...] }` envelope (the array present, empty or populated); the SDK unwraps it and returns the point array directly, matching the sibling Agent timeline services (memory, traces). They share the private `buildAgentFilterBody` helper, which serializes the required `startTime`/`endTime` window (`Date` → ISO 8601) plus optional `folderKeys`/`agentNames`/`projectKeys`/`agentId`/`processVersion` (and `limit`, errors-timeline only) from `AgentFilterOptions`.

## Example Usage

```typescript
import { Agents } from '@uipath/uipath-typescript/agents';

const agents = new Agents(sdk);

// Latency timeline in May 2025
const result = await agents.getLatencyTimeline(
  new Date('2025-05-01T00:00:00Z'),
  new Date('2025-06-01T00:00:00Z'),
);
result.forEach((point) => {
  console.log(`${point.date} ${point.name}: ${point.value} ms`);
});

// Scoped — errors timeline, top 5 agents in specific folders
const errors = await agents.getErrorsTimeline(
  new Date('2025-05-01T00:00:00Z'),
  new Date('2025-06-01T00:00:00Z'),
  { folderKeys: ['<folderKey1>'], agentNames: ['JokeAgent', 'StoryAgent'], limit: 5 },
);
```

## Types & naming

Each method exposes a **response/options pair** following the `{Entity}Get{Operation}` convention:

[...]

PR #478, opened 2026-06-01
feat(insights): add getInstanceStats to MaestroProcesses and Cases [PLT-103276]

ninja-shreyash:
## Method Added

| Layer | Method | Signature |
|-------|--------|-----------|
| Service | `maestroProcesses.getInstanceStats()` | `getInstanceStats(request: MaestroProcessStatsRequest): Promise<InstanceStats>` |
| Service | `cases.getInstanceStats()` | `getInstanceStats(request: MaestroProcessStatsRequest): Promise<InstanceStats>` |
| Bound | `process.getInstanceStats()` | `getInstanceStats(packageVersion: string, startTime: Date, endTime: Date): Promise<InstanceStats>` |
| Bound | `case.getInstanceStats()` | `getInstanceStats(packageVersion: string, startTime: Date, endTime: Date): Promise<InstanceStats>` |

`MaestroProcessStatsRequest` bundles `processKey`, `packageId`, `packageVersion`, `startTime`, `endTime` — shared with `getElementStats` since both endpoints take the same process scope + time range.

## Endpoint Called

| Method | HTTP | Endpoint | OAuth Scope |
|--------|------|----------|-------------|
| `getInstanceStats()` | POST | `/insightsrtm_/agenticInstanceStatus/InstanceCountByStatus` | `Insights.RealTimeData Insights OR.Folders.Read` |

- Reuses `buildInsightsCommonBody()` (same `commonParams` shape as `getElementStats`)
- Both services call `this.post()` directly and apply `transformData()` for field renaming
- Bound methods on process/case objects auto-fill `processKey` and `packageId`
- Extracted shared `DurationStats` interface — both `ElementStats` and `InstanceStats` extend it (6 shared duration percentile fields)

## Example Usage

```typescript
import { MaestroProcesses } from '@uipath/uipath-typescript/maestro-processes';

const maestroProcesses = new MaestroProcesses(sdk);

// Get instance stats for a process
const counts = await maestroProcesses.getInstanceStats({
  processKey: '<processKey>',
  packageId: '<packageId>',
  packageVersion: '1.0.1',
  startTime: new Date('2026-04-01'),
  endTime: new Date(),
});

console.log(`Total: ${counts.totalCount}`);
console.log(`Running: ${counts.runningCount}, Completed: ${counts.completedCount}`);
console.log(`Faulted: ${counts.faultedCount}, Avg duration: ${counts.avgDurationMs}ms`);

// Using bound method on a process object — auto-fills processKey and packageId
[...]

PR #748, opened 2026-09-17
chore: gate public API changes behind a committed surface snapshot

Raina451:
To catch breaking changes, CI now builds a list of every public signature — one line per exported class, interface, type and method — for the merge base and for your branch, and compares the two. So any change to what a consumer can call shows up.

we do this:
```
if (breaking)
    add `breaking-change` label + post comment with what changed
else if (new stuff added)
    short comment saying what got added
else
    nothing
```

Screenshot 2026-09-17 at 9 59 08 PM
Screenshot 2026-09-17 at 9 59 27 PM

Fork PRs get a read only token so the label and comment are skipped there, same as our other PR bots.

## Testing

Replayed against the real commits of every breaking change we've shipped, via `scripts/api-surface-replay.sh <ref>`. All five caught.

| Demo | Mimics | Change shape | Expected |
|---|---|---|---|
| #755 | #682 / 1.7.1 | parameter type changed in place | label + comment |
| #756 | #681 / 1.7.0 | public methods removed | label + comment |
| #757 | #478 / 1.5.0 | positional params to a Request object | label + comment |
| #758 | #626 / 1.6.0 | public field type narrowed | label + comment |
| #749 | #525 / 1.4.1 | public type renamed | label + comment |
| #759 | — | new optional field | comment, no label |
| #751 | — | internal-only change | nothing at all |
 will close these demo prs later


.github/workflows/pr-checks.yml
```diff
@@ -70,6 +70,83 @@ jobs:
...
         run: npm test
 
+  api-surface:
+    # Reports changes to the public API. Never fails: a breaking change gets the
+    # `breaking-change` label plus a comment, anything else gets a short comment,
+    # and a PR that touches nothing public gets neither.
+    #
+    # The label is the part that matters downstream -- it makes "does this release
+    # contain a breaking change?" answerable at release time, which is what went
+    # wrong in #682 (shipped in a patch, no release note).
```
Sarath1018, line 80: nit: not needed

.github/workflows/pr-checks.yml
```diff
@@ -70,6 +70,83 @@ jobs:
...
+          code=0
+          node scripts/check-api-surface.mjs --base "origin/${{ github.base_ref }}" \
+            > api-surface-report.txt 2>&1 || code=$?
+          echo "code=$code" >> "$GITHUB_OUTPUT"
+          cat api-surface-report.txt
+
+      # Nobody should have to run a command to satisfy a check. CI regenerates
+      # the snapshot and commits it back, so the PR diff carries the change and
+      # the baseline stays current on its own. Pushes made with GITHUB_TOKEN do
+      # not start another workflow run, so this cannot loop.
```
Sarath1018, line 125: According to this comment PR checks will not run. so will that require any human push to run the gates? looks like an issue

.github/workflows/pr-checks.yml
```diff
@@ -70,6 +70,83 @@ jobs:
...
+    # `breaking-change` label plus a comment, anything else gets a short comment,
+    # and a PR that touches nothing public gets neither.
+    #
+    # The label is the part that matters downstream -- it makes "does this release
+    # contain a breaking change?" answerable at release time, which is what went
+    # wrong in #682 (shipped in a patch, no release note).
+    #
+    # Fork PRs get a read-only token, so the label and comment are skipped there,
+    # same as the other PR bots in this repo.
+    runs-on: ubuntu-latest
```
Sarath1018, line 84: we use uipath-ubuntu-latest for other jobs

.github/workflows/pr-checks.yml
Raina451: not using commit feature now, so we are good. we always create on the fly for default(main) branch and pr branch to compare.

Sarath1018 (approved):

PR #750, opened 2026-09-17
demo: breaking change with the snapshot updated (expect CI green, removal visible)

Raina451:
Demonstration for #748. **Close, do not merge.** Stacked on the branch in demo 1.

Same rename, with `npm run api-surface:gen` run and the result committed.

Expected: the **API Surface** job passes, and the diff carries the breaking change in plain sight:

```diff
-  Entities.getByName(entityName:string,options?:EntityGetByNameOptions):Promise<EntityGetResponse>
+  Entities.getByName(entityName:string,options?:EntityLookupByNameOptions):Promise<EntityGetResponse>
-interface EntityGetByNameOptions extends EntityGetByIdOptions
+interface EntityLookupByNameOptions extends EntityGetByIdOptions
```

The gate does not block the change. It converts it from an invisible side effect into a reviewable decision — which is exactly what did not happen on #682, where the equivalent three lines were buried in a 2,274-line diff.

Generated with Claude Code


Raina451:
Superseded. This demonstrated a breaking change with the snapshot already committed, and the committed snapshot has been removed from #748 — the classification is now merge-base vs branch, so this case no longer differs from #749.

PR #766, opened 2026-09-21
fix(samples): fix npm install and build for sample apps

swati354:
Four samples could not be installed or built from a clean checkout:

- `conversational-agent-app`, `data-fabric-app`, `data-fabric-app-angular` declared SDK floors older than the APIs they use, so a fresh install resolved a version that fails typecheck.
- `data-fabric-app-angular` pinned typescript `~5.8.3` against Angular 22, which requires `>=6.0` `<6.1`. The committed lockfile encoded the peer violation.
- functions-app's lockfile resolved five `@uipath` packages from `npm.pkg.github.com`, which needs auth, so npm install failed with a 401.

Follow up - when there are changes in samples, build the samples apps in CI


vnaren23 (commented):
How do we prevent this in future?

deepeshrai-tech (approved):

PR #767, opened 2026-09-21
fix(auth): refresh access tokens ahead of expiry [APPS-37340]

shivendra6720:
## Summary
- External apps can now be configured with a shortened access token lifetime (300–3600s). At 300s the SDK refreshes 12x more often than at the 3600s default, so a request signed in a token's final moments can be rejected server-side once network latency and clock skew are added — and that 401 is not retried.
- A token is now treated as expired 60s before its actual expiry, so it is renewed before it can expire in flight.
- The buffer applies only where a refresh is actually possible. `hasValidToken()` and `loadFromStorage()` keep using the true expiry, so a reload late in a token's life restores the session instead of forcing a re-login.

## Side effect worth flagging

Both expiry checks now resolve `expiresAt` through `getExpiryMs()` before comparing:

```ts
// before
return new Date() >= tokenInfo.expiresAt;
// after
return Date.now() >= getExpiryMs(tokenInfo.expiresAt) - bufferMs;

// token-expiry.ts
export function getExpiryMs(expiresAt: Date | string): number {
  return expiresAt instanceof Date ? expiresAt.getTime() : new Date(expiresAt).getTime();
}
```

For a `Date` value the result is identical to the old comparison — that form already coerced both sides to numbers.

It differs when `expiresAt` arrives as an **ISO string**. The old line compared a `Date` object against a raw string, which JavaScript coerces with `Number()`, yielding `NaN`. Every comparison against `NaN` is false, so the token was treated as **never expiring**: it would never refresh, and calls would eventually 401.

That is reachable today. `ActionCenterTokenManager` forwards `expiresAt` straight from a `postMessage` without converting it, and the embedded protocol declares the field as `string` — the `Date` on `TokenInfo` is not enforced at runtime. **Both expiry readers now resolve either shape, so the bug is fixed everywhere the value is actually used.**


src/core/auth/host-token-request.ts
```diff
@@ -25,9 +26,19 @@ export function isValidHostOrigin(origin: string | null): boolean {
   }
 }
```
deepeshrai-tech, line 28: This allowlist doesn't cover gov cloud, dedicated tenants, on-prem Orchestrator, or regional cloud subdomains , is this intentional not sure but i guess eventually it will run on these envs as well?

shivendra6720: Fair point, the allowed list is an existing list, a static list can't cover on-prem/dedicated at all, so I think the fix is to also accept the origin of the configured baseUrl — the instance the SDK already talks to. Will create a ticket for this fix and add to sprint.

shivendra6720: this change was already done by the pr https://github.com/UiPath/uipath-typescript/pull/704

